PR #2, opened 2016-01-19
Simplify Connection

dasch:
* Require an open socket when initializing.
* Set up encoder and decoder when initializing.


codella:
nice! :+1:

Issue #9, opened 2016-01-21
Implement automatic partitioning


dasch:
There are several strategies for assigning partitions to messages:

1. Based on a digest of the message key;
2. round-robin;
3. randomized;
4. based on a digest of a user-supplied partition key; or
5. based on a user-supplied custom strategy.

I think if we have 1) and 4) we should have all cases covered, as the rest can be implemented in terms of those.

PR #12, opened 2016-01-22
Allow automatic partition assignment

dasch:
If `Producer#write` is passed a `partition_key` parameter, it will be used to compute the partition to which the message will be written. If neither `partition` nor `partition_key` is specified, the message key will be used instead.

Implements #9.


dasch:
@rbarfoed @codella I could use a review here.

codella:
Makes sense! :+1:

PR #13, opened 2016-01-22
Enforce configurable timeouts when using a socket

### LINKED ISSUE #8: Add timeouts to all network calls
Timeouts should be added when

- [x] connecting to a broker;
- [ ] writing requests; and
- [ ] reading responses.

We should *never* hang indefinitely.

dasch:
Closes #8.


codella:
:+1:

PR #17, opened 2016-01-25
Use Docker containers to build a test cluster

### LINKED ISSUE #16: Set up a full integration test suite
In order to ensure we don't introduce regressions, we should have a test suite running with realistic scenarios. Currently, setting up the Kafka cluster needed to run the tests is not even fully automated. Ideally, the suite would also allow simulating network failures, partition leader changes, etc.

I'm thinking that a Docker based setup would work the best.

dasch:
Closes #16.

### DIFF
--- a/spec/test_cluster.rb
+++ b/spec/test_cluster.rb
@@ -1,0 +1,116 @@
+require "docker"
+
+Docker.url = ENV.fetch("DOCKER_HOST")
+
+class TestCluster
+  DOCKER_HOSTNAME = URI(ENV.fetch("DOCKER_HOST")).host
+  KAFKA_IMAGE = "wurstmeister/kafka:0.8.2.0"
+  ZOOKEEPER_IMAGE = "wurstmeister/zookeeper"
+  KAFKA_CLUSTER_SIZE = 3
+
+  def initialize
+    @zookeeper = create(
+      "Image" => ZOOKEEPER_IMAGE,
+      "ExposedPorts" => {
+        "2181/tcp" => {}
+      },
+    )
+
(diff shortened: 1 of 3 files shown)


spec/test_cluster.rb
```diff
@@ -0,0 +1,116 @@
+require "docker"
+
+Docker.url = ENV.fetch("DOCKER_HOST")
+
+class TestCluster
+  DOCKER_HOSTNAME = URI(ENV.fetch("DOCKER_HOST")).host
```
codella: maybe we should also have a default?

spec/test_cluster.rb
```diff
@@ -0,0 +1,116 @@
...
+        "Image" => KAFKA_IMAGE,
+        "Links" => ["#{@zookeeper.id}:zk"],
+        "ExposedPorts" => {
+          "9092/tcp" => {}
+        },
+        "Volumes" => {
+          "/var/run/docker.sock" => {}
+        },
+        "Env" => [
+          "KAFKA_ADVERTISED_HOST_NAME=192.168.42.45",
```
codella: ip set in stone? :D

codella:
dropped a couple of comments, but nice job anyway!!! :+1:

spec/test_cluster.rb
dasch: I don't think there's a valid default, at least not on OS X :-/ This ENV var has to be set for the CLI Docker client to function anyway.

spec/test_cluster.rb
dasch: Nice catch, that should be the Docker host.

PR #19, opened 2016-01-25
Retry produce requests on node failure

dasch:
When flushing messages to the cluster, we initially discover the leaders for the partitions we need to write to. If one of the leaders has become unavailable since we last discovered the cluster, we'll have to rediscover it to find out which nodes have taken over leadership of the partitions previously owned by that node. Then we can retry. Since leader election is not instantaneous, we need to back off before trying again.

There's still a chance that one of the nodes becomes unavailable between the time we discover the partition leaders and the time we try to actually send the messages to those nodes – that case has not yet been handled, and is left for a future PR.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -12,6 +12,8 @@ module Kafka
   LeaderNotAvailable = Class.new(Error)
   NotLeaderForPartition = Class.new(Error)
   RequestTimedOut = Class.new(Error)
+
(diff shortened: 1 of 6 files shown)


dasch:
@codella can you review this?

bastien:
:+1:

rbarfoed:
:+1:

dasch:
:heart:

codella:
awesome! :+1:

PR #21, opened 2016-01-25
Use pry for the console.

dkatten:
Not a big deal, but I always find pry better for poking around and exploring (`cd` ftw)


dasch:
:heart:

PR #25, opened 2016-01-26
Topic is necessary

liukgg:
Topic is necessary when to create a new message

### DIFF
--- a/lib/kafka/message.rb
+++ b/lib/kafka/message.rb
@@ -2,10 +2,10 @@ module Kafka
   class Message
     attr_reader :value, :key, :topic, :partition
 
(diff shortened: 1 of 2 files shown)


dasch:
@liukgg thanks for the PR! However, keyword parameters without default values are still mandatory, so the code doesn't change anything in that regard.

liukgg:
Yes, you are right.
But I think that we should not  make the parameters "key" and  "partition" mandatory,
do you think so?

dasch:
In the latest code, Message has been removed. There's always a partition, it's just that's it's not required that the user specifies it.

I'm going to allow skipping the key, as well.

PR #29, opened 2016-01-26
Implement error handling and retry logic

dasch:
See the code documentation for a description of the design. If you have questions, I'd like to answer them in the docs rather than here :-)

I'd like to get some feedback on the design of the error handling. Specifically the way the buffer is used. In the current design, calling `flush` would never fail with an exception – rather, a bunch of errors would be logged and the buffer would retain all messages that could not be written. Alternatively, `flush` could raise an exception if any messages are left in the buffer after all attempts have been exhausted, but I'm not sure what the value of that would be. For us at least, it would be better to just keep running.

I'll have to implement a max buffer size, though, so eventually the client code would run into trouble. But with sufficient logging, I feel that that's the right way to go: make sure there's some time for the cluster to heal in case of availability issues.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -13,6 +13,9 @@ module Kafka
   NotLeaderForPartition = Class.new(Error)
   RequestTimedOut = Class.new(Error)
 
+  # Raised when a producer buffer has reached its maximum size.
+  BufferOverflow = Class.new(Error)
+
   # Raised if not all messages could be sent by a producer.
   FailedToSendMessages = Class.new(Error)
 
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -13,6 +13,9 @@ module Kafka
   NotLeaderForPartition = Class.new(Error)
   RequestTimedOut = Class.new(Error)
 
+  # Raised if not all messages could be sent by a producer.
+  FailedToSendMessages = Class.new(Error)
+
   # Raised if a replica is expected on a broker, but is not. Can be safely ignored.
   ReplicaNotAvailable = Class.new(Error)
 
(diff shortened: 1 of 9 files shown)


codella:
:+1:

codella:
.. and for now I haven't an actually opinion on the error handling. I think we should see how what you did will "scale" adding more code later on.

osheroff:
I think that masking the errors only really works for one type of producer, namely a web producer.  

If I'm building a stream processor with this library, I might try to write something like:

```
   msg = stream_a.read()
   stream_b.write(transform(msg))
   if msg.offset % 1000 == 0 
     stream_b.flush()
     stream_a.set_offset(msg.offset + 1)
   end
```

If the brokers for stream_b are down, I've ended up completely losing messages, and recovery is now painful.. whereas if `flush` threw me back an exception (maybe just the latest one?  Or an exception that wraps an array of exceptions encountered, or something), at least I haven't lost my position in stream_a.

You also note that there will need to be a max_queue size.  I think this is absolutely necessary, otherwise the failure mode for a broker being down is to run your caller out of memory.  The caller should probably be able to choose their behavior when the queue is full, either to block or to drop messages on the floor.

dasch:
@osheroff thanks for the feedback! You raise some good points. I think the best course of action will be to raise an exception if any messages are still left in the buffer after a flush – but I think wrapping existing exceptions would be overly complex, at least at first. I may simply include info on the topics and partitions that have unsent messages.

As for the full-queue behavior, I'd rather not implement client-side blocking right now, so raising an exception on `#write` would be the natural choice for me. When I implement an async producer I plan on having a background thread communicate with the client through a bounded queue, so I think the functionality will pop up there.

Thanks again for the feedback!

dasch:
I've incorporated the feedback:

* When the buffer is full, calling `#write` will result in BufferOverflow.
* When the buffer is non-empty after flushing, `#flush` will raise FailedToSendAllMessages (too verbose?).

PR #41, opened 2016-02-01
Extract transmission logic from Producer

dasch:
Transmission is responsible for sending buffered messages to the brokers in a cluster. There's no retry logic – that stuff is still handled by Producer.

Downside: the mutable MessageBuffer will be shared between Producer and Transmission. Part of the response handling is marking messages as sent in the buffer, so it's hard to avoid.

I'm not quite sure about the name. The idea is to represent a single pass at transmitting messages to the cluster, but I'm not sure if we should instead use something closer to "send", e.g. MessageSender... but that sounds pretty horrible...

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -1,6 +1,7 @@
 require "kafka/partitioner"
 require "kafka/message_buffer"
 require "kafka/protocol/message"
+require "kafka/transmission"
 
 module Kafka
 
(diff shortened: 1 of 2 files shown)


dasch:
@bquorning I'd like your take on this.

bquorning:
Messenger?

dasch:
@bquorning it sends message batches, possible subdividing the batches and spreading them among multiple recipients.

bquorning:
In that case Transmitter sounds better than Messenger. But I’m still not sure it’s the right name.

dasch:
Hmm. I'll leave it as is, then. I can always rename it, I guess, since it's not part of the public API. I should probably mark APIs as public/private...

PR #45, opened 2016-02-02
Test against Kafka 0.9

dasch:


codella:
@dasch is there something missing here?

dasch:
@codella I'm just waiting until we have 0.9 in production.

dasch:
Going to start looking into implementing consumer groups, so this has to be merged.

Issue #40, opened 2016-02-01
Add instrumentation hooks


dasch:
There are several moving parts in the producer, and it will be important to measure:

- [x] The size of local buffers, both in number of pending messages _and_ relative to the max buffer size.
- [x] Latency and throughput of individual API requests/responses.
- [x] Latency and throughput of high-level operations such as produce operations: #81.
- [x] Errors, timeouts, and retries.
- [x] Throughput and latency for individual topics/partitions: #60.
- [x] Per-message delivery delay – time spent in the buffer: #137.
- [x] Producer delivery rate – time between deliveries.

Exactly how the hooks will work also needs to be decided. One option would be to simply use ActiveSupport::Notifications if it's available, or an API compatible null sink if it's not. An alternative is a custom callback system, but that feels a bit like overkill to me.

PR #46, opened 2016-02-03
Add support for instrumentation

dasch:
Implements #40.

### DIFF
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -1,5 +1,6 @@
 require "stringio"
 require "kafka/socket_with_timeout"
+require "kafka/instrumentation"
 require "kafka/protocol/request_message"
 require "kafka/protocol/encoder"
 require "kafka/protocol/decoder"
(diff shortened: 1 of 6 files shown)


dasch:
@bquorning I'm not sure if this can be done in a cleaner way. Care to review?

dasch:
I think it may be possible to use a delegate pattern to separate instrumentation from execution – but in some cases it could be tricky. It would require adding more info to some exceptions, and subscribers would have to correlate multiple events, e.g. the `:attempts` key on `send_messages.kafka` wouldn't be directly available, but rather we'd instrument Transmission as well and emit an event for each attempt. I believe ActiveSupport::Subscriber has support for nested events, i.e. you'd do something like

```ruby
def send_messages(event)
  attempts = event.children.select {|e| e.name == "transmission.kafka" }.count

  # ...  
end

attach_to "kafka"
```

kafka.gemspec
```diff
@@ -32,4 +32,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "dotenv"
   spec.add_development_dependency "docker-api"
   spec.add_development_dependency "rspec-benchmark"
+  spec.add_development_dependency "activesupport", "~> 4.2.0"
```
bquorning: Doesn’t it work with the Rails 5 betas?

bquorning:
:+1: for the current approach.

kafka.gemspec
dasch: I always use the latest stable release from libs.

dasch:
I'm taking this in a different direction: instead of instrumenting the high-level parts, I'm instrumenting the low-level ones, starting with Connection.

Individual requests are tracked, and the following keys are reported in the payload:

* `:api`: a symbol referencing which API was invoked.
* `:request_size`: the number of bytes sent to the server.
* `:response_size` the number of bytes received from the server.

In case of a connection error, the `:exception` key will hold the name of the exception class as well as any error message.

PR #50, opened 2016-02-04
Handle flaky network conditions

dasch:
We ran into an issue in our staging environment where network connection errors caused `Errno::EPIPE` to be raised from `Socket#read`. This was not handled gracefully, and the error propagated up through the stack.

This PR contains a few fixes:

* Before, socket timeouts were not being properly set. This was discovered during testing.
* Sockets can no longer get into a bad state. Rather, if they raise an exception they're closed and discarded, and a new one will be opened the next time a connection is used.
* Gracefully handle unexpectedly reaching the end of a buffer when reading a response.

### DIFF
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -102,7 +102,7 @@ module Kafka
           end
         end
       end
-    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT => e
+    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
       @logger.error "Connection error: #{e}"
 
       close
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -38,7 +38,15 @@ module Kafka
 
       @logger.info "Opening connection to #{@host}:#{@port} with client id #{@client_id}..."
 
-      @socket = SocketWithTimeout.new(@host, @port, timeout: @connect_timeout)
+      connect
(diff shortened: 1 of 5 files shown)


lib/kafka/connection.rb
```diff
@@ -89,6 +102,12 @@ def request(api_key, request, response_class)
           end
         end
       end
+    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT => e
+      @logger.error "Connection error: #{e}"
+
+      close
```
ilkkao: Woudn't it be good to start new connection setup attempt right away and not to wait for next `request` call?

dasch: Unfortunately not – setting up a new connection may take time, so the code responsible for handling retries (Producer) should get to decide whether or not to try to reconnect. You can imagine what would happen if a broker is permanently down: if we tried to reconnect in Connection we'd be stuck in an endless loop.

ilkkao: Yes a backoff timer would be needed. Probably best then to leave the decision to the user as you said.

dasch: Retries and backoff is handled here: https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/producer.rb#L182-L186

dasch: One important design decision is to have a centralized retry loop. Since a typical produce request would include sending message batches to more than one broker, it makes sense to postpone a retry until messages have been sent to the other brokers.

Issue #47, opened 2016-02-03
Implement basic message fetching


dasch:
Fetching messages from a single partition should be easy to implement.

mensfeld:
Would be awesome if there would be an option for autocommit after data is being fetched. Also isn't there a way to get offset automatically from Kafka/Zookeeper to keep reading after restart or on a next fetch loop?

dasch:
Offset commits only make sense if there's the notion of a consumer group involved. This would just be for ad-hoc fetches, although some of the plumbing could be re-used when group consumers need to fetch messages.

mensfeld:
@dasch  yeah but consumer groups will be introduced right? ;)

dasch:
Sure – but they'll be separate from this. I imagine this will look something like:

```ruby
kafka = Kafka.new(...)

messages = kafka.fetch_messages(topic: "impressions", partition: 42, offset: 1234)
```

So there'll be no coordination. That's fine for some, non-distributed workflows, and should provide us with feature parity for Poseidon.

In theory it's possible to build a simplistic 0.8 compatible consumer on top of this, but I'm going to go for the full 0.9 API instead.

mensfeld:
@dasch basic means basic and indeed for non-distributed workflows it will act great but having groups in 1.0 is in my opinion a must be.

dasch:
Yup, that's on the roadmap for 1.0. I just need to make sure the Producer is rock solid before embarking on such a task.

mrkamel:
+1

dasch:
Implemented in #51.

PR #51, opened 2016-02-04
Allow fetching messages from a specific partition

dasch:
Implements #47.

Fetches a batch of messages from a specified partition.

###### To Do

- [x] Resolve magic offset names `:earliest` and `:latest` using the [Offset API](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetAPI(AKAListOffset)).

### DIFF
--- a/examples/simple-consumer.rb
+++ b/examples/simple-consumer.rb
@@ -1,0 +1,48 @@
+# Consumes lines from a Kafka partition and writes them to STDOUT.
+#
+# You need to define the environment variable KAFKA_BROKERS for this
+# to work, e.g.
+#
+#     export KAFKA_BROKERS=localhost:9092
+#
+
+$LOAD_PATH.unshift(File.expand_path("../../lib", __FILE__))
+
+require "kafka"
+
+# We don't want log output to clutter the console. Replace `StringIO.new`
+# with e.g. `$stderr` if you want to see what's happening under the hood.
+logger = Logger.new(StringIO.new)
+
+brokers = ENV.fetch("KAFKA_BROKERS").split(",")
+
+# Make sure to create this topic in your Kafka cluster or configure the
+# cluster to auto-create topics.
+topic = "text"
+
+kafka = Kafka.new(
+  seed_brokers: brokers,
+  client_id: "simple-consumer",
+  socket_timeout: 20,
+  logger: logger,
+)
+
+begin
+  offset = :latest
+  partition = 0
+
+  loop do
+    messages = kafka.fetch_messages(
+      topic: topic,
+      partition: partition,
+      offset: offset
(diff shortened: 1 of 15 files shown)


examples/simple-consumer.rb
```diff
@@ -0,0 +1,48 @@
...
+#
+# You need to define the environment variable KAFKA_BROKERS for this
+# to work, e.g.
+#
+#     export KAFKA_BROKERS=localhost:9092
+#
+
+$LOAD_PATH.unshift(File.expand_path("../../lib", __FILE__))
+
+require "kafka"
```
mensfeld: would be nice to add rubocop - would catch some small things like "" instead of ''

lib/kafka/broker.rb
```diff
@@ -30,6 +30,20 @@ def fetch_metadata(**options)
       @connection.send_request(request, response_class)
     end
 
+    def fetch_messages(**options)
```
mensfeld: would be nice if you could document things using (for example) yard - otherwise using this gem when it gets bigger and bigger might get harder

examples/simple-consumer.rb
dasch: I'm actually trying to stick to double quotes.

lib/kafka/broker.rb
dasch: This API is still in flux, and Broker is an internal concept. I'm thinking about ways to get rid of it completely. I'll exhaustively document all user-facing APIs once they're stable.

lib/kafka/client.rb
```diff
@@ -38,11 +39,109 @@ def initialize(seed_brokers:, client_id: DEFAULT_CLIENT_ID, logger:, connect_tim
...
+    #   response message set. Default is 1 MB. You need to set this higher if you
+    #   expect messages to be larger than this.
+    #
+    # @return [Array<Kafka::FetchedMessage>] the messages returned from the broker.
+    def fetch_messages(topic:, partition:, offset: :latest, max_wait_time: 10, min_bytes: 1, max_bytes: 1048576)
+      broker = @broker_pool.get_leader(topic, partition)
+
+      offset = resolve_offset(topic, partition, offset, broker)
+
+      options = {
```
mensfeld: can't we extract a metod for building options out of fetching?

examples/simple-consumer.rb
mensfeld: is there a reason? rubocop and other tools advice using single once. Also as I said - maybe it is worth considering adding rubocop to this gem. That way when more people contribute - code styling would be the same for all of contributors.

PR #60, opened 2016-02-10
Instrument writes to topic partitions

dasch:
Subscribers can track writes to individual partitions. This allows monitoring throughput per topic and partition.

### DIFF
--- a/lib/kafka/produce_operation.rb
+++ b/lib/kafka/produce_operation.rb
@@ -1,4 +1,24 @@
 module Kafka
+  # A produce operation attempts to send all messages in a buffer to the Kafka cluster.
+  # Since topics and partitions are spread among all brokers in a cluster, this usually
+  # involves sending requests to several or all of the brokers.
+  #
+  # ## Instrumentation
+  #
+  # When executing the operation, an `append_message_set.kafka` notification will be
+  # emitted for each message set that was successfully appended to a topic partition.
+  # The following keys will be found in the payload:
+  #
+  # * `:topic` — the topic that was written to.
+  # * `:partition` — the partition that the message set was appended to.
+  # * `:offset` — the offset of the first message in the message set.
+  # * `:message_count` — the number of messages that were appended.
+  #
+  # If there was an error appending the message set, the key `:exception` will be set
+  # in the payload. In that case, the message set will most likely not have been
+  # appended and will possibly be retried later. Check this key before reporting the
+  # operation as successful.
+  #
   class ProduceOperation
     def initialize(broker_pool:, buffer:, required_acks:, ack_timeout:, logger:)
       @broker_pool = broker_pool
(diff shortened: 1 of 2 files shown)


dasch:
@bquorning I could use some feedback on this.

lib/kafka/produce_operation.rb
```diff
@@ -1,4 +1,25 @@
 module Kafka
+  # A produce operation attempts to send all messages in a buffer to the Kafka cluster.
+  # Since topics and partitions are spread among all brokers in a cluster, this usually
+  # involves sending requests to several or all of the brokers.
+  #
+  # ## Instrumentation
+  #
+  # When executing the operation, a `append_message_set.kafka` notification will be
```
bquorning: `s/a/an/`

lib/kafka/produce_operation.rb
```diff
@@ -1,4 +1,25 @@
...
+  # emitted for each message set that was successfully appended to a topic partition.
+  # The following keys will be found in the payload:
+  #
+  # * `:topic` — the topic that was written to.
+  # * `:partition` — the partition that the message set was appended to.
+  # * `:offset` — the offset assigned to the first message in the message set appended
```
bquorning: “assigned” sounds wrong here. Could it be “the offset of the first message…”?

lib/kafka/produce_operation.rb
```diff
@@ -1,4 +1,25 @@
...
+  #   to the partition.
+  # * `:message_count` — the number of messages that were appended.
+  #
+  # If there was an error appending the message set the key `:exception` will be set
```
bquorning: Comma after “set”, please.

lib/kafka/produce_operation.rb
dasch: It's actually taken from the Kafka protocol docs. I can change it, though.

lib/kafka/produce_operation.rb
```diff
@@ -1,4 +1,25 @@
 module Kafka
+  # A produce operation attempts to send all messages in a buffer to the Kafka cluster.
+  # Since topics and partitions are spread among all brokers in a cluster, this usually
+  # involves sending requests to several or all of the brokers.
+  #
+  # ## Instrumentation
+  #
+  # When executing the operation, a `append_message_set.kafka` notification will be
+  # emitted for each message set that was successfully appended to a topic partition.
```
bquorning: All the time I’m reading “message set” as if a message has been set. Would it be possible to write `message_set` instead?

dasch: Nah, it's a set of messages. Nothing peculiar about that. It's also the standard Kafka nomenclature, and I don't want to diverge from that too much.

lib/kafka/produce_operation.rb
bquorning: You know its meaning better than I do. Only change it if you agree it’s better.

lib/kafka/produce_operation.rb
bquorning: Sure, ok.

bquorning:
:+1:

PR #61, opened 2016-02-10
Cleaning up here and there

bquorning:
Removing trailing spaces, prefer using regular quotation instead of `%q`, etc.

I noticed that there is no spec covering the `if offset < 0` condition that I have changed. That may be an oversight.

@dasch


dasch:
:heart:

PR #65, opened 2016-02-11
Allow buffering messages even when Kafka is unavailable

dasch:
Until now, `Producer#produce` has required access to the cluster metadata in order to determine the partition that a given message should be assigned to. In the case of Kafka being unavailable, this would have blocked clients from buffering messages, as the metadata request would fail. Furthermore, since ffc1a34 the metadata has been invalidated the first time `#produce` is called with a topic not seen by the Producer instance before. This would mean a thundering herd of metadata requests when e.g. a cluster of web application processes are restarted. By postponing the metadata request until we're ready to send all the messages, we'll be able to do a single request containing *all* the topics.

This is done by adding a separate first-stage buffer to Producer. This is a flat list of pending messages that have not yet been assigned a partition. When sending messages, we first get the set of topics referenced by these messages and make sure there's metadata for them, possibly re-fetching the metadata. Then we use the metadata to assign each message to a partition and copy the messages into the MessageBuffer. From there it's business as usual.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -244,13 +244,18 @@ module Kafka
     private
 
     def assign_partitions!
-      @pending_messages.each do |message|
+      until @pending_messages.empty?
+        # We want to keep the message in the first-stage buffer in case there's an error.
+        message = @pending_messages.first
+
         partition = message.partition
 
         if partition.nil?
           # If no explicit partition key is specified we use the message key instead.
           partition_key = message.partition_key || message.key
-          partitioner = Partitioner.new(@broker_pool.partitions_for(message.topic))
+
+          partition_count = @broker_pool.partitions_for(message.topic)
+          partitioner = Partitioner.new(partition_count)
           partition = partitioner.partition_for_key(partition_key)
         end
 
@@ -260,9 +265,12 @@ module Kafka
           topic: message.topic,
           partition: partition,
         )
-      end
 
-      @pending_messages.clear
+        # Now it's safe to remove the message from the first-stage buffer.
+        @pending_messages.shift
+      end
+    rescue Kafka::Error => e
+      @logger.error "Failed to assign pending message to a partition: #{e}"
     end
   end
 end
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -1,7 +1,7 @@
 require "kafka/partitioner"
 require "kafka/message_buffer"
-require "kafka/protocol/message"
 require "kafka/produce_operation"
+require "kafka/pending_message"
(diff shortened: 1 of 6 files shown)


dasch:
@tcrayford if you have time to look at one PR, this would be it.

lib/kafka/producer.rb
```diff
@@ -181,6 +180,10 @@ def produce(value, key: nil, topic:, partition: nil, partition_key: nil)
     def send_messages
       attempt = 0
 
+      # Make sure we get metadata for this topic.
+      target_topics = @pending_messages.map(&:topic).uniq
+      @broker_pool.add_target_topics(target_topics)
```
dasch: An alternative approach would be to successively call e.g. `BrokerPool#add_target_topic` whenever we buffer a message and then e.g. `BrokerPool#ensure_topic_metadata_available!` here. I'm not sure it's an improvement, though, as we'd then need to keep a buffer of topics that we need to fetch metadata from in BrokerPool...

tcrayford:
> This is done by adding a separate first-stage buffer to Producer. This is a flat list of pending messages that have not yet been assigned a partition. When sending messages, we first get the set of topics referenced by these messages and make sure there's metadata for them, possibly re-fetching the metadata. Then we use the metadata to assign each message to a partition and copy the messages into the MessageBuffer. From there it's business as usual.

Is there a max limit to that buffer? I couldn't see it in the code. I'd worry about causing OOM with any kind of unbounded buffer like that.

dasch:
@tcrayford yup, `#produce` checks if `buffer_size` (which is the sum of both buffer sizes) exceeds the max buffer size.

dasch:
(Added a spec).

tcrayford:
aha, got it. This looks good. Have you got a TODO somewhere to document all the options in the README or something? Or are we just going to use generated docs for that?

tcrayford:
oh, nevermind, I found the docs.

dasch:
@tcrayford I've actually been debating whether to have a lot of code-level documentation or if it's better to just explain everything in the README. Doing both is a bit arduous, although I'd always have examples and high-level design stuff in the README. I'm waiting a bit until the design solidifies, then I'll do a bigger writeup on the README.

tcrayford:
@dasch that makes sense to me. I like having the details documented in the code docs, but think that could use some bolder calling out in the README - I didn't see it on a first scan.

dasch:
I added a link from the Usage section, how's that?

tcrayford:
Better. Thanks!

dasch:
This will have to do for now – I'll see if more refactoring is needed later.

PR #70, opened 2016-02-14
Makes the Partitioner configurable(no documentation yet)

Ben-M:
Not all Kafka clients have the same default partitioning logic (librdkafka and the Java client had different defaults last time I checked). This means that you need to be able to use a custom partitioner if you care about message order and have multiple clients producing to the same topic.

This pull request allows you to specify a custom Partitioner in calls to Client#get_producer. If one is not specified it continues to use the standard Partitioner (now rename DefaultPartitioner). If you like this approach let me know and I'll update the pull request with documentation.

It _is_ possible to keep the custom partitioner logic out of ruby-kafka, and instead have gem users handle this themselves, but to me it feels like a natural part of the Producer.

### DIFF
--- a/spec/producer_spec.rb
+++ b/spec/producer_spec.rb
@@ -5,6 +5,7 @@ describe Kafka::Producer do
   let(:broker1) { FakeBroker.new }
   let(:broker2) { FakeBroker.new }
   let(:broker_pool) { double(:broker_pool) }
+  let(:partitioner) { nil }
 
   let(:producer) {
     Kafka::Producer.new(
@@ -12,6 +13,7 @@ describe Kafka::Producer do
       logger: logger,
       max_retries: 2,
       retry_backoff: 0,
+      partitioner: partitioner
     )
   }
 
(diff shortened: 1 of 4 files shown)


spec/producer_spec.rb
```diff
@@ -140,5 +142,53 @@
       # The producer was not able to write the message, but it's still buffered.
       expect(producer.buffer_size).to eq 0
     end
+
+    context "when a partitioner is not specified" do
```
Ben-M: I'm not overly happy with how these specs turned out. Suggestions are very welcome here.

dasch:
@Ben-M I'd prefer to avoid adding too much complexity at this point – I feel that the combination of manual partition assignment and the `partition_key` API covers 95% of the use cases. How difficult would it be to add some code in front of ruby-kafka that assigned each message to a partition in a way that's compatible with whatever scheme your producers need to adhere to?

```ruby
def add_message(value, key)
  partition = ...
  @producer.produce(value, key: key, partition: partition, topic: ...)
end
```

I worry that adding too much API surface this early in the library's life cycle will make maintenance burdensome. Another issue is that many partitioning strategies are semantic in nature, and ruby-kafka was specifically designed to *not* handle serialization. This means that any partition strategy managed by ruby-kafka will only have serialized data to base decisions on. That's how Poseidon works, and it makes it a lot more difficult to write partitioning logic.

In one of my use cases, I wish to partition by an account id. That's easy to do with ruby-kafka, as I can simply pass the id as the `partition_key` and everything will work. In Poseidon I'd either have to use the message key for that purpose (thus losing the ability to store other stuff there) or have a partitioner that decodes the JSON, extracts the account id, and assigns the partition.

Ben-M:
@dasch fair points. Pushing this out of the gem (which the existing partition API makes easy) sounds like a good trade off to keep the API surface small.

I was personally surprised to learn that different Kafka clients partition based on a partition key differently. Would you be open to a PR adding a warning about this to the README?

dasch:
About having the partitioner receive the `partition_key`: I think that's a bad API, since it still doesn't cater to the needs of semantic partitioning, which would need access to the unserialized data to make a decision. I'd rather have a single way to do it: either have ruby-kafka assign the partition based on a key or assign the partition manually before calling `produce`.

I'll add an API for getting the partition count for a given topic.

Ben-M:
Sounds good, thanks.

I was definitely surprised when I found out that different Kafka clients implement the partition_key logic differently (so that partitioning by the same thing in different clients ends up with the messages on different partitions). Would you be open to a PR adding a note about this in the README?

dasch:
Sure! However, I'm planning on adding a longer section on partitioning in general to the README, so maybe I'll just add a note there.

Ben-M:
Cool - if you want a second pair of eyes on the README let me know.

dasch:
@Ben-M here's a PR: https://github.com/zendesk/ruby-kafka/pull/85

Feel free to leave feedback there.

PR #80, opened 2016-02-17
Rename Producer#send_messages to #deliver_messages

dasch:
I feel that "send" doesn't capture what's happening: the producer may "send" multiple requests, but it's all part of a single attempt at "delivering" the buffered messages.

This is a breaking API change, but at this point in ruby-kafka's development I'm not too worried about that. I want to have a nice API once we reach 1.0.

I'm also thinking about renaming `Producer#produce`, although I'm not sure what would be better. I'm afraid people will think `#produce` actually delivers the message...

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,10 +66,10 @@ If you don't know exactly how many partitions are in the topic, or you'd rather 
 producer.produce("hello4", topic: "test-messages", partition_key: "yo")
 ```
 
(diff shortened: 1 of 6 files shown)


bquorning:
I’m fine with “deliver” instead of “send”, but why do you add an exclamation mark?

dasch:
I guess because it's a "dangerous" operation in the sense that it may take several seconds to complete, depending on your timeout and retry settings.

codella:
@dasch `send` -> `deliver` looks like a good renaming, but I am not sure about the bang though - if you configure the library, then you know when it is going to "explode" :)

I'm :+1: with this one anyway.

dasch:
Removed the bang :-)

bquorning:
I removed the bang from the PR title, woot!

:+1:

dasch:
:party:

PR #81, opened 2016-02-17
Instrument producer operations

dasch:
Instruments the producer at two levels:

* `send_messages.kafka` – individual delivery attempts. The duration includes only the serialization, network, and server time spent delivering the messages.
* `deliver_messages.kafka` – a full delivery loop, including any retries, backoffs, cluster discoveries, etc. This will be the sum of `send_messages.kafka` durations.

The latter notification captures the duration of time the caller is blocked, and so is important to track. The former notification more granularly captures network and server performance.

I'm not entirely sure about the send/deliver nomenclature.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -46,6 +46,16 @@ module Kafka
   # not, we do another round of requests, this time with just the remaining messages.
   # We do this for as long as `max_retries` permits.
   #
+  # ## Instrumentation
+  #
+  # After {#deliver_messages} completes, the notification `deliver_messages.kafka`
+  #
+  # * `message_count` – the total number of messages that the producer tried to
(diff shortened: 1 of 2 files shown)


lib/kafka/producer.rb
```diff
@@ -225,22 +268,6 @@ def deliver_messages
       end
     end
 
-    # Returns the number of messages currently held in the buffer.
```
codella: IS this code snippet removal related in anyway to this PR? (just to understand if I am missing something in the picture :^))

codella:
It looks good, but I left a comment to understand something.

:+1:

lib/kafka/producer.rb
dasch: Nah, it's just the diff that's a bit weird – if you look further up in the file you can see these methods are there, still intact.

PR #85, opened 2016-02-18
Explain partitioning strategies in the README

dasch:


README.md
```diff
@@ -74,6 +74,65 @@ producer.deliver_messages
...
+
+# Make sure that we don't exceed the partition count!
+partition = some_number % partitions
+
+producer.produce(event, topic: "events", partition: partition)
+```
+
+#### Compatibility with Other Clients
+
+There's no standardized way to assign messages to partitions across different Kafka client implementation. If you have a heterogeneous set of producers it may be important to ensure a consistent partitioning scheme. This library doesn't try to implement all schemes, so you'll have to figure out which scheme the other client is using an replicate it. An example:
```
codella: maybe `implementation` -> `implementations`

codella:
Very nice description! :+1:

README.md
livathinos: `s/implementation/implementations`

livathinos: `s/is using an replicate/is using and replicate`

dasch: Thanks! Fixed.

livathinos:
:+1:

README.md
```diff
@@ -74,6 +74,65 @@ producer.deliver_messages
...
+producer.produce(msg1, topic: "messages", partition_key: partition_key)
+producer.produce(msg2, topic: "messages", partition_key: partition_key)
+
+# ...
+```
+
+You can also base the partition key on some property of the producer, for example the host name.
+
+#### Semantic Partitioning
+
```
Ben-M: Nice stuff!

Is it worth being explicit that message order is only guaranteed within a partition?  This is implied here, but might be good to be explicit.

PR #86, opened 2016-02-19
Implement compression

dasch:
Implementation of #7.

Adds support for Snappy and Gzip compression, both when producing and when fetching messages.

### DIFF
--- a/lib/kafka/compression.rb
+++ b/lib/kafka/compression.rb
@@ -1,16 +1,20 @@
 require "kafka/snappy_codec"
+require "kafka/gzip_codec"
 
 module Kafka
   module Compression
     def self.find_codec(name)
       case name
+      when nil then nil
       when :snappy then SnappyCodec.new
-      else nil
+      when :gzip then GzipCodec.new
+      else raise "Unknown compression codec #{name}"
       end
     end
 
     def self.find_codec_by_id(codec_id)
       case codec_id
+      when 1 then GzipCodec.new
       when 2 then SnappyCodec.new
       else raise "Unknown codec id #{codec_id}"
       end
--- a/lib/kafka/compression.rb
+++ b/lib/kafka/compression.rb
@@ -1,0 +1,19 @@
+require "kafka/snappy_codec"
+
+module Kafka
+  module Compression
+    def self.find_codec(name)
+      case name
+      when :snappy then SnappyCodec.new
+      else nil
+      end
+    end
+
+    def self.find_codec_by_id(codec_id)
+      case codec_id
+      when 2 then SnappyCodec.new
+      else raise "Unknown codec id #{codec_id}"
+      end
+    end
+  end
+end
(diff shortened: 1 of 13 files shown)


lib/kafka/compression.rb
```diff
@@ -0,0 +1,23 @@
+require "kafka/snappy_codec"
+require "kafka/gzip_codec"
+
+module Kafka
+  module Compression
+    def self.find_codec(name)
+      case name
+      when nil then nil
+      when :snappy then SnappyCodec.new
+      when :gzip then GzipCodec.new
```
mensfeld: wouldn't it be smarter to store an array of codecs like that:

```
CODECS = [
  { snappy: SnappyCodec },
  { gzip: GzipCodec }
].freeze
```
and then if nil - nothing, else if found new, otherwise fail. WOuld be easier for this and find_codec_by_id as well

lib/kafka/gzip_codec.rb
```diff
@@ -0,0 +1,28 @@
+module Kafka
+  class GzipCodec
```
mensfeld: Maybe would be better to Kafka::Codecs::Gzip, etc?

lib/kafka/gzip_codec.rb
```diff
@@ -0,0 +1,28 @@
...
+    def initialize
+      require "zlib"
+    end
+
+    def codec_id
+      1
```
mensfeld: it is per class not per instance - maybe a class attribute?

lib/kafka/compression.rb
dasch: Probably smarter, but definitely not easier to read. This is dumb and readable, and works great.

lib/kafka/gzip_codec.rb
dasch: I'm working very hard to avoid DSLs. There's nothing to be confused about here.

lib/kafka/gzip_codec.rb
dasch: I'm considering making these singletons.

lib/kafka/compression.rb
mensfeld: totally disagree (but its your gem ;) ) - having ids, names and classes defined in one place makes it easier to maintain and develop

lib/kafka/gzip_codec.rb
mensfeld: it's not about being confusing but about providing a nice and clean way to extend this gem. What if I would like to implement my own compressing layer? I would just inherit Kafka::Codecs::Base, implement what I need and that should be all...

dasch: My plan for that is to just pass in the codec in `compression_codec:`.

dasch: There's no reason for subclassing anything here, there's no need for shared behavior.

mensfeld: it;s not a shared behaviour - it is a dsl that would allow people to build things on top of this library

dasch: The codecs are defined by the Kafka spec – there's no way to build on top of this and stay compatible.

PR #87, opened 2016-02-19
Implement an async producer

### LINKED ISSUE #10: Implement an asynchronous producer
A worker should run in a background thread, writing buffered messages to the brokers according to some policy, e.g. periodically, when the buffer reaches a specified size, etc.

Clients can write to the buffer through e.g. a queue.

dasch:
Adds an async producer with options for automatic delivery:

* at a fixed time interval; and
* when the producer buffer reaches a configured threshold.

###### Tasks

- [x] Make sure there's a bound on the number of messages in the queue.
- [ ] Should we allow blocking on `#produce` if the queue is full?
- [ ] Ensure there are no races with the shared broker pool.

Closes #10.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -1,0 +1,181 @@
+require "thread"
+
+module Kafka
+
+  # A Kafka producer that does all its work in the background so as to not block
+  # the calling thread. Calls to {#deliver_messages} are asynchronous and return
+  # immediately.
+  #
+  # In addition to this property it's possible to define automatic delivery
+  # policies. These allow placing an upper bound on the number of buffered
+  # messages and the time between message deliveries.
+  #
+  # * If `delivery_threshold` is set to a value _n_ higher than zero, the producer
+  #   will automatically deliver its messages once its buffer size reaches _n_.
+  # * If `delivery_interval` is set to a value _n_ higher than zero, the producer
+  #   will automatically deliver its messages every _n_ seconds.
+  #
+  # By default, automatic delivery is disabled and you'll have to call
+  # {#deliver_messages} manually.
+  #
+  # The calling thread communicates with the background thread doing the actual
+  # work using a thread safe queue. While the background thread is busy delivering
+  # messages, new messages will be buffered in the queue. In order to avoid
+  # the queue growing uncontrollably in cases where the background thread gets
+  # stuck or can't follow the pace of the calling thread, there's a maximum
+  # number of messages that is allowed to be buffered. You can configure this
+  # value by setting `max_queue_size`.
+  #
+  # ## Example
+  #
+  #     producer = kafka.async_producer(
+  #       # Keep at most 1.000 messages in the buffer before delivering:
+  #       delivery_threshold: 1000,
+  #
+  #       # Deliver messages every 30 seconds:
+  #       delivery_interval: 30,
+  #     )
+  #
+  #     # There's no need to manually call #deliver_messages, it will happen
+  #     # automatically in the background.
+  #     producer.produce("hello", topic: "greetings")
+  #
+  #     # Remember to shut down the producer when you're done with it.
+  #     producer.shutdown
+  #
+  class AsyncProducer
+
+    # Initializes a new AsyncProducer.
+    #
+    # @param sync_producer [Kafka::Producer] the synchronous producer that should
+    #   be used in the background.
+    # @param max_queue_size [Integer] the maximum number of messages allowed in
+    #   the queue.
+    # @param delivery_threshold [Integer] if greater than zero, the number of
+    #   buffered messages that will automatically trigger a delivery.
+    # @param delivery_interval [Integer] if greater than zero, the number of
+    #   seconds between automatic message deliveries.
+    #
+    def initialize(sync_producer:, max_queue_size: 1000, delivery_threshold: 0, delivery_interval: 0)
+      raise ArgumentError unless max_queue_size > 0
+      raise ArgumentError unless delivery_threshold >= 0
+      raise ArgumentError unless delivery_interval >= 0
+
+      @queue = Queue.new
+      @max_queue_size = max_queue_size
+
+      @worker_thread = Thread.new do
+        worker = Worker.new(
+          queue: @queue,
+          producer: sync_producer,
+          delivery_threshold: delivery_threshold,
+        )
+
+        worker.run
+      end
+
+      @worker_thread.abort_on_exception = true
+
+      if delivery_interval > 0
+        Thread.new do
+          Timer.new(queue: @queue, interval: delivery_interval).run
+        end
+      end
+    end
+
+    # Produces a message to the specified topic.
+    #
+    # @see Kafka::Producer#produce
+    # @param (see Kafka::Producer#produce)
+    # @raise [BufferOverflow] if the message queue is full.
+    # @return [nil]
+    def produce(*args)
+      raise BufferOverflow if @queue.size >= @max_queue_size
+      @queue << [:produce, args]
+
+      nil
+    end
+
+    # Asynchronously delivers the buffered messages. This method will return
+    # immediately and the actual work will be done in the background.
+    #
+    # @see Kafka::Producer#deliver_messages
+    # @return [nil]
+    def deliver_messages
+      @queue << [:deliver_messages, nil]
+
+      nil
+    end
+
+    # Shuts down the producer, releasing the network resources used. This
+    # method will block until the buffered messages have been delivered.
+    #
+    # @see Kafka::Producer#shutdown
+    # @return [nil]
+    def shutdown
+      @queue << [:shutdown, nil]
+      @worker_thread.join
+
+      nil
+    end
+
+    class Timer
+      def initialize(interval:, queue:)
+        @queue = queue
+        @interval = interval
+      end
+
+      def run
+        loop do
+          sleep(@interval)
+          @queue << [:deliver_messages, nil]
+        end
+      end
+    end
+
+    class Worker
+      def initialize(queue:, producer:, delivery_threshold:)
+        @queue = queue
+        @producer = producer
+        @delivery_threshold = delivery_threshold
+      end
+
+      def run
+        loop do
+          operation, payload = @queue.pop
+
+          case operation
+          when :produce
+            @producer.produce(*payload)
+            deliver_messages if threshold_reached?
+          when :deliver_messages
+            deliver_messages
+          when :shutdown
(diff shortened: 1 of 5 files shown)


lib/kafka/client.rb
```diff
@@ -51,6 +52,16 @@ def get_producer(**options)
       Producer.new(cluster: @cluster, logger: @logger, **options)
     end
 
+    def get_async_producer(delivery_interval: 0, delivery_threshold: 0, **options)
```
tcrayford: I never said this previously, but you have any feelings about just calling these methods `async_producer` and `producer` without the `get_` prefix? I think it's nicer.

spec/functional/producer_spec.rb
```diff
@@ -11,10 +11,6 @@
     producer.shutdown
   end
 
-  example "listing all topics in the cluster" do
```
tcrayford: why was this removed in this PR?

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,117 @@
...
+  #       # Keep at most 1.000 messages in the buffer before delivering:
+  #       delivery_threshold: 1000,
+  #
+  #       # Deliver messages every 30 seconds:
+  #       delivery_interval: 30,
+  #     )
+  #
+  #     # There's no need to manually call #deliver_messages, it will happen
+  #     # automatically in the background.
+  #     producer.produce("hello", topic: "greetings")
```
tcrayford: this doc should include shutting down the producer

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,117 @@
...
+  #
+  class AsyncProducer
+    def initialize(sync_producer:, delivery_threshold: 0, delivery_interval:)
```
tcrayford: I'd like to see defaults here that aren't 0/nil. Maybe just 16 buffered messages and a delivery interval of 0? The Go producer has a 0 here, and it means every single production use we've done with it runs into issues until they tweak it.

The JVM and Go Kafka producers use bytes instead of number of messages here, and I think that's strictly superior - the reason you want to limit the max number of pending things is to prevent your memory from bloating to infinity when a runaway process produces too many things. (Your task about limiting the queue also has impacts there).

From a design perspective, the JVM producer also lets you pass a callback to track errors and successes. I guess in `ruby-kafka` the intention there is to just use the inbuilt metrics mechanism? I think that's notably worse than being able to pass a callback - a callback lets you do specific things on each success/failure, and in my experience that's often very useful (e.g. you may want to block the response of an http request until the kafka message it produced has been sent).

I don't think all of the above has to land in this PR, but I think they're important to making the producer operationally stable and usable without tuning.

lib/kafka/client.rb
tcrayford: also this method should validate that delivery_interval/threshold are numbers and not negatives, right? (and exception if they are bad). Otherwise things could get weird.

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    def produce(*args)
+      @queue << [:produce, args]
+    end
+
+    def deliver_messages
+      @queue << [:deliver_messages, nil]
+    end
+
+    def shutdown
```
tcrayford: maybe this method needs a doc that shutting down is asynchronous? Should we add a mechanism for doing synchronous shutdown? (you'd want to `Thread#join` the worker thread)

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,117 @@
...
+            @producer.deliver_messages if threshold_reached?
+          when :deliver_messages
+            @producer.deliver_messages
+          when :shutdown
+            break
+          else
+            raise "Unknown operation #{operation.inspect}"
+          end
+        end
+      ensure
```
tcrayford: woah, does `ensure` work with `break`?

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,117 @@
...
+        loop do
+          operation, payload = @queue.pop
+
+          case operation
+          when :produce
+            @producer.produce(*payload)
+            @producer.deliver_messages if threshold_reached?
+          when :deliver_messages
+            @producer.deliver_messages
+          when :shutdown
```
tcrayford: should this call a last `deliver_messages` before shutting down? Guess we could add a `shutdown_now` or something that doesn't do that?

lib/kafka/client.rb
dasch: I agree on both counts.

spec/functional/producer_spec.rb
dasch: It was a leftover – the example doesn't pertain to Producer.

lib/kafka/async_producer.rb
dasch: > I'd like to see defaults here that aren't 0/nil. Maybe just 16 buffered messages and a delivery interval of 0? The Go producer has a 0 here, and it means every single production use we've done with it runs into issues until they tweak it.

0 here means that it's disabled – I assume it's the same in the Go producer. I'm wary about setting too many defaults for the user. As it currently stands, the async producer works the same way as the sync producer, it's just async. Settings these options, however, enables _automatic_ behavior that I feel should be opt-in. There's still the max queue and buffer settings – those have non-default values, so if you forget to call `#deliver_messages` you'll run into errors eventually.


> The JVM and Go Kafka producers use bytes instead of number of messages here, and I think that's strictly superior - the reason you want to limit the max number of pending things is to prevent your memory from bloating to infinity when a runaway process produces too many things. (Your task about limiting the queue also has impacts there).

The delivery settings aren't really about setting a max – that's for the `max_buffer_size` setting (and the upcoming `max_queue_size` setting). The delivery settings are really about tuning performance – balancing efficiency (batch size) vs. message delay. I guess the README could be more explicit about this.

As for defining the max buffer size and auto-deliver buffer threshold in terms of bytes: I agree that it would be better, but I'm having a hard time figuring out how to do it correctly. I postpone encoding the data until the very last second in order to be able to buffer as much data as possible first, so I won't know the _exact_ byte size of the data. The closest approximation would be to keep track of the byte size of the key + value of buffered messages, and assume either a zero or fixed per-message overhead. That may be better than just tracking the number of buffered messages, but it's also slightly more difficult to define and explain. Would you consider an approximation to be good enough?

> From a design perspective, the JVM producer also lets you pass a callback to track errors and successes. I guess in `ruby-kafka` the intention there is to just use the inbuilt metrics mechanism? I think that's notably worse than being able to pass a callback - a callback lets you do specific things on each success/failure, and in my experience that's often very useful (e.g. you may want to block the response of an http request until the kafka message it produced has been sent).

This is unfortunately a bit complex – the actual success/failure handling is done in a separate thread, and allowing the user to pass in arbitrary closures would introduce nasty bugs into a typical, non-thread-safe Ruby program. As for your example: that's a perfect use case for the synchronous producer.

In the future I may reconsider a more flexible callback system, but I think it's premature at the moment :-/

> I don't think all of the above has to land in this PR, but I think they're important to making the producer operationally stable and usable without tuning.

:heart: the feedback!

lib/kafka/async_producer.rb
dasch: This _is_ synchronous – the next line joins the worker thread.

lib/kafka/async_producer.rb
dasch: Ensure means _ensure_ – the code will _always_ be executed (except for SIGKILL or something, of course...).

lib/kafka/async_producer.rb
dasch: Good idea.

lib/kafka/async_producer.rb
tcrayford: gah, missed it. Sorry.

dasch:
I'm leaning towards merging this even if there _are_ races in the broker pool. For typical use cases it's not a problem, and this isn't 1.0 yet. I'd like to do a separate review of the broker pool design.

tcrayford:
This looks reasonable to me.

lib/kafka/async_producer.rb
dasch: :heavy_check_mark:

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,170 @@
...
+    # @param sync_producer [Kafka::Producer] the synchronous producer that should
+    #   be used in the background.
+    # @param max_queue_size [Integer] the maximum number of messages allowed in
+    #   the queue.
+    # @param delivery_threshold [Integer] if greater than zero, the number of
+    #   buffered messages that will automatically trigger a delivery.
+    # @param delivery_interval [Integer] if greater than zero, the number of
+    #   seconds between automatic message deliveries.
+    #
+    def initialize(sync_producer:, max_queue_size: 1000, delivery_threshold: 0, delivery_interval: 0)
```
Gawyn: With the default parameters (`delivery_threshold` and `delivery_interval`) the messages are never delivered, aren't they?

dasch: They're only delivered when `#deliver_messages` is called.

lib/kafka/async_producer.rb
```diff
@@ -0,0 +1,181 @@
...
+  # work using a thread safe queue. While the background thread is busy delivering
+  # messages, new messages will be buffered in the queue. In order to avoid
+  # the queue growing uncontrollably in cases where the background thread gets
+  # stuck or can't follow the pace of the calling thread, there's a maximum
+  # number of messages that is allowed to be buffered. You can configure this
+  # value by setting `max_queue_size`.
+  #
+  # ## Example
+  #
+  #     producer = kafka.async_producer(
```
codella: I just love this interface...

codella:
I see that you have still 2 tasks pending, but I am :+1: for what I see here.

lib/kafka/async_producer.rb
dasch: :heart:

Gawyn:
:+1: Really nice piece of code. :smile:

dasch:
I think this is good enough – we're still at the beta stage and advising people against using this gem for critical use cases. I'll take a look at the remaining tasks after merging this and using it in prod a bit.

dasch:
Thanks for reviewing, everyone!

dolzenko:
Guys, on a quick glance it looks like this (AsyncProducer) should be safe for use from multiple threads?

dasch:
@dolzenko it may be, but you'll want to stress test that. Try starting a lot of threads and have them write to a shared producer, logging all errors.

PR #90, opened 2016-02-22
Allow specifying a compression threshold

dasch:
In our use case we have a single producer that writes to dozens of topics, with an uneven throughput distribution. This means that it would make sense to enable compression for some topics and not others. On the other hand, our writes sometimes spike on some topics.

Rather than decide up front which topics should have compression enabled and which should not, this API would allow the user to set a threshold defined in terms of the number of messages buffered for a single partition. Only if the threshold is reached are the messages compressed.

### DIFF
--- a/lib/kafka/protocol/message_set.rb
+++ b/lib/kafka/protocol/message_set.rb
@@ -3,9 +3,14 @@ module Kafka
     class MessageSet
       attr_reader :messages
 
-      def initialize(messages: [], compression_codec: nil)
+      def initialize(messages: [], compression_codec: nil, compression_threshold: 1)
         @messages = messages
         @compression_codec = compression_codec
+        @compression_threshold = compression_threshold
+      end
+
+      def size
+        @messages.size
       end
 
       def ==(other)
@@ -13,10 +18,10 @@ module Kafka
       end
 
       def encode(encoder)
-        if @compression_codec.nil?
-          encode_without_compression(encoder)
-        else
+        if compress?
           encode_with_compression(encoder)
+        else
+          encode_without_compression(encoder)
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/kafka/protocol/message_set.rb
```diff
@@ -3,9 +3,14 @@ module Protocol
     class MessageSet
       attr_reader :messages
 
-      def initialize(messages: [], compression_codec: nil)
+      def initialize(messages: [], compression_codec: nil, compression_threshold: 1)
```
bquorning: I’m looking at that `nil` and wondering we should pass a null codec as the default, and reduce `#compress?` to just compare the size vs threshold.?

dasch: I'm not quite sure I understand your feedback – I assume you want to have a null codec that simply doesn't touch the messages?

The problem with that is that the code paths for when compression is enabled and when it's disabled aren't the same – there's an additional "wrap" step we have to go through when compressing.

bquorning: > I assume you want to have a null codec that simply doesn't touch the messages?

That’s exactly what I had in mind. But now I see the differences in `encode_with_compression` and `encode_without_compression `.

dasch: Unfortunately the compression protocol in Kafka is slightly funky :-/

bquorning:
Is there a test already covering this use case?

dasch:
There's a functional test for compression – it's tricky to do a functional test for this code, since it's rather transparent to the user of the library.

dasch:
@bquorning added a unit test, how's that?

bquorning:
:+1:

PR #91, opened 2016-02-23
Allow specifying buffer limits in terms of byte size

dasch:
Since buffer size limits are mainly in place in order to avoid unbounded memory growth, defining the limits in terms of memory size makes sense. However, this does introduce some complexity – this is an attempt to see if it's worth the trouble.

### DIFF
--- a/lib/kafka/message_buffer.rb
+++ b/lib/kafka/message_buffer.rb
@@ -6,22 +6,28 @@ module Kafka
   class MessageBuffer
     include Enumerable
 
-    attr_reader :size
+    attr_reader :size, :bytesize
 
     def initialize
       @buffer = {}
       @size = 0
+      @bytesize = 0
     end
 
     def write(value:, key:, topic:, partition:)
-      @size += 1
       message = Protocol::Message.new(key: key, value: value)
+
       buffer_for(topic, partition) << message
+
+      @size += 1
+      @bytesize += message.bytesize
     end
 
     def concat(messages, topic:, partition:)
-      @size += messages.count
       buffer_for(topic, partition).concat(messages)
+
+      @size += messages.count
+      @bytesize += messages.map(&:bytesize).reduce(:+)
     end
 
     def to_h
@@ -48,6 +54,7 @@ module Kafka
     # @return [nil]
     def clear_messages(topic:, partition:)
       @size -= @buffer[topic][partition].count
+      @bytesize -= @buffer[topic][partition].map(&:bytesize).reduce(:+)
 
       @buffer[topic].delete(partition)
       @buffer.delete(topic) if @buffer[topic].empty?
--- a/lib/kafka/pending_message.rb
+++ b/lib/kafka/pending_message.rb
@@ -2,12 +2,16 @@ module Kafka
   class PendingMessage
     attr_reader :value, :key, :topic, :partition, :partition_key
 
+    attr_reader :bytesize
+
     def initialize(value:, key:, topic:, partition:, partition_key:)
       @key = key
       @value = value
       @topic = topic
       @partition = partition
       @partition_key = partition_key
+
+      @bytesize = key.to_s.bytesize + value.to_s.bytesize
     end
   end
 end
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -1,6 +1,8 @@
+require "set"
 require "kafka/partitioner"
 require "kafka/message_buffer"
 require "kafka/produce_operation"
+require "kafka/pending_message_queue"
 require "kafka/pending_message"
 require "kafka/compression"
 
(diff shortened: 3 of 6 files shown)


lib/kafka/producer.rb
```diff
@@ -218,6 +223,10 @@ def buffer_size
       @pending_messages.size + @buffer.size
     end
 
+    def buffer_bytesize
+      @pending_messages.map(&:bytesize).inject(0, &:+) + @buffer.bytesize
+    end
```
dasch: This would need to be optimized before merging.

tcrayford: by tracking it in a new instance var upon message insertion?

tcrayford:
Docs work for me.

lib/kafka/producer.rb
dasch: Yup. It's just slightly messy...

dasch: Added a smarter buffer for pending messages.

lib/kafka/producer.rb
```diff
@@ -247,9 +264,7 @@ def shutdown
     def deliver_messages_with_retries(notification)
       attempt = 0
 
-      # Make sure we get metadata for this topic.
-      target_topics = @pending_messages.map(&:topic).uniq
-      @cluster.add_target_topics(target_topics)
+      @cluster.add_target_topics(@target_topics)
```
dasch: Yay, killed an O(_n_) operation!

lib/kafka/message_buffer.rb
```diff
@@ -6,22 +6,28 @@ module Kafka
...
+      @size += 1
+      @bytesize += message.bytesize
     end
 
     def concat(messages, topic:, partition:)
-      @size += messages.count
       buffer_for(topic, partition).concat(messages)
+
+      @size += messages.count
+      @bytesize += messages.map(&:bytesize).inject(0, &:+)
```
bquorning: *Slightly* faster without the `&`:
```rb
@bytesize += messages.map(&:bytesize).reduce(:+)
```

lib/kafka/message_buffer.rb
```diff
@@ -48,6 +54,7 @@ def each
     # @return [nil]
     def clear_messages(topic:, partition:)
       @size -= @buffer[topic][partition].count
+      @bytesize -= @buffer[topic][partition].map(&:bytesize).inject(0, &:+)
```
bquorning: Same here, use `:+`.

lib/kafka/message_buffer.rb
dasch: Thx.

lib/kafka/pending_message.rb
```diff
@@ -9,5 +9,9 @@ def initialize(value:, key:, topic:, partition:, partition_key:)
       @partition = partition
       @partition_key = partition_key
     end
+
+    def bytesize
+      @key.to_s.bytesize + @value.to_s.bytesize
```
bquorning: I don't understand this line.

lib/kafka/producer.rb
```diff
@@ -178,11 +188,11 @@ def initialize(cluster:, logger:, compression_codec: nil, ack_timeout: 5, requir
     # @raise [BufferOverflow] if the maximum buffer size has been reached.
     # @return [nil]
     def produce(value, key: nil, topic:, partition: nil, partition_key: nil)
-      unless buffer_size < @max_buffer_size
+      unless buffer_size < @max_buffer_size && buffer_bytesize < @max_buffer_bytesize
         raise BufferOverflow, "Max buffer size #{@max_buffer_size} exceeded"
       end
```
bquorning: I think the exception message needs to be changed as well.

lib/kafka/pending_message.rb
dasch: It's the sum of the byte size of the message's key and value. Both can be nil.

dasch: I could compute the size in `initialize` and provide a reader...

lib/kafka/producer.rb
dasch: I should probably split that into two separate checks...

codella:
:+1:

lib/kafka/producer.rb
```diff
@@ -270,7 +285,7 @@ def deliver_messages_with_retries(notification)
         assign_partitions!
         operation.execute
 
-        if @pending_messages.empty? && @buffer.empty?
+        if @pending_message_buffer.empty? && @buffer.empty?
           break
         elsif attempt <= @max_retries
           @logger.warn "Failed to send all messages; attempting retry #{attempt} of #{@max_retries} after #{@retry_backoff}s"
```
bquorning: Maybe `buffer_size.zero?`?

dasch: :+1:

lib/kafka/producer.rb
```diff
@@ -218,6 +224,17 @@ def produce(value, key: nil, topic:, partition: nil, partition_key: nil)
         partition_key: partition_key,
       )
 
+      if buffer_size >= @max_buffer_size
+        raise BufferOverflow, "Max buffer size (#{@max_buffer_size} messages) exceeded"
+      end
+
+      if buffer_bytesize + message.bytesize >= @max_buffer_bytesize
```
bquorning: For one of the checks, you add the current `message` and not for the other. Off by one-error?

dasch: For the count limit, we know whether we'll exceed the limit if we're currently _at it_. For the byte size limit we need to see if adding the message would put us _over the limit_.

PR #93, opened 2016-02-23
Block the async producer thread during buffer overflows

dasch:
The producer thread should not dequeue operations as long as the underlying sync producer has a full buffer. Instead, the queue should start filling up, eventually propagating the buffer overflow error to the calling thread.

The implementation in this PR will rescue buffer overflows raised by the sync producer and call `#deliver_messages` until `#produce` no longer raises BufferOverflow. This could quite possibly be a bad idea, depending on the retry backoff time of the sync producer.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -18,6 +18,8 @@ module Kafka
   # By default, automatic delivery is disabled and you'll have to call
   # {#deliver_messages} manually.
   #
+  # ## Buffer Overflow and Backpressure
+  #
   # The calling thread communicates with the background thread doing the actual
   # work using a thread safe queue. While the background thread is busy delivering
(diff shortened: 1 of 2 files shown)


tcrayford:
This makes sense to me.

I think documenting the semantics here would be a good idea.

The JVM producer by default just blocks if the async producer's buffers are too full, but doing that in ruby seems somewhat tricky. Think we want that eventually though - it let's you backpressure incoming requests (if it's a web API for example) without having to catch `BufferOverflow` and retry (which I think is potentially quite tricky).

dasch:
@tcrayford I agree, and if only the Ruby standard library provided a bounded queue implementation we would already be there...

dasch:
@tcrayford updated the docs. I'll look into adding blocking behavior in a separate PR when I get around to it.

PR #95, opened 2016-02-23
Instrument compression

dasch:


lib/kafka/compression.rb
```diff
@@ -3,21 +3,45 @@
...
+      )
+
+      message_set = Protocol::MessageSet.new(messages: [wrapper_message])
+
+      Instrumentation.instrument("compress.producer.kafka", {
+        codec_name: codec.name,
+        codec_id: codec.codec_id,
+        original_bytesize: data.bytesize,
+        compressed_bytesize: compressed_data.bytesize,
+      })
```
dasch: Not quite sure about this approach – ideally we'd be able to measure compression rates per topic.

PR #97, opened 2016-02-24
Extract compression logic into Compressor

dasch:
Wraps the compression settings in an object that knows how to compress entire message sets.


dasch:
@bquorning does this make you all warm and fuzzy on the inside?

lib/kafka/compressor.rb
```diff
@@ -0,0 +1,34 @@
...
+    #   that will trigger compression.
+    def initialize(codec_name:, threshold:)
+      @codec = Compression.find_codec(codec_name)
+      @threshold = threshold
+    end
+
+    # @param message_set [Protocol::MessageSet]
+    # @return [Protocol::MessageSet]
+    def compress(message_set)
+      return message_set if @codec.nil? || message_set.size < @threshold
```
bquorning: I kinda liked the small `compress?` method. Why inline it?

bquorning:
Nice cleanup! :heart:

:+1: apart from a single comment.

lib/kafka/compressor.rb
dasch: Because `message_set` is now passed in as a param. I could have `compress?(message_set)`, but it wouldn't be quite as nice.

dasch: I'll add it.

dasch: Hmm, it's actually more readable to check that the message set _shouldn't_ be compressed. Think I'm going to leave it as-is.

dasch: Would have been this change:

```diff
diff --git a/lib/kafka/compressor.rb b/lib/kafka/compressor.rb
index 452ce84..0f633ac 100644
--- a/lib/kafka/compressor.rb
+++ b/lib/kafka/compressor.rb
@@ -18,7 +18,7 @@ module Kafka
     # @param message_set [Protocol::MessageSet]
     # @return [Protocol::MessageSet]
     def compress(message_set)
-      return message_set if @codec.nil? || message_set.size < @threshold
+      return message_set unless compress?(message_set)

       data = Protocol::Encoder.encode_with(message_set)
       compressed_data = @codec.compress(data)
@@ -30,5 +30,11 @@ module Kafka

       Protocol::MessageSet.new(messages: [wrapper_message])
     end
+
+    private
+
+    def compress?(message_set)
+      !@codec.nil? && message_set.size >= @threshold
+    end
   end
 end
```

bquorning: Good call.

PR #99, opened 2016-02-24
Implement Consumer Group API

dasch:
This is still very _very_ alpha level.

The consumer group API is very generic and therefore somewhat hard to grok. It's basically a general-purpose coordination API, with the consumer group API being a sub-protocol. The clients themselves are responsible for assigning partitions among themselves; one is selected as the leader for a specific generation and sends a list of member -> partitions assignments to the Kafka broker. This allows the clients to control assignment policies, but also makes it a bit more upfront work.

###### Tasks

- [x] Ensure heartbeats are sent periodically.
- [x] ~~Figure out a threading model that isn't too difficult to work with.~~ screw that, let's just keep everything synchronous for now.
- [x] Handle rebalances.
- [x] Fetch and commit offsets.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -66,6 +66,36 @@ module Kafka
       connect_to_broker(get_leader_id(topic, partition))
     end
 
+    def get_group_coordinator(group_id:)
+      @logger.debug "Getting group coordinator for `#{group_id}`"
+
+      refresh_metadata_if_necessary!
+
+      cluster_info.brokers.each do |broker_info|
+        begin
+          broker = connect_to_broker(broker_info.node_id)
+          response = broker.find_group_coordinator(group_id: group_id)
+
+          Protocol.handle_error(response.error_code)
+
+          coordinator_id = response.coordinator_id
+          coordinator = connect_to_broker(coordinator_id)
+
+          @logger.debug "Coordinator for group `#{group_id}` is #{coordinator}"
+
+          return coordinator
+        rescue GroupCoordinatorNotAvailable
+          @logger.debug "Coordinator not available; retrying in 1s"
+          sleep 1
+          retry
+        rescue => e
+          @logger.error "Failed to get group coordinator info from #{broker}: #{e}"
+        end
+      end
+
+      raise Kafka::Error, "Failed to find group coordinator"
+    end
+
     def partitions_for(topic)
       add_target_topics([topic])
       cluster_info.partitions_for(topic)
--- a/lib/kafka/consumer_group.rb
+++ b/lib/kafka/consumer_group.rb
@@ -1,0 +1,170 @@
+require "set"
+require "kafka/protocol/member_assignment"
+
+module Kafka
+  class ConsumerGroup
+    attr_reader :assigned_partitions
+
+    def initialize(cluster:, logger:, group_id:)
+      @cluster = cluster
+      @logger = logger
+      @group_id = group_id
+      @session_timeout = 30
+      @member_id = ""
+      @generation_id = nil
+      @members = {}
+      @topics = Set.new
+      @assigned_partitions = {}
+    end
+
+    def subscribe(topic)
+      @topics.add(topic)
+      @cluster.add_target_topics([topic])
+    end
+
+    def member?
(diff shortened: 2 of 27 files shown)


codella:
:+1:

lib/kafka/cluster.rb
```diff
@@ -66,6 +66,36 @@ def get_leader(topic, partition)
       connect_to_broker(get_leader_id(topic, partition))
     end
 
+    def get_group_coordinator(group_id:)
```
tcrayford: any reason not to call this just `group_coordinator` ?

lib/kafka/cluster.rb
```diff
@@ -66,6 +66,36 @@ def get_leader(topic, partition)
...
+          coordinator = connect_to_broker(coordinator_id)
+
+          @logger.debug "Coordinator for group `#{group_id}` is #{coordinator}"
+
+          return coordinator
+        rescue GroupCoordinatorNotAvailable
+          @logger.debug "Coordinator not available; retrying in 1s"
+          sleep 1
+          retry
+        rescue => e
```
tcrayford: feels like catching nearly all kinds of exceptions here is a thing for improvement in later PRs at the very least

lib/kafka/cluster.rb
```diff
@@ -66,6 +66,36 @@ def get_leader(topic, partition)
...
+
+          coordinator_id = response.coordinator_id
+          coordinator = connect_to_broker(coordinator_id)
+
+          @logger.debug "Coordinator for group `#{group_id}` is #{coordinator}"
+
+          return coordinator
+        rescue GroupCoordinatorNotAvailable
+          @logger.debug "Coordinator not available; retrying in 1s"
+          sleep 1
```
tcrayford: worth making this configurable?

lib/kafka/consumer_group.rb
```diff
@@ -0,0 +1,170 @@
...
+
+module Kafka
+  class ConsumerGroup
+    attr_reader :assigned_partitions
+
+    def initialize(cluster:, logger:, group_id:)
+      @cluster = cluster
+      @logger = logger
+      @group_id = group_id
+      @session_timeout = 30
```
tcrayford: make it configurable at some point

tcrayford:
@dasch this is a very nice start. I've left some feedback, but think nearly all of it is "TODO in a later PR".

I didn't comment on docs, but assume those will come later.

With regards to the threading model, the JVM 0.9 kafka consumer is synchronous everywhere, doesn't ever spawn any threads, and I think that's even more fitting in ruby.

lib/kafka/cluster.rb
dasch: Just consistency with `get_leader`. Will probably rename both in a separate PR.

lib/kafka/cluster.rb
dasch: I should probably fix that before merging, actually...

lib/kafka/cluster.rb
dasch: Maybe in the future – not sure when you would want that, though. I'd like to see a case for a specific value.

lib/kafka/consumer_group.rb
dasch: Definitely – this is a use case specific tradeoff. Will make it configurable later and add some docs.

dasch:
I think the Java version uses some fancy futures or something to do concurrent IO, but other than that it seems that you're right. I've also thought about splitting up the request/response handling in order to have the cluster handle multiple requests in parallel, but that's for another time.

tcrayford:
Ah: I think theirs triggers async io, but does the polling of the state of that IO synchronously.

They for sure do some fancy prefetching (at least it is documented that they do), in that they start fetching the *next* batch of messages whilst returning the current one to the user.

dasch:
I also thought about that – I'm going to push that a bit into the future. I don't think I'm going to optimize this library for crazy performance just yet, anyway. If you want that you probably don't want to use Ruby :-) I'm mainly focused on a nice API, good docs, good operability, etc.

tcrayford:
:+1: to that

PR #105, opened 2016-02-26
Assign partitions to consumers using a consistent hash ring

### LINKED ISSUE #103: Assign partitions to consumers using a consistent hash ring
To be elaborated upon.

dasch:
Still a work in progress – the distribution is too uneven at the moment.

Closes #103.

### DIFF
--- a/lib/kafka/consistent_hash_ring_assignment_strategy.rb
+++ b/lib/kafka/consistent_hash_ring_assignment_strategy.rb
@@ -1,0 +1,65 @@
+require "zlib"
+require "kafka/protocol/member_assignment"
+
+module Kafka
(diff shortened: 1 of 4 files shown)


dasch:
@osheroff what's your thinking on this? I probably won't merge this until I have some concrete use cases that would benefit. Did you use the high level Java consumer in Maxwell? Would partition stickiness make sense?

SyedMuhammadSufyian:
Guys any update on the feature ? Is it in any plan to merge ? 
Will not it be more better if we leave it to the user how the partitions are assigned to the consumer , which can be based on some specific strategy and which could be extendible as well . Just like data partioning strageies available in Kafka Connect , we have set od defined strategies and user can also have its custom strategy for the partioning.

PR #115, opened 2016-03-04
Allow specifying the offset commit policy as a threshold

dasch:
The threshold represents the number of messages that can be processed before an offset commit is required, thus putting an upper bound on duplicate message processing in case a consumer dies.


tcrayford:
:+1:

Issue #100, opened 2016-02-25
SSL Authentication


dasch:
Here's the Kafka docs: http://kafka.apache.org/documentation.html#security_configclients

jturkel:
@dasch - Coincidently I was going to start looking into adding SSL authentication support today. Let me know if there's anything I can do to help (e.g. code, tests, etc.).

tcrayford:
@jturkel I made a stab at it a few weeks ago but didn't get that far. I talked to @dasch and we agreed the right thing to do would be to make a new thing like https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/socket_with_timeout.rb that uses ruby's openssl library.

I won't have time to contribute until next week, but am very interested in this being worked on :)

dasch:
@jturkel @tcrayford you guys can collaborate if you want to – I don't have the setup to properly test SSL authentication, so it's not going to happen without some help.

@jturkel in general I'm interested in having the tires kicked, seeing what breaks, what performs poorly, etc. I'm also interested in various use cases and how the parameters have been tuned for them. That would help me improve documentation and provide a few default configurations for common use cases.

geemus:
This may be a helpful reference: https://github.com/excon/excon/blob/master/lib/excon/ssl_socket.rb. Happy to discuss/clarify as needed also.

jturkel:
@tcrayford - I should have time to work on this late next week. As a POC, I managed to hack `SocketWithTimeout` to talk to a Kafka broker via SSL so hopefully getting this working properly won't be too bad. I'll ping you next week so we can figure out how to tackle this.

tcrayford:
@jturkel you have sample code somewhere? I'm happy to take it over and productionize it.

jturkel:
@tcrayford - The prototype code (along with a script I used to generate the certificates) can be found here: https://gist.github.com/jturkel/282b44cad254a3ea26de. It's extremely rough so I won't be offended when you throw it all away ;)

tcrayford:
@jturkel nice, thankyou. I've been working on this all of today, and expect to hopefully PR tomorrow.

dasch:
Thanks @tcrayford and @jturkel!

PR #116, opened 2016-03-04
SSL connection support

tcrayford:
This lets this client support ssl connections. I was relatively unsure of how to add tests for this, but have a staging environment setup with SSL support, so I've tested a number of failure scenarios and such.

This implements https://github.com/zendesk/ruby-kafka/issues/100.

Ruby SSL socket support is a nightmare, and actually impossible to use to do timeouts "correctly". I've left a lot of comments in the SSLSocketWithTimeout class to that effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,9 @@ Although parts of this library work with Kafka 0.8 – specifically, the Produce
   2. [Consuming Messages from Kafka](#consuming-messages-from-kafka)
 3. [Logging](#logging)
 4. [Understanding Timeouts](#understanding-timeouts)
-5. [Development](#development)
-6. [Roadmap](#roadmap)
+5. [SSL](#ssl)
+6. [Development](#development)
+7. [Roadmap](#roadmap)
 
 ## Installation
 
@@ -306,6 +307,33 @@ When sending many messages, it's likely that the client needs to send some messa
     n * (connect_timeout + socket_timeout + retry_backoff) * max_retries
 
 Make sure your application can survive being blocked for so long.
+
+### SSL
+
+Kafka 0.9+ supports SSL for authentication and authorization. There are two possible modes of operation:
+
+#### You have a 0.9 broker with SSL setup and only want to encrypt traffic to it:
+
+In this case you just need to pass `ssl: true` to `Kafka.new`. You don't need an SSL context.
+
+```ruby
+kafka = Kafka.new(ssl: true, ...)
+```
+
+#### You have a 0.9 broker with SSL setup and want to use client certificates and Kafka's authentication mechanism:
+
+In this case you need to pass `ssl: true`, and an `ssl_context` setup with the client certificate you want to use:
+
+```ruby
+ssl_context = OpenSSL::SSL::SSLContext.new
+ssl_context.set_params(
+  cert: OpenSSL::X509::Certificate.new(YOUR_CLIENT_CERTIFICATE),
+  key: OpenSSL::PKey::RSA.new(YOUR_CLIENT_CERTIFICATE_KEY),
+)
+kafka = Kafka.new(ssl: true, ssl_context: ssl_context ...)
+```
+
+Without these options, ssl support is disabled, and ruby-kafka will throw exceptions when trying to connect to an SSL socket.
 
 ## Development
 
--- a/lib/kafka/broker_pool.rb
+++ b/lib/kafka/broker_pool.rb
@@ -2,12 +2,14 @@ require "kafka/broker"
 
 module Kafka
   class BrokerPool
-    def initialize(client_id:, connect_timeout: nil, socket_timeout: nil, logger:)
+    def initialize(client_id:, connect_timeout: nil, socket_timeout: nil, logger:, ssl: nil, ssl_context: nil)
       @client_id = client_id
       @connect_timeout = connect_timeout
       @socket_timeout = socket_timeout
       @logger = logger
       @brokers = {}
+      @ssl = ssl
+      @ssl_context = ssl_context
     end
 
     def connect(host, port, node_id: nil)
@@ -21,6 +23,8 @@ module Kafka
         connect_timeout: @connect_timeout,
         socket_timeout: @socket_timeout,
         logger: @logger,
+        ssl: @ssl,
+        ssl_context: @ssl_context,
       )
 
       @brokers[node_id] = broker unless node_id.nil?
--- a/lib/kafka/ssl_socket_with_timeout.rb
+++ b/lib/kafka/ssl_socket_with_timeout.rb
@@ -1,0 +1,158 @@
+require "socket"
+
+module Kafka
+
+  # Opens sockets in a non-blocking fashion, ensuring that we're not stalling
+  # for long periods of time.
+  #
+  # It's possible to set timeouts for connecting to the server, for reading data,
+  # and for writing data. Whenever a timeout is exceeded, Errno::ETIMEDOUT is
+  # raised.
+  #
+  class SSLSocketWithTimeout
+
+    # Opens a socket.
+    #
+    # @param host [String]
+    # @param port [Integer]
+    # @param connect_timeout [Integer] the connection timeout, in seconds.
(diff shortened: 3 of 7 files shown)


lib/kafka/broker_pool.rb
```diff
@@ -2,12 +2,14 @@
 
 module Kafka
   class BrokerPool
-    def initialize(client_id:, connect_timeout: nil, socket_timeout: nil, logger:)
+    def initialize(client_id:, connect_timeout: nil, socket_timeout: nil, logger:, ssl: nil, ssl_context: nil)
```
dasch: I'm not sure about this API – why not just have e.g. `ssl_context` and disable SSL if it's nil?

lib/kafka/ssl_socket_with_timeout.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    # Opens a socket.
+    #
+    # @param host [String]
+    # @param port [Integer]
+    # @param connect_timeout [Integer] the connection timeout, in seconds.
+    # @param timeout [Integer] the read and write timeout, in seconds.
+    # @param ssl_context [OpenSSL::SSL::SSLContext] which SSLContext the ssl connection should use
+    # @raise [Errno::ETIMEDOUT] if the timeout is exceeded.
+    def initialize(host, port, connect_timeout: nil, timeout: nil, ssl_context: nil)
```
dasch: I guess the `ssl_context` parameter isn't optional here.

lib/kafka/broker_pool.rb
tcrayford: because you don't actually always need an ssl context. If your broker doesn't use ssl client certs for auth, but instead just wants encrypted messages, then you don't need a context.

lib/kafka/ssl_socket_with_timeout.rb
```diff
@@ -0,0 +1,159 @@
...
+        end
+      end
+
+      # once that's connected, we can start initiating the ssl socket
+      if ssl_context
+        @ssl_socket = OpenSSL::SSL::SSLSocket.new(@tcp_socket, ssl_context)
+      else
+        @ssl_socket = OpenSSL::SSL::SSLSocket.new(@tcp_socket)
+      end
+      @ssl_socket.sync_close = true
```
dasch: What does `sync_close=` do?

dasch: I.e., moar comment plz

tcrayford: actually we don't need it. Going to rebase that fix into the main commit here.

lib/kafka/ssl_socket_with_timeout.rb
tcrayford: it is indeed. Want a comment on why? There are quite a few places I could put that comment…

lib/kafka/broker_pool.rb
dasch: Okay. I'm not familiar with the different use cases – it would be nice if SSL setup was documented in the README with a few use cases. Would also make it easier to review.

lib/kafka/ssl_socket_with_timeout.rb
dasch: I guess it's best to keep the docs and comments at the outermost layer.

lib/kafka/broker_pool.rb
tcrayford: ok. I started to write that, then realized I didn't know if you wanted to do it. I will attach it to this PR now.

lib/kafka/ssl_socket_with_timeout.rb
```diff
@@ -0,0 +1,158 @@
...
+          # catch exceptions from read_nonblock, and gradually build up
+          # our read buffer.
+          buffer << @ssl_socket.read_nonblock(num_bytes - buffer.length)
+        rescue IO::WaitReadable
+          unless IO.select([@ssl_socket], nil, nil, @timeout)
+            raise Errno::ETIMEDOUT
+          end
+          retry
+        rescue IO::WaitWritable
+          unless IO.select(nil, [@ssl_socket], nil, @timeout)
```
dasch: I'm not sure I quite understand how the SSLSocket interacts with the normal Socket. In [the SSLSocket docs](http://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLSocket.html#method-i-connect_nonblock) it seems that `IO.select` is being passed the original Socket, not the SSLSocket.

dasch: So would this maybe work?

```ruby
def read(num_bytes)
  unless IO.select([@tcp_socket], nil, nil, @timeout)
    raise Errno::ETIMEDOUT
  end

  @ssl_socket.read(num_bytes)
end
```

lib/kafka/broker_pool.rb
dasch: Awesome! I'll just edit stuff after merging if it needs to be consistent with the writing style of the README.

lib/kafka/ssl_socket_with_timeout.rb
tcrayford: it doesn't, and the SSLSocket docs are wrong as far as I can tell. I looked at two different ruby http libraries, and they all take this approach.

Your example doesn't work because the TCP socket can (and does) have data to read that's just for SSL communication, and nothing to do with the underlying layer.

lib/kafka/ssl_socket_with_timeout.rb
tcrayford: https://github.com/zendesk/ruby-kafka/commit/6af968998c3fd31fa1cbdd6c5028a648ff977118

lib/kafka/broker_pool.rb
tcrayford: https://github.com/zendesk/ruby-kafka/commit/6af968998c3fd31fa1cbdd6c5028a648ff977118

lib/kafka/ssl_socket_with_timeout.rb
dasch: OK. The SSL docs are horrible...

README.md
```diff
@@ -307,6 +308,33 @@ When sending many messages, it's likely that the client needs to send some messa
...
+```ruby
+ssl_context = OpenSSL::SSL::SSLContext.new
+ssl_context.set_params(
+  cert: OpenSSL::X509::Certificate.new(YOUR_CLIENT_CERTIFICATE),
+  key: OpenSSL::PKey::RSA.new(YOUR_CLIENT_CERTIFICATE_KEY),
+)
+kafka = Kafka.new(ssl: true, ssl_context: ssl_context ...)
+```
+
+Without these options, ssl support is disabled, and ruby-kafka will throw exceptions when trying to connect to an SSL socket.
```
dasch: Would a typical use case read the certificate and key from disk? In that case, maybe the example should include those parts.

dasch: Also: how heterogenous would the configurations be? i.e. would it make sense to simply have options for `ssl_certificate` and `ssl_key`?

dasch: It seems the [Kafka Client SSL docs](http://kafka.apache.org/documentation.html#security_configclients) describe a higher-level config, based on encrypted key files. I'm not sure I want to go that far, but having an easy-to-use API would be great.

dasch:
:heart:

README.md
tcrayford: there are advanced options that some folk will want to mess with. I think just passing in an `SSLContext` is easier to start off with for this PR, if that's ok? I'd be happy making a "higher level" API in a new PR, just want to be careful about the proliferation of more and more options.

dasch: If some people will want to be able to mess with the SSLContext directly then I think it's fine to just accept that as an option. We can add a nicer API that does configuration for you in the future.

lib/kafka/ssl_socket_with_timeout.rb
geemus: The worst!

PR #120, opened 2016-03-04
cleaner SSL interface

tcrayford:
instead of forcing the user to construct the SSL cert, take the most
common options and create an SSLContext from them.

This sidesteps a lot of potential security worries, and means our users don't need to read the docs for `SSLContext`.


dasch:
:heart:

PR #122, opened 2016-03-04
Set a no output logger by default

### LINKED ISSUE #121: Logger is not currently optional
The README says by default nothing is logged, but currently exceptions are raised unless a logger is passed to `Kafka.new`

```ruby
require "kafka"
kafka = Kafka.new(seed_brokers: BROKERS)
producer = kafka.producer
producer.produce("foo", topic: "test")
producer.deliver_messages

NoMethodError: undefined method `info' for nil:NilClass
    from .../ruby-kafka-0.2.0/lib/kafka/cluster.rb:39:in `add_target_topics'
```

`#info` is trying to be called on the nil logger object. This could be fixed by either setting a default logger (e.g. `Logger.new("/dev/null")`) or fixing all the logger calls to guard against nil.

sdball:
There are a lot of calls to `logger` so either a nil logger argument needs to either raise an ArgumentError or have a reasonable default.

Here I'm presuming that the logger argument being nil is merely development exploration of the gem, so defaulting to a nil logger with a warning is acceptable.

Fixes #121

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -21,7 +21,8 @@ module Kafka
         raise ArgumentError, "At least one seed broker must be configured"
       end
 
-      @logger = logger
+      warn("No logger was configured, defaulting to not log output.") if logger.nil?
+      @logger = logger || Logger.new(nil)
       @seed_brokers = seed_brokers
       @broker_pool = broker_pool
       @cluster_info = nil
--- a/spec/cluster_spec.rb
+++ b/spec/cluster_spec.rb
@@ -1,4 +1,42 @@
 describe Kafka::Cluster do
+  describe "#initialize" do
+    let(:seed_brokers) { ["test1:9092"] }
+    let(:broker_pool) { double(:broker_pool) }
+    let(:logger) { Logger.new(LOG) }
+
+    subject {
+      Kafka::Cluster.new(
+        seed_brokers: seed_brokers,
+        broker_pool: broker_pool,
+        logger: logger
+      )
+    }
+
+    context "when logger is not nil" do
+      it "sets the logger and does not warn" do
+        expect {
+          subject
+        }.to_not output.to_stderr
+        set_logger = subject.instance_variable_get("@logger")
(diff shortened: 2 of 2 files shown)


dasch:
Closed by https://github.com/zendesk/ruby-kafka/commit/4ab0619eb3f9dfbeff4f16d6ba78d0e7973987f7 instead.

guyisra:
@dasch doing this 
```
kafka = Kafka.new(seed_brokers: BROKERS)
producer = kafka.producer
producer.produce("foo", topic: "test")
producer.deliver_messages
```

on master still doesnt work. adding `logger: Logger.new(nil)` does help.
Was the fix merge?

dasch:
On latest master that snippet works.

dasch:
What exception are you seeing?

guyisra:
using 
`gem 'ruby-kafka', git: "https://github.com/zendesk/ruby-kafka.git", branch: "master"`

I am getting 
```
NoMethodError: undefined method `info' for nil:NilClass
    from .../ruby-kafka-0.2.0/lib/kafka/cluster.rb:39:in `add_target_topics'
```
unless I add the Logger.new(nil)

sdball:
@guyisra On master I don't get the error when I go through the snippet code anymore. Maybe try `bundle update ruby-kafka`?

sdball:
The fix works for me! Kinda bummed not to see specs around the behavior though.

dasch:
@sdball I typically only specify positive properties (i.e., stuff that _should_ happen) and not negative properties (stuff that should _not_ happen). Over-specification is just as problematic as under-specification is.

sdball:
Interesting, I haven't really seen that approach before. But passing a logger argument should result in a nil logger, right?

For my view I treat tests as the first code reuse of a system as well as executable documentation. That's why I like to cover both directions with a clear explanation.

If something changes and the tests need to change then that's should that should be carefully considered because the stated API is changing.

sdball:
Oops: **not** passing a logger should result in a nil logger.

PR #123, opened 2016-03-07
Refactor SSL context setup

dasch:


lib/kafka/client.rb
```diff
@@ -275,5 +257,32 @@ def partitions_for(topic)
...
+      return nil unless ca_cert || client_cert || client_cert_key
+
+      ssl_context = OpenSSL::SSL::SSLContext.new
+
+      if client_cert && client_cert_key
+        ssl_context.set_params(
+          cert: OpenSSL::X509::Certificate.new(client_cert),
+          key: OpenSSL::PKey::RSA.new(client_cert_key)
+        )
+      elsif !client_cert_key
```
tcrayford: this is incorrect - users should be able to not pass client cert stuff at all (see the README)

lib/kafka/client.rb
```diff
@@ -275,5 +257,32 @@ def partitions_for(topic)
...
+        raise ArgumentError, "Kafka client initialized with `ssl_client_cert` but no `ssl_client_cert_key`. Please provide both."
+      elsif !client_cert
```
tcrayford: this is incorrect - users should be able to not pass client cert stuff at all (see the README)
 Add a line note

lib/kafka/client.rb
dasch: Yeah, I see it now.

PR #124, opened 2016-03-07
Handle oversize messages being produced and rejected by the brokers

dasch:
Not sure how to deal with these:
- Discard the message?
- Raise an exception and have the user code deal with it?
- Magic?

### DIFF
--- a/spec/functional/producer_spec.rb
+++ b/spec/functional/producer_spec.rb
@@ -92,4 +92,14 @@ describe "Producer API", functional: true do
       KAFKA_CLUSTER.start_kafka_broker(0)
     end
   end
+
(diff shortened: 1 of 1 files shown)


michaeldauria:
I've had this happen to me with Poseidon, but they offer a setting that allows you to adjust the max_bytes: https://github.com/bpot/poseidon/blob/master/lib/poseidon/partition_consumer.rb#L102

Maybe it would make sense to take a pass at the Java properties and support most of them? https://kafka.apache.org/documentation.html#newconsumerconfigs

dasch:
What I mean by this is we should be able to handle the client producing a message that gets rejected by the broker because it's too big without losing other messages. It's not a simple problem to solve, though.

Issue #88, opened 2016-02-22
Allow blocking when producing with the async producer


dasch:
Currently, when calling `#produce` on an async producer when the message queue is full a BufferOverflow exception is raised. It would be nice to have the option of blocking instead. Bonus points if there's a configurable timeout.

Note that in order to do this correctly a more sophisticated bounded queue would need to be implemented.

hkroger:
This would be nice since it's also impossible to check the queue length at the moment to halt producing messages if queue is getting too long.

dasch:
Closing due to a lack of activity.

PR #125, opened 2016-03-07
Add an experimental blocking queue implementation

dasch:
An attempt at implementing #93 and #88.


dasch:
We might want to use [SizedQueue](https://ruby-doc.org/core-2.2.1/SizedQueue.html) instead.

PR #131, opened 2016-03-09
Allow stopping a consumer loop

dasch:
Work in progress. Not sure about thread safety here.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -52,6 +52,9 @@ module Kafka
 
       # Send two heartbeats in each session window, just to be sure.
       @heartbeat_interval = @session_timeout / 2
+
+      # Whether or not the consumer is currently consuming messages.
+      @running = false
     end
 
     # Subscribes the consumer to a topic.
@@ -86,7 +89,9 @@ module Kafka
     # @yieldparam message [Kafka::FetchedMessage] a message fetched from Kafka.
     # @return [nil]
     def each_message
-      loop do
+      @running = true
+
+      while @running
         begin
           batch = fetch_batch
 
@@ -107,6 +112,8 @@ module Kafka
 
             send_heartbeat_if_necessary
             mark_message_as_processed(message)
+
+            break if !@running
           end
         rescue ConnectionError => e
           @logger.error "Connection error while sending heartbeat; rejoining"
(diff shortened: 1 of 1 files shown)


tcrayford:
@dasch IMO we should put a mutex around this. Can we also force sending the offsets one last time in the loop before breaking?

dasch:
1) I'm mainly worried about contention here. I'd like to find a lock-free solution.
2) Offsets are being committed in the `ensure` clause, so that should be taken care of.

franckverrot:
> Not sure about thread safety here.

Should we use something like `Thread.current[:running]` do hold that value instead?

Also, `#each_message` doesn't guarantee we only consume each message once, right? Multiple threads could retrieve the same messages concurrently, right?

tcrayford:
@franckverrot the thing with `Thread.current[:running]` is that users may want to trigger a shutdown externally to the consumer thread. I think we'd want some concrete use cases there though.

tcrayford:
@dasch here's how to do it without locks on `each_message`: you just put a mutex on *writing* the variable, and no mutex on reading it. That means we may overrun a few messages longer whilst the different threads sync the memory, but that's IMO not a big concern.

dasch:
@franckverrot wrote:
> Also, `#each_message` doesn't guarantee we only consume each message once, right? Multiple threads could retrieve the same messages concurrently, right?

The Consumer is specifically single-threaded – adding asynchronous processing of messages is a bad idea in my opinion. Just add more partitions and spin up more consumers.

dasch:
@tcrayford wrote:
> @dasch here's how to do it without locks on `each_message`: you just put a mutex on *writing* the variable, and no mutex on reading it. That means we may overrun a few messages longer whilst the different threads sync the memory, but that's IMO not a big concern.

That's what i was thinking – as long as there's a reasonable bound on shutdown time it doesn't matter. However, at that point I'm wondering why it's even necessary to wrap the write with a mutex? If it's just a matter of memory barriers, and the mutex doesn't fix that anyway, why not just get rid of the mutex entirely?

I'm not 100% sure I understand how the compiler would treat these cases, though – would it just inline the value itself, making the write have no effect in some cases? I doubt MRI would take things that far, they seem to favor friendliness over performance, but maybe other implementations?

tcrayford:
MRI for sure doesn't do anything of the sort. JRuby might well, but I doubt any jruby folk will use this gem - they should be using a wrapper around the java API IMO.

tcrayford:
and yeah, can probably even just kill the mutex :)

dasch:
... so maybe we should just merge this as-is and see if anyone complains? :-)

tcrayford:
:+1:

dasch:
Okay, merging this as-is, seeing as this is still an alpha API.

PR #132, opened 2016-03-09
Batch oriented consumer API

dasch:
This API allows consumers to work on entire _batches_ of messages, where a _batch_ is a continuous set of messages from the same topic/partition. The use cases for typically revolve around wanting to amortize the cost of writing messages to an external store.

This is the proposed API:

```ruby
consumer = kafka.consumer(group_id: "my-group")

  txn = SomeStore::Transaction.new
  batch.messages.each {|message| txn.add(message.value) }
  txn.commit
end
```

Each message batch is yielded to the provided block; after the block returns, a heartbeat is sent to the group coordinator and perhaps the offset of the partition is committed (depending on configuration). This means that

1. The block should not spend more time than `session_timeout`, or the consumer will be kicked from the group.
2. The block must only return if _all_ messages in the batch have been successfully processed – otherwise, an exception must be raised.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -142,6 +142,50 @@ module Kafka
       @running = false
     end
 
+    def each_batch
+      loop do
+        begin
+          fetch_batches.each do |batch|
+            unless batch.empty?
+              Instrumentation.instrument("process_batch.consumer.kafka") do |notification|
+                notification.update(
+                  topic: batch.topic,
+                  partition: batch.partition,
+                  highwater_mark_offset: batch.highwater_mark_offset,
+                  message_count: batch.messages.count,
+                )
+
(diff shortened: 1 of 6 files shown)


lib/kafka/consumer.rb
```diff
@@ -88,25 +88,26 @@ def subscribe(topic, default_offset: :earliest)
...
-
-              yield message
+          poll.each do |batch|
+            batch.messages.each do |message|
+              Instrumentation.instrument("process_message.consumer.kafka") do |notification|
+                notification.update(
+                  topic: message.topic,
+                  partition: message.partition,
+                  offset: message.offset,
+                  lag: batch.highwater_mark_offset - message.offset,
```
dasch: This would be a key metric to track IMO.

PR #134, opened 2016-03-09
Discard partial responses from Kafka

### LINKED ISSUE #133: The last message in each group of fetched messages can have a truncated value
When I put 100,000 lorem ipsum messages into a topic they can all be retrieved fine from the kafka command line client, but the Ruby Kafka consumer was only getting 99,913 complete messages and 87 incomplete messages.

After a lot of digging I determined that the broker was running 87 requests that received 1148 messages and then one final request that received the remaining 124. The incomplete message was always the last message of each "full-size" batch of 1148.

Setting the `min_bytes` key did not seem to have any effect.

In my testing I've been able to get a consumer to retrieve all 100,000 complete records by adding a value size check to the loop. With this check the incomplete messages are not "processed" and the offset isn't incremented.

```
next unless message.value.size > 800
```

I'm going to keep digging and see if I can find out where the actual cut off is happening. I've done a fair bit of debugging in `fetch_operation.rb` and it looks like the problem is farther in.

```
# fetch_operation.rb#L66
response = broker.fetch_messages(**options) # => exhibits the truncated value problem
```

sdball:
The Kafka API docs say:

> As an optimization the server is allowed to return a partial message
> at the end of the message set. Clients should handle this case.

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchAPI

Since the message declares the number of bytes to read it's possible to
ensure that many bytes were read from the socket. If there's a mismatch
then do not process those bytes as a response.

Fixes #133

### DIFF
--- a/lib/kafka/protocol/decoder.rb
+++ b/lib/kafka/protocol/decoder.rb
@@ -92,7 +92,14 @@ module Kafka
       #
       # @return [String]
       def read(number_of_bytes)
-        @io.read(number_of_bytes) or raise EOFError
+        socket_data = @io.read(number_of_bytes) or raise EOFError
+
+        # check for partial response that should be ignored
+        if socket_data.size != number_of_bytes
+          raise EOFError
+        end
+
+        socket_data
       end
     end
   end


dasch:
This looks good :heart:

PR #139, opened 2016-03-14
Add an example of integrating with Rails

dasch:


tcrayford:
Looks good to me

dasch:
I thought about adding a null producer for the test env, but I guess that'll just complicate things...

PR #146, opened 2016-03-23
Need nil check for block because some points don't pass block

repeatedly:
For example, this point doesn't pass block: https://github.com/zendesk/ruby-kafka/blob/7ffcf2c953b3560014c9312533e6a7eed8663fb4/lib/kafka/producer.rb#L207


repeatedly:
Test failure on Circle-CI seems to not relate my patch.

dasch:
Thanks! I'll review these PRs once I'm back from vacation next week.

dasch:
:heart:

PR #147, opened 2016-03-23
Producer should not retry when required_acks is 0

repeatedly:
required_acks 0 means the producer never waits acks from the brokers.
Previous code causes duplicated messages because buffer isn't updated
by producer opertaions with required_acks 0

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -290,6 +290,12 @@ module Kafka
 
         if buffer_size.zero?
           break
+        elsif @required_acks.zero?
+          # No response is returned by the brokers, so we can't know which messages
+          # have been successfully written. Our only option is to assume that they all
+          # have.
+          @buffer.clear
+          break
         elsif attempt <= @max_retries
           @logger.warn "Failed to send all messages; attempting retry #{attempt} of #{@max_retries} after #{@retry_backoff}s"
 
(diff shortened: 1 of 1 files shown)


lib/kafka/producer.rb
```diff
@@ -300,13 +306,6 @@ def deliver_messages_with_retries(notification)
         end
       end
 
-      if @required_acks == 0
-        # No response is returned by the brokers, so we can't know which messages
-        # have been successfully written. Our only option is to assume that they all
-        # have.
-        @buffer.clear
-      end
```
dasch: I'd actually prefer moving this block above the `if buffer_size.zero?` line. Makes the code a bit more readable.

lib/kafka/producer.rb
```diff
@@ -290,6 +290,12 @@ def deliver_messages_with_retries(notification)
...
         if buffer_size.zero?
           break
+        elsif @required_acks.zero?
+          # No response is returned by the brokers, so we can't know which messages
+          # have been successfully written. Our only option is to assume that they all
+          # have.
+          @buffer.clear
+          break
         elsif attempt <= @max_retries
           @logger.warn "Failed to send all messages; attempting retry #{attempt} of #{@max_retries} after #{@retry_backoff}s"
```
mensfeld: wouldn't it be more readable if we would just break if... break if...?

lib/kafka/producer.rb
repeatedly: So the code looks like this?

```rb
if @required_acks.zero?
  @buffer.clear
  break
elsif buffer_size.zero?
  break
#...
```

PR #148, opened 2016-03-23
Change log level of 'Sending N messages' to debug

repeatedly:
ruby-kafka logs this message at each deliver_messages call and
it fills up log files with this message.
Other similar message is debug level so this should also be debug.


dasch:
We need *some* messages at the info level – I think this one is pretty essential.

PR #151, opened 2016-03-29
Add list_offset method to Kafka::Client

deltaroe:
Our use case required a way of seeing the offsets available on a topic/partition. The protocol method was implemented, but not exposed so I added a method to do that. It's not the cleanest, but maybe someone else will find it useful.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -259,6 +259,38 @@ module Kafka
       operation.execute.flat_map {|batch| batch.messages }
     end
 
+    # Returns the earliest or latest offset available for a topic and partition
+    #
+    # @param topic [String] the topic that messages should be fetched from.
+    #
+    # @param partition [Integer] the partition that messages should be fetched from.
+    #
+    # @param offset [Symbol] the offset to start reading from.
+    #   valid options are earliest/latest
+    #
+    # @return [Integer] the offset
+    def list_offset(topic:, partition:, offset:)
+      if offset == :earliest
+        offset = -2
+      elsif offset == :latest
+        offset = -1
+      end
+
+      broker = @cluster.get_leader(topic, partition)
+      response = broker.list_offsets(
+        topics: {
+          topic => [
+            {
+              partition: partition,
+              time: offset,
+              max_offsets: 1
+            }
+          ]
+        }
+      )
+      response.offset_for(topic, partition)
+    end
+
     # Lists all topics in the cluster.
     #
     # @return [Array<String>] the list of topic names.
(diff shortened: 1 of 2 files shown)


dasch:
Can you describe your use case? Does it make sense to expose a generic `list_offset` method, or would two separate methods, `earliest_offset` and `latest_offset` suffice?

deltaroe:
No reason it couldn't be two different methods and would simplify things slightly. I'll see about changing the contract.

dasch:
@deltaroe I'm interested in why this is even necessary for you, though. I only want to add APIs for use cases I understand, so as not to bloat the library.

deltaroe:
In our use case we're spawning many parallel threads to process messages and need the available offsets to be able to know how many workers to swan.

dasch:
How are the threads accessing the partition? Do each thread get a contiguous slice of it?

This is slightly veering off the standard Kafka model of concurrency, where a single thread is responsible for a partition. Have you thought about just increasing the partition count and using the high-level consumer? You'll lose in-order processing once you start sharing a partition among threads, which complicates things quite a bit.

deltaroe:
Yes the parallel workers can be assigned a range of offsets to process. This is only for when parallel operations are desired. In our setup we have a mix of consumers where some need strict ordering on a topic/partition and some can process messages in parallel just fine. To support the parallel workers our wrapper system needs to be aware of the available offsets.

dasch:
So you're using the low-level message fetch operations rather than consumer groups? One alternative would be to have a single consumer loop pull in large batches, then split those up and push them onto a thread safe queue that the workers read off. You could have a dynamically sized worker pool:

```ruby
require "kafka"
require "thread"

kafka = Kafka.new(...)
batch_queue = Queue.new
ack_queue = Queue.new

# The number of messages in a slice that a worker is assigned.
slice_size = 100

# Fetch at least 5MB at a time.
min_bytes = 1024 * 1024 * 5

consumer = kafka.consumer(group_id: "some-group")

consumer.each_batch(min_bytes: min_bytes) do |batch|
  batch.messages.each_slice(slice_size) do |slice|
    batch_queue << slice
  end

  # Here we really ought to wait for ACKs from all the workers – perhaps identify the slices
  # by their starting offset and wait for all of them to be acknowledged. If we don't do that
  # we risk message loss. 
  slice_offsets = batch.messages.each_slice(slice_size).map {|slice| slice.first.offset }
  until slice_offsets.empty?
    ack_offset = ack_queue.pop
    slice_offsets.delete(ack_offset)
  end
end
```

dasch:
@deltaroe do you have an update on this?

dasch:
@deltaroe closing this as there have been no updates.

PR #152, opened 2016-03-30
Instrument errors in the producer

dasch:
This change will make the producer send notifications whenever a) it's not possible to find the leader of a partition or b) there's an error for a partition in a produce response. Clients can subscribe to these notifications in order to implement error monitoring.


ncalca:
:+1:

PR #156, opened 2016-04-08
Prevent TypeError in favor of user friendly "No partitions assigned!"

dansimpson:
Although an edge case, the resulting error is understandable compared to: TypeError: no implicit conversion of nil into Array.


dasch:
:heart:

dasch:
I'll have to look at the behavior when there aren't enough partitions for all consumers – a valid scenario for a fail-over setup with e.g. one partition and two consumers.

PR #159, opened 2016-04-12
Continue assigning partitions when there's an error

dasch:
Before, a topic error would stop messages for other topics from being assigned to a partition and therefore from being delivered. In order to better isolate topic errors, this change goes through all pending messages and writes those without errors to the produce operation buffer, leaving the rest in the pending message queue.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -327,25 +327,35 @@ module Kafka
     end
 
     def assign_partitions!
-      @pending_message_queue.dequeue_each do |message|
+      failed_messages = []
+
+      @pending_message_queue.each do |message|
         partition = message.partition
 
-        if partition.nil?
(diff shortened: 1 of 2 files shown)


dasch:
@bquorning I could use a review of this.

lib/kafka/producer.rb
```diff
@@ -327,25 +327,35 @@ def deliver_messages_with_retries(notification)
...
+          end
+
+          @buffer.write(
+            value: message.value,
+            key: message.key,
+            topic: message.topic,
+            partition: partition,
+            create_time: message.create_time,
+          )
+        rescue Kafka::Error
```
bquorning: You're absolutely sure that this exception cannot occur during `@buffer.write`?

dasch: `@buffer.write` just stores the message in the buffer, it doesn't do any I/O.

PR #160, opened 2016-04-12
Support urls

tcrayford:
Quite often, services internally specify all resources in terms of URIs. E.g. postgres clients all accept `postgres://yourhost:yourport`. This handles that case for `ruby-kafka` as well.

Further, kafka seed broker lists are *very* commonly specified as a comma separated string rather than an array.

This is just a small ease of use thing to make getting started easier. I've cleaned up the examples to match.


lib/kafka/client.rb
```diff
@@ -307,5 +309,21 @@ def build_ssl_context(ca_cert, client_cert, client_cert_key)
...
+      seed_brokers.each do |connection|
+        if connection =~ /:\/\//
+          u = URI.parse(connection)
+          brokers << "#{u.host}:#{u.port}"
+        else
+          brokers << connection
+        end
+      end
+      brokers
+    end
```
dasch: Maybe

```ruby
[*seed_brokers].map { |str|
  if str.include?("://")
    uri = URI.parse(str)
    "#{uri.host}:#{uri.port}"
  else
    str
  end
}
```

dasch:
:heart:

tcrayford:
:heart:

PR #161, opened 2016-04-12
Instrument topic errors

dasch:
Currently, there's no direct way of tracking the errors encountered when assigning partitions to messages. Furthermore, the partition error instrument is confusing and too granular.

### DIFF
--- a/lib/kafka/produce_operation.rb
+++ b/lib/kafka/produce_operation.rb
@@ -65,9 +65,8 @@ module Kafka
         rescue Kafka::Error => e
           @logger.error "Could not connect to leader for partition #{topic}/#{partition}: #{e.message}"
 
-          @instrumenter.instrument("partition_error.producer", {
+          @instrumenter.instrument("topic_error.producer", {
             topic: topic,
-            partition: partition,
             exception: [e.class.to_s, e.message],
           })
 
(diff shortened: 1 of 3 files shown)


tcrayford:
LGTM :)

dasch:
We had a few production issues where the lack of direct instrumentation at the topic level made it harder than it should have been to figure out the root cause. Basically: before, writing to a non-existing topic and not specifying a partition would stop the producer from producing _any_ messages, as it would stop copying messages to the produce operation buffer when it encountered the message. This would cause data loss if the topic could not be created before the producer ran out of buffer space...

tcrayford:
yeah, oof :(

Do y'all create topics dynamically? I think that's a kinda antipattern for Kafka…

dasch:
We don't, it's just that we don't have proper automation in place yet – and a lot of production envs...

dasch:
BTW, balancing fault tolerance with error visibility is pretty hard, it turns out.

tcrayford:
ah, lots of envs :/ welp. And yeah that balance is tough.

dasch:
Even worse: the errors we've experienced are possibly ephemeral, so it's not possible to have a clear distinction between e.g. an expected, partial, and temporary unavailability (such as cluster changes) and a hard error scenario that needs manual intervention (a missing topic).

I'd like to eventually add the instrumentation code we've set up in Zendesk to ruby-kafka in some form, but unfortunately there's a lot of fragmentation in this area. We use the custom statsd library by Datadog in order to get tags, for example.

bquorning:
Perhaps the changelog should mention the rename from `partition_error.producer` to `topic_error.producer`?

dasch:
Those APIs are currently private and not documented.

bquorning:
Got it

PR #166, opened 2016-04-26
Lazily start async producer threads

### LINKED ISSUE #162: Can't produce async messages
I tried adding the example configuration for async producing of messages to an initializer in my Rails app as follows:

```ruby
$kafka = Kafka.new(seed_brokers: ['localhost:9092'], logger: Rails.logger)

$kafka_producer = $kafka.async_producer(delivery_interval: 1)

at_exit { $kafka_producer.shutdown }
```

Nothing happens when I try to produce a message from my Rails console. 

If I instantiate a new producer it works however. Any ideas?

dasch:
In some contexts, such as when starting a producer in a Rails initializer and using Unicorn as the web server, producer threads may die because the process forks. By starting the threads lazily, we postpone initialization until the fork has completed, thus avoiding the problem.

Should fix #162.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -77,23 +77,14 @@ module Kafka
       @queue = Queue.new
       @max_queue_size = max_queue_size
 
-      @worker_thread = Thread.new do
-        worker = Worker.new(
-          queue: @queue,
-          producer: sync_producer,
-          delivery_threshold: delivery_threshold,
-        )
+      @worker = Worker.new(
+        queue: @queue,
+        producer: sync_producer,
+        delivery_threshold: delivery_threshold,
+      )
 
-        worker.run
-      end
-
-      @worker_thread.abort_on_exception = true
-
-      if delivery_interval > 0
-        Thread.new do
-          Timer.new(queue: @queue, interval: delivery_interval).run
-        end
-      end
+      # The timer will no-op if the delivery interval is zero.
+      @timer = Timer.new(queue: @queue, interval: delivery_interval)
     end
 
     # Produces a message to the specified topic.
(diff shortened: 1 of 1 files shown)


lib/kafka/async_producer.rb
```diff
@@ -77,23 +77,13 @@ def initialize(sync_producer:, max_queue_size: 1000, delivery_threshold: 0, deli
...
+        producer: sync_producer,
+        delivery_threshold: delivery_threshold,
+      )
 
-      if delivery_interval > 0
-        Thread.new do
-          Timer.new(queue: @queue, interval: delivery_interval).run
-        end
-      end
+      @timer = Timer.new(queue: @queue, interval: delivery_interval)
```
dasch: This will _always_ start the timer thread, even when it's not needed. I'm not sure if that's a performance problem.

lib/kafka/async_producer.rb
```diff
@@ -128,18 +120,36 @@ def deliver_messages
...
+    end
+
     class Timer
       def initialize(interval:, queue:)
         @queue = queue
         @interval = interval
       end
 
       def run
+        Thread.stop if @interval.zero?
```
dasch: My understanding is that this will put the thread in a permanent sleep state:

```
irb(main):001:0> t = Thread.new { Thread.stop; puts "hello" }
=> #<Thread:0x007ffb4984ff98@(irb):1 run>
irb(main):002:0> t.alive?
=> true
```

lib/kafka/async_producer.rb
tcrayford: worth a comment at least though?

tcrayford:
:+1:

lib/kafka/async_producer.rb
dasch: Probably, yeah...

PR #174, opened 2016-04-27
Don't use stale offset information when consuming messages

### LINKED ISSUE #173: Consumer batch API consumes last message multiple times
The following code demonstrates the problem:

```ruby
kafka = Kafka.new(seed_brokers: ['kafka.server.local:9092'])
topic = "consumer-group-bug-#{Time.now.to_f}"

producer = kafka.producer
producer.produce('hello', key: 'world', topic: topic)
producer.deliver_messages

consumer = kafka.consumer(group_id: 'foo')
consumer.subscribe(topic)
consumer.each_batch do |message_batch|
  message_batch.messages.each do |message|
    puts "Received key=#{message.key}, value=#{message.value}"
  end
end
```

This outputs the following:
```
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
Received key=world, value=hello
...
```

dasch:
After committing offsets, the consumer would clear the information of which offsets had been processed but fail to refresh the broker supplied table, leading to erroneous offset information.

Should fix #173.

~~This means that whenever we commit offsets, we also fetch the updated offset table, which is obviously inefficient. I'll try to improve performance soon.~~ Added a commit that only clears the necessary info, since the other caches are cleared whenever the consumer re-joins the group.

### DIFF
--- a/lib/kafka/offset_manager.rb
+++ b/lib/kafka/offset_manager.rb
@@ -40,8 +40,8 @@ module Kafka
         @group.commit_offsets(@processed_offsets)
 
         @last_commit = Time.now
-        @processed_offsets.clear
-        @uncommitted_offsets = 0
+
+        clear_offsets
       end
     end
 
--- a/lib/kafka/offset_manager.rb
+++ b/lib/kafka/offset_manager.rb
@@ -41,7 +41,7 @@ module Kafka
 
         @last_commit = Time.now
 
-        clear_offsets
+        @uncommitted_offsets = 0
       end
     end
 


lib/kafka/offset_manager.rb
```diff
@@ -40,7 +40,7 @@ def commit_offsets
         @group.commit_offsets(@processed_offsets)
 
         @last_commit = Time.now
-        @processed_offsets.clear
```
MartinNowak: Think you should either clear `@committed_offsets` or merge `@processed_offsets` into `@committed_offsets.offsets`, b/c it now contains stale data.

dasch: The idea is that the consumer will simply keep using and updating `@processed_offsets`. Only when it has to rejoin the group does it clear its cache and re-sync.

PR #180, opened 2016-05-01
mark the just processed message before commit offset

MartinNowak:
- avoids redundant commiting and double processing on crash


dasch:
:heart:

PR #187, opened 2016-05-12
fix MessageBuffer#clear

MartinNowak:
- must reset bytesize or appending a message on a cleared
  buffer will fail w/ a Max buffer bytesize exception


dasch:
:heart:

PR #189, opened 2016-05-16
Allow configuring how producer buffer overflows are handled

dasch:
By default, an exception will be raised if the producer's buffer becomes full. This change will allow the client to configure the producer to instead prune the oldest messages in the buffer but keep the message that's currently being written. This handles an error condition where a topic or partition has is permanently unavailable for writes, making messages destined for it fill up the buffer and preventing messages for other topics from being delivered. By using this policy, the messages least likely to ever be successfully delivered -- those that have been buffered for the longest time -- will be dropped, rather than new messages that may have a chance of being delivered.

The option is called `buffer_overflow_policy`, and the allowed values are

* `:raise` — raise an exception (current behavior, still the default).
* `:prune_oldest` — new option.

I'd also like to allow the client thread to block when trying to write a message to a full buffer. Whether it makes sense to add that as an option is an open question.

###### Tasks

- [ ] Documentation.
- [ ] Fuzz testing.


osheroff:
So to be clear; this is a head-of-line blocking problem, so if one partition/topic goes unavailable it unstops the rest of the queue?

The other solution that comes to mind is maintaining per-topic (or even partition?) queues, so one may fill up and start dropping messages whilst the others proceed. This solution does do that, statistically, though maintaining separate buffers may be a bit more precise.

dasch:
> So to be clear; this is a head-of-line blocking problem, so if one partition/topic goes unavailable it unstops the rest of the queue?

Exactly.

> The other solution that comes to mind is maintaining per-topic (or even partition?) queues, so one may fill up and start dropping messages whilst the others proceed. This solution does do that, statistically, though maintaining separate buffers may be a bit more precise.

The current approach is pretty coarse-grained – I wanted something with minimal impact to the code base. The issue is that there are actually _two_ buffers in place: one for new messages that have not yet been assigned to a partition and one for messages that have been assigned. This change actually only covers the first buffer, since that's where you'll typically end up overflowing. There _are_ cases where it's possible to assign a partition (the topic itself has been created) but it's not possible to actually write to it (there's no leader or something) so I _should_ also cover the second buffer.

The first buffer is a simple wrapper around an array. The second is a specialized structure that keeps messages separately for each partition. I can iterate a bit on this.

codella:
@dasch it looks neat, so far!

PR #195, opened 2016-05-25
Added Kafka::ConnectionError to fix #194

### LINKED ISSUE #194: Make FetchOperation#execute handle Kafka::ConnectionError
My consumer is running into an infinite loop when the current kafka leader breaks (i.e. is shut down, has no connection, etc) even though a new leader is elected and up. This is happening, because FetchOperation#execute doesn't handle Kafka::ConnectionError to mark the cluster info as stale.

See: https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/fetch_operation.rb#L93

My consumer:

```ruby
client = Kafka.new(:seed_brokers => [...], :client_id => "consumer")

offset = :earliest

loop do
  begin
    messages = client.fetch_messages(:topic => "standard", :partition => 0, :offset => offset, :max_wait_time => 8)

    messages.each do |message|
      puts message.value
    end 

    offset = messages.last.offset + 1 if messages.size > 0 
  rescue => e
    puts e
    puts e.backtrace

    sleep 5
  end 
end
```

When i add Kafka::ConnectionError to the list of handled exceptions, everything is working fine and the consumer remains working using the newly elected leader.

Would be nice if Kafka::ConnectionError would be added.

Thanks

mrkamel:
See #194 for details

### DIFF
--- a/lib/kafka/fetch_operation.rb
+++ b/lib/kafka/fetch_operation.rb
@@ -90,7 +90,7 @@ module Kafka
           }
         }
       }
-    rescue Kafka::LeaderNotAvailable, Kafka::NotLeaderForPartition
+    rescue Kafka::ConnectionError, Kafka::LeaderNotAvailable, Kafka::NotLeaderForPartition
       @cluster.mark_as_stale!
 
       raise


dasch:
:heart:

PR #208, opened 2016-06-03
Accept IPv6 addresses for seed_brokers

charkost:
With the previous normalization/parsing process it was impossible
to use IPv6 addresses directly in the seed_brokers array/string.

This commit enables the client to accept IPv6 addresses with port
in the following URI-like format:

`[2a00:1450:400d:802::200e]:9092`

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -328,12 +328,10 @@ module Kafka
       end
       brokers = []
       seed_brokers.each do |connection|
-        if connection =~ /:\/\//
-          u = URI.parse(connection)
-          brokers << "#{u.host}:#{u.port}"
-        else
-          brokers << connection
-        end
+        connection.prepend("kafka://") unless connection =~ /:\/\//
+        uri = URI.parse(connection)
+        uri.port ||= 9092 # Default Kafka port.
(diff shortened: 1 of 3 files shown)


lib/kafka/client.rb
```diff
@@ -328,12 +328,9 @@ def normalize_seed_brokers(seed_brokers)
...
       seed_brokers.each do |connection|
-        if connection =~ /:\/\//
-          u = URI.parse(connection)
-          brokers << "#{u.host}:#{u.port}"
-        else
-          brokers << connection
-        end
+        connection.prepend("kafka://") unless connection =~ /:\/\//
+        u = URI.parse(connection)
+        brokers << { host: u.hostname, port: u.port || 9092 }
```
dasch: I actually think it would be preferable to just always store the broker addresses as a URI.

dasch: So:

```ruby
uri = URI.parse(connection)
uri.port ||= 9092
brokers << uri
```

charkost:
@dasch `Hash` -> `URI`. Please check again.

dasch:
:heart:

PR #211, opened 2016-06-08
Add a Datadog metrics reporter

dasch:
Metrics will be reported using Datadog's modified Statsd client.

This reporter is specific to [Datadog](https://www.datadoghq.com/), but other reporters can easily be added.


ciaranarcher:
👍 works for me

PR #212, opened 2016-06-09
PendingMessage: Use struct instead of class to improve performance

repeatedly:
Now I'm working on replacing poseidon with ruby-kafka in fluent-plugin-kafka.
I checked the performance and I found PendingMessage initialization consumes CPU because Producer#produce is called frequently.

I tested simple benchmark and it shows class with keyword argument is slower.
So using struct is better for `PendingMessage` usecase.

- Benchmark code:

```rb
require 'benchmark'

class PendingMessage1
  attr_reader :value, :key, :topic, :partition, :partition_key

  attr_reader :bytesize, :create_time

  def initialize(value:, key:, topic:, partition:, partition_key:, create_time:)
    @key = key
    @value = value
    @topic = topic
    @partition = partition
    @partition_key = partition_key
    @create_time = create_time

    @bytesize = key.to_s.bytesize + value.to_s.bytesize
  end
end

class PendingMessage2
  attr_reader :value, :key, :topic, :partition, :partition_key

  attr_reader :bytesize, :create_time

  def initialize(value, key, topic, partition, partition_key, create_time)
    @key = key
    @value = value
    @topic = topic
    @partition = partition
    @partition_key = partition_key
    @create_time = create_time

    @bytesize = key.to_s.bytesize + value.to_s.bytesize
  end
end

PendingMessageS = Struct.new("PendingMessageS", :value, :key, :topic, :partition, :partition_key, :create_time, :bytesize)


n = 1000000
value = "foooooooooooooooooooooooo"
key = "key"
topic = "kafka"
partition = 1
partition_key = "field"
create_time = Time.now.to_i

Benchmark.bmbm do |x|
  x.report('class1') {
    n.times do
      PendingMessage1.new(
        value: value,
        key: key,
        topic: topic,
        partition: partition,
        partition_key: partition_key,
        create_time: create_time)
    end
  }
  x.report('class2') {
    n.times do
      PendingMessage2.new(
        value, key, topic, partition, partition_key, create_time)
    end
  }
  x.report('struct') {
    n.times do
      PendingMessageS.new(
        value, key, topic, partition, partition_key, create_time,
        key.to_s.bytesize + value.to_s.bytesize)
    end
  }
end
```

- Result with ruby 2.3.1

```
Rehearsal ------------------------------------------
class1   1.830000   0.020000   1.850000 (  1.853243)
class2   0.630000   0.000000   0.630000 (  0.648974)
struct   0.430000   0.010000   0.440000 (  0.436509)
--------------------------------- total: 2.920000sec

             user     system      total        real
class1   1.830000   0.010000   1.840000 (  1.839984)
class2   0.630000   0.000000   0.630000 (  0.639363)
struct   0.430000   0.000000   0.430000 (  0.432769)
```

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -185,12 +185,13 @@ module Kafka
       create_time = Time.now
 
       message = PendingMessage.new(
-        value: value,
-        key: key,
-        topic: topic,
-        partition: partition,
-        partition_key: partition_key,
-        create_time: create_time,
+        value,
+        key,
+        topic,
+        partition,
+        partition_key,
+        create_time,
+        key.to_s.bytesize + value.to_s.bytesize
       )
 
       if buffer_size >= @max_buffer_size
--- a/lib/kafka/pending_message.rb
+++ b/lib/kafka/pending_message.rb
@@ -1,18 +1,11 @@
 module Kafka
-  class PendingMessage
-    attr_reader :value, :key, :topic, :partition, :partition_key
-
-    attr_reader :bytesize, :create_time
-
-    def initialize(value:, key:, topic:, partition:, partition_key:, create_time:)
-      @key = key
-      @value = value
-      @topic = topic
-      @partition = partition
-      @partition_key = partition_key
-      @create_time = create_time
-
-      @bytesize = key.to_s.bytesize + value.to_s.bytesize
-    end
-  end
+  PendingMessage = Struct.new(
+    "PendingMessage",
+    :value,
+    :key,
+    :topic,
+    :partition,
+    :partition_key,
+    :create_time,
+    :bytesize)
 end


dasch:
I don't think this matters at all in the great scheme of things. Can you measure an improvement in an actual use case that writes to a Kafka cluster? A microbenchmark doesn't prove that there's a real world effect.

repeatedly:
In fluent-plugin-kafka case, I tested this change with 20000msgs/sec.
The result is

- CPU usage is reduced from avg 70+% to around 65%
- One flush time(BufferedOutput#flush) is reduced from avg 0.7 sec to 0.55 sec. 

I'm not sure this case is popular in ruby-kafka but
some fluentd users use kafka plugin with 10000+msgs/sec.

lib/kafka/producer.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
dasch: Is this necessary for the change?

repeatedly: Ah, sorry. My editor inserts this automatically.
Will remove it.

dasch:
@repeatedly I'd like to set up a profiler script to verify that this is actually the bottleneck before merging this in, as I think the code is harder to follow.

dasch:
Hmm

```
%self      total      self      wait     child     calls  name
 31.82      0.142     0.048     0.000     0.095    10000   Kafka::Producer#produce
 16.59      0.035     0.025     0.000     0.010    10000   Kafka::PendingMessage#initialize
 10.27      0.050     0.015     0.000     0.035    10000   Class#new
  5.33      0.008     0.008     0.000     0.000    20000   Kafka::Producer#buffer_size
  4.94      0.007     0.007     0.000     0.000    10000   Kafka::PendingMessageQueue#write
  4.92      0.150     0.007     0.000     0.142        1   Integer#times
  4.70      0.014     0.007     0.000     0.007    10000   <Class::Time>#now
  4.30      0.006     0.006     0.000     0.000    10000   Set#add
  3.57      0.007     0.005     0.000     0.002    10000   Time#initialize
  3.15      0.005     0.005     0.000     0.000    10000   Kafka::Instrumenter#instrument
  2.83      0.004     0.004     0.000     0.000    20000   String#bytesize
  2.60      0.004     0.004     0.000     0.000    10000   Kafka::Producer#buffer_bytesize
  2.19      0.003     0.003     0.000     0.000    10000   NilClass#to_s
  1.45      0.002     0.002     0.000     0.000    10000   String#to_s
  1.27      0.002     0.002     0.000     0.000    10000   Fixnum#+
  0.02      0.150     0.000     0.000     0.150        1   Global#[No method]
```

dasch:
Why all the calls to `Class#new`?

repeatedly:
I'm not sure. How to measure above profiling?

dasch:
I just merged in a new profile script. Try rebasing on master and run `ruby performance/profile.rb`

dasch:
It may be better to simply remove the pending message queue altogether. It's there so that `#produce` has no dependency on Kafka network APIs, but I'm not sure if that level of resilience is necessary.

repeatedly:
Here is the result of profile.rb:

```
 %self      total      self      wait     child     calls  name
 34.44      0.106     0.039     0.000     0.067    10000   Kafka::Producer#produce
  7.02      0.014     0.008     0.000     0.006    10000   <Class::Time>#now
  6.77      0.113     0.008     0.000     0.106        1   Integer#times
  6.52      0.009     0.007     0.000     0.002    10000   Kafka::PendingMessageQueue#write
  6.22      0.010     0.007     0.000     0.003    10000   <Class::Struct::PendingMessage>#new
  5.98      0.007     0.007     0.000     0.000    20000   Kafka::Producer#buffer_size
  4.94      0.006     0.006     0.000     0.000    10000   Set#add
  4.87      0.006     0.006     0.000     0.000    10000   NilClass#to_s
  3.88      0.005     0.004     0.000     0.000    10000   Kafka::Instrumenter#instrument
  3.70      0.004     0.004     0.000     0.000    20000   String#bytesize
  3.56      0.006     0.004     0.000     0.002    10000   Time#initialize
  3.18      0.004     0.004     0.000     0.000    20000   Struct::PendingMessage#bytesize
  2.97      0.003     0.003     0.000     0.000    10000   Kafka::Producer#buffer_bytesize
  2.44      0.003     0.003     0.000     0.000    10000   Struct#initialize
  1.95      0.002     0.002     0.000     0.000    10000   String#to_s
  1.45      0.002     0.002     0.000     0.000    10000   Fixnum#+
  0.03      0.113     0.000     0.000     0.113        1   Global#[No method]
```

repeatedly:
> It may be better to simply remove the pending message queue altogether.

This is other good approach to reduce overhead.

dasch:
I'm still not quite sure it's a good idea, though – the focus of ruby-kafka is operational safety, not high performance. That means that the code base should be easy to use and that error scenarios are well defined.

repeatedly:
> That means that the code base should be easy to use and that error scenarios are well defined.

Agree so I simply update PendingMessage, not re-design produce mechanizm.

dasch:
Okay, let me merge this and then think about how the design can evolve going forward. I'd like to have fewer moving parts.

dasch:
:heart:

repeatedly:
Thanks!

PR #224, opened 2016-06-20
Require uri in case of running without bundler

### LINKED ISSUE #223: Initializing kafka.new Object
Having a problem trying to use the function Kafka.new.

Sample Code

require 'kafka'
kaf = Kafka.new(seed_brokers: ["ipaddr:9092"], client_id: "client-id")

When using the code above I get this error

.gem/ruby/2.2.3/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:409:in `block in normalize_seed_brokers': uninitialized constant Kafka::Client::URI (NameError)
At line 411 there is a reference to
uri = URI.parse(connection)

However I see no reference in any of the library files that requires the uri module. After manually setting this I was able to connect to my kafka broker.

Is this on purpose?

charkost:
After the release of v0.3.9 every kafka client tries to parse seed
brokers using URI which is not required by default without bundler.

This patch explicitly requires uri and allows the kafka client
to work without bundler.

Fixes #223

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -1,4 +1,5 @@
 require "openssl"
+require "uri"
 
 require "kafka/cluster"
 require "kafka/producer"


dasch:
:heart:

PR #232, opened 2016-06-28
Add an API for retrieving the last offset in a partition

### LINKED ISSUE #225: Retrieve Latest Offset for Partition
Hi there! I'm attempting to migrate some code from Poseidon, and one feature I can't figure out how to replicate is fetching the latest offset for a partition.

Normally, I'd fetch the `:ealiest` message in the partition and use the `highwater_mark_offset` from the response, but I can't figure out a way to do this easily. It _seems_ like that highwater mark is only accessible in the `FetchedBatch` response, and I can't figure out a way to get to it.

Both `Client#fetch_messages` and `Client#each_message` map to only the `messages` inside the batch, not the batch itself.

`Consumer#each_batch` seems to give me those batches back, but it seems crazy to go through that just to do a simple single fetch.

If all my above assumptions are correct, would you be open to a PR that adds `Client#fetch_batch`, which basically does `fetch_messages` but skips the `flat_map`?

dasch:
Add a simple API for fetching the offset of the last message in a partition:

```ruby
kafka.last_offset_for("some-topic", 42) #=> 123
```

I may want to change the API such that the partition is specified with a keyword parameter, e.g. `partition: 42`. This would allow extending the API to support multiple partitions, e.g. `partitions: [42, 42]`.

Closes #225.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -378,6 +378,19 @@ module Kafka
       @cluster.partitions_for(topic).count
     end
 
+    # Retrieve the offset of the last message in a partition. If there are no
+    # messages in the partition -1 is returned.
+    #
+    # @param topic [String]
+    # @param partition [Integer]
+    # @return [Integer] the offset of the last message in the partition, or -1 if
+    #   there are no messages in the partition.
+    def last_offset_for(topic, partition)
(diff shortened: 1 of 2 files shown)


sam-obeid:
This would be a nice feature to have! 👍

PR #239, opened 2016-07-08
Skip empty batches

### LINKED ISSUE #237: each_message for empty topic causes  undefined method `offset' for nil:NilClass

* Version of Ruby: 2.3.0
* Version of Kafka: 0.9.0.0
* Version of ruby-kafka: 0.3.9

###### Steps to reproduce

This may be a rare bug, but I just came across and wanted to file it. I'd be happy to submit a PR if the maintainers want. Basically if you call each_message on a kafka topic with no messages in it you get the following:

```ruby
kafka = Kafka.new(...)
kafka.each_message(topic: "an_empty_topic") do |message|
  puts message
end
```

###### Expected outcome

I would expect this to sit for a little bit and then not do anything because the kafka topic is empty.

###### Actual outcome

I get the following error:

```
NoMethodError: undefined method `offset' for nil:NilClass
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/fetched_batch.rb:29:in `last_offset'
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:334:in `block (2 levels) in each_message'
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:332:in `each'
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:332:in `block in each_message'
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:318:in `loop'
	from /Users/foobar/.rvm/gems/ruby-2.3.0/gems/ruby-kafka-0.3.9/lib/kafka/client.rb:318:in `each_message'
```

Basically it looks like in https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/fetched_batch.rb#L29 there is a check where we call `messages.last.offset` but we never seem to check that messages is empty. I can easily throw an extra guard in there to make sure messages isn't empty and submit a PR if that would help.

natesholland:
Previously we were erroring out in fetch_batch#last_offset because messages
would be empty so last would return nil and then nil wouldn't define offset.
I fixed this by returning -2 if messages was empty, thereby avoiding the error.

Fixes #237

### DIFF
--- a/lib/kafka/fetched_batch.rb
+++ b/lib/kafka/fetched_batch.rb
@@ -26,6 +26,7 @@ module Kafka
     end
 
     def last_offset
+      return -2 if messages.empty?
       messages.last.offset
     end
 


natesholland:
@dasch here's my proposed fix for #237

dasch:
Why not return nil instead?

natesholland:
If I return nil then I get a no method error on the nil object.

natesholland:
This was fixed by #240 which was a better solution.

PR #243, opened 2016-07-14
Enable keepalive for the sockets

dasch:
The Kafka brokers automatically kill client connections after a period of inactivity, typically 10 minutes. In order to avoid getting a network error and having to re-connect, the sockets should be kept alive.

Unfortunately, Linux and Mac OS are not consistent in the APIs they offer, so we have to use both.


tcrayford:
:+1:

dasch:
@tcrayford unfortunately this wasn't enough to keep the brokers from killing the connections, but I'm wondering whether it would make sense to enable keep-alive regardless – the Java producer uses keep-alive connections. On the other hand I'd like to have a concrete problem to solve, and the mechanism implemented in #244 should solve any problem that keep-alive would solve...

PR #244, opened 2016-07-14
Retry requests when the broker closes the connection

dasch:
Kafka brokers automatically close client connections that have been inactive for a while. If the client then tries to use the connection, a network error will be raised. In order to avoid polluting logs and monitoring systems with these errors, we can retry requests.


tcrayford:
:+1:

PR #248, opened 2016-07-26
Don't discard all consumer offset data on resync

dasch:
When a group resyncs we were discarding all local data about which partition offsets had been processed. Any data that had not been committed would be lost, even if the consumer was assigned some of the same partitions after the sync. With this change, consumers only discard offset data for partitions they are no longer assigned.

This should result in less reprocessing of messages. Running the consumer group fuzz test indeed indicates that this is the case, with the number of duplicates going from upwards of 10% to about 0.1%

#### Tasks

- [x] Documentation

### DIFF
--- a/lib/kafka/offset_manager.rb
+++ b/lib/kafka/offset_manager.rb
@@ -64,9 +64,15 @@ module Kafka
       end
     end
 
-    def clear_offsets
-      @uncommitted_offsets = 0
-      @processed_offsets.clear
+    def clear_offsets_excluding(excluded)
+      # Clear all offsets that aren't in `excluded`.
+      @processed_offsets.each do |topic, partitions|
+        partitions.keep_if do |partition, _|
+          excluded.fetch(topic, []).include?(partition)
+        end
+      end
+
+      # Clear the cached commits from the brokers.
       @committed_offsets = nil
     end
 
--- a/lib/kafka/offset_manager.rb
+++ b/lib/kafka/offset_manager.rb
@@ -64,6 +64,13 @@ module Kafka
       end
     end
 
+    def clear_offsets
+      @processed_offsets.clear
+
+      # Clear the cached commits from the brokers.
+      @committed_offsets = nil
+    end
+
     def clear_offsets_excluding(excluded)
       # Clear all offsets that aren't in `excluded`.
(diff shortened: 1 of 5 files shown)


dasch:
I'm thinking about introducing the notion of _checkpoints_ as a more clear name of offset commits.

lib/kafka/offset_manager.rb
```diff
@@ -62,9 +62,15 @@ def commit_offsets_if_necessary
...
+    def clear_offsets_excluding(excluded)
+      # Clear all offsets that aren't in `excluded`.
+      @processed_offsets.each do |topic, partitions|
+        partitions.keep_if do |partition, _|
+          excluded.fetch(topic, []).include?(partition)
+        end
+      end
+
+      # Clear the cached commits from the brokers.
       @committed_offsets = nil
```
dasch: `@processed_offsets` refers to the consumers in-memory checkpoints, updated after _every_ message. `@committed_offsets` is a cache of what the broker API returned, i.e. the _committed_ offsets.

osheroff:
There's the case of:

- consumer A loses connectivity
- consumer B takes over, processes a bunch of messages
- consumer A resumes, ends up re-processing from the old offset.

Maybe it could make a spot-check that the stored consumer position isn't ahead of it?

osheroff:
@dasch were you replying to me with: 

> @processed_offsets refers to the consumer's in-memory checkpoints, updated after every message. @committed_offsets is a cache of what the broker API returned, i.e. the committed offsets.

?

dasch:
Nah, just wanted to make the diff clear. 

I was thinking that I could maybe use the group generation id to ensure there hasn't been a gap in which another member was assigned the partitions...

dasch:
That is, we can check if `prev_generation_id + 1 == new_generation_id`. That should rule out the case you described (which was also the one I was worried about, actually).

Can you think of other scenarios that could cause problems?

dasch:
Okay, if the new generation isn't the immediate successor of the last we simply clear all the local offsets and resume from the commits.

tcrayford:
@dasch does the upstream consumer do this? If not I may make an jira for them.

dasch:
@tcrayford I'm actually not sure. This seems to significantly decrease the reprocessing rate.

PR #252, opened 2016-08-01
Also retry reading connection responses

dasch:
We're still seeing errors from idle connections being used – hopefully this will fix the problem once and for all.

### DIFF
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -145,29 +145,14 @@ module Kafka
       )
 
       data = Kafka::Protocol::Encoder.encode_with(message)
-      retried = false
       notification[:request_size] = data.bytesize
 
-      begin
-        @encoder.write_bytes(data)
-      rescue Errno::ETIMEDOUT
-        @logger.error "Timed out while writing request #{@correlation_id}"
-        raise
-      rescue Errno::EPIPE, Errno::ECONNRESET, EOFError
-        # Kafka brokers automatically close client connections after a period of
-        # inactivity. If this has happened, it's safe to re-open the connection
-        # and retry the request.
-        if retried
-          raise
-        else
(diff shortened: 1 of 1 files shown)


lib/kafka/connection.rb
```diff
@@ -196,6 +181,21 @@ def read_response(response_class, notification)
       raise
     end
 
+    # Kafka brokers automatically close client connections after a period of
+    # inactivity. If this has happened, it's safe to re-open the connection
+    # and retry the request.
+    def with_retry
+      yield
+    rescue Errno::EPIPE, Errno::ECONNRESET, EOFError
```
tcrayford: Are there equivalent errors for SSL sockets we should catch here?

dasch: I'd assume that we'd get the same errors – my understanding is that the SSL "socket" is really just two normal sockets under the hood.

dasch: And the errors would be at the TCP level.

tcrayford: Having done some digging, seems accurate.

dasch: :+1:

ctrochalakis:
Just a quick thought,

We were doing some experiments on that and when the server closes the connection we see that the EOFError is always raised in `read_response()`.

In this patch you retry the `read_response()` part but perhaps you should retry sending the **request**. Trying reading a response (in a request/response cycle) doesn't make sense as the new connection has lost context.

Maybe retrying should be moved to upper levels where we have more context on what's the better thing to do. In the consumer, for example, we should retry sending the request.

dasch:
@ctrochalakis good point. Let me think about this a bit. Currently we *do* retry requests, but not immediately and not always. I would only want to do retries at this low a level if it wasn't dangerous.

dasch:
Closing this, it doesn't make sense.

PR #253, opened 2016-08-02
Rescue EHOSTUNREACH in case of turned off broker

charkost:
EHOSTUNREACH is raised when we try to connect to a turned off broker.
We should rescue this error in order to mark cluster as stale and fetch
the metadata of the new cluster state without this broker.


dasch:
:heart:

PR #257, opened 2016-08-03
Avoid calling the ListOffsets Kafka API for each assigned partition

dasch:
We've run into problems when a consumer is assigned a _lot_ of partitions – since the current code resolves the starting position for each partition in a separate Kafka API call, the consumer would exceed the session timeout doing those calls before being able to do any work. Since it would be kicked out of the group before being ready to commit those offsets, it would have to start over, causing an infinite loop of ListOffsets calls.

This change makes each consumer execute the minimum number of API calls to resolve the starting position for _all_ partitions in a given topic. It's possible to further optimize by having just a single call to each broker that queries across all topics, but that would complicate matters further right now.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -119,10 +119,16 @@ module Kafka
       raise
     end
 
-    def resolve_offset(topic, partition, offset)
+    def resolve_offsets(topic, partitions, offset)
       add_target_topics([topic])
       refresh_metadata_if_necessary!
-      broker = get_leader(topic, partition)
+
+      partitions_by_broker = partitions.each_with_object({}) {|partition, hsh|
+        broker = get_leader(topic, partition)
+
+        hsh[broker] ||= []
+        hsh[broker] << partition
+      }
 
       if offset == :earliest
         offset = -2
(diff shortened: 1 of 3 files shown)


lib/kafka/cluster.rb
```diff
@@ -119,30 +119,48 @@ def partitions_for(topic)
...
       add_target_topics([topic])
       refresh_metadata_if_necessary!
-      broker = get_leader(topic, partition)
+
+      partitions_by_broker = partitions.inject({}) {|hsh, partition|
+        broker = get_leader(topic, partition)
+        hsh[broker] ||= []
+        hsh[broker] << partition
+        hsh
+      }
```
bquorning: I’m wondering if this code would be clearer by using `each_with_object` and `Hash.new([])`?

dasch: Gonna try. `Hash.new([])` wouldn't work, as the Array would be a shared object – I'd have to use `Hash.new {|hsh, key| hsh[key] = [] }` at which point it's simpler to just do `hsh[key] ||= []` in the block...

bquorning:
:+1:

PR #262, opened 2016-08-04
Add a minimum one second interval between delivery attempts

### LINKED ISSUE #261: Limit the rate of message deliveries in the async producer
We experienced a situation where the async producer got into a loop where it started delivering messages at a very high rate. There should be a minimal interval between each attempt in order to spare the brokers.

dasch:
If there's a sub-second delivery interval configured, use that as the minimum instead.

Fixes #261.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -57,6 +57,8 @@ module Kafka
   #     producer.shutdown
   #
   class AsyncProducer
+    # Allow at least one second between message delivery attempts.
+    MIN_DELIVERY_INTERVAL = 1
 
     # Initializes a new AsyncProducer.
     #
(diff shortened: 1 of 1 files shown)


tcrayford:
I'm -1 on this change as it stands. The original issue was about metadata requests being sent too fast, not produce requests. This limits max potential throughput, and kafka brokers can handle a *lot* of produce requests being sent one after the other just fine.

I'd rather figure out a better backoff mechanism for metadata requests in the producer than creating artificial slowdown here.

dasch:
Good point. The current issue is that the metadata is marked as stale if there's a problem with just a single partition, after which it needs to be refreshed. Maybe setting a minimum interval between refreshes and allowing the producer to use stale metadata in that interval would be a good solution?

dasch:
Something like #263.

PR #263, opened 2016-08-04
Allow at least 5 seconds between metadata requests

dasch:
In order to avoid pummeling the brokers with metadata requests in cases where e.g. a partition is unavailable and the client is waiting for the cluster to figure things out, we add a minimum interval that has to pass between two metadata requests.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -27,6 +27,9 @@ module Kafka
       @cluster_info = nil
       @stale = true
 
+      # Refresh cluster metadata at most every 5 seconds.
+      @metadata_refresh_interval = 5
+
       # This is the set of topics we need metadata for. If empty, metadata for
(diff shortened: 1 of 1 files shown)


tcrayford:
:+1:

dasch:
Darn, I'll have to make it configurable, or the tests will take forever...

dasch:
Hmm. Having to make the interval configurable just in order to make the tests pass seems wrong to me. Maybe I'll switch to a rate limiting approach, allowing up to _n_ requests before throttling.

osheroff:
seems like a good start, you may want to consider certain conditions that trigger an immediate refresh tho.  Notably the `NotLeaderForPartition` exception, which will happen after a leader re-election.

dasch:
Would those also occur when the broker has never been leader of that partition, or would that return UnknownTopicOrPartition?

PR #264, opened 2016-08-04
Reset the consumer's position when an offset is out of range

dasch:
If the committed offset for a given partition is invalid, there's not much to do except resetting to the default offset.

This could be a horrible idea. I need to do some more research in order to verify that commits can in fact become corrupted and it's not just ruby-kafka messing up right now.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -19,6 +19,7 @@ module Kafka
   end
 
   class OffsetOutOfRange < ProtocolError
+    attr_accessor :topic, :partition
   end
 
   # The request is for a topic or partition that does not exist on the broker.
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -263,6 +263,12 @@ module Kafka
       end
 
       operation.execute
+    rescue OffsetOutOfRange => e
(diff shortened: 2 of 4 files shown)


dasch:
Looks like I'm not the only one with this problem: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetOutOfRangeException.java

dasch:
@tcrayford have you run into cases where fetch requests fail with OffsetOutOfRange in the JVM consumer? I'm unsure whether there's a race in the ruby-kafka code or something.

tcrayford:
@dasch we haven't that much. They can happen during node failure under specific conditions but aren't part of normal operation in our experience.

dasch:
I see a few scenarios in which a committed offset is no longer valid:

1. The message expires.
2. Log compaction is enabled and the message is removed.
3. The topic is deleted and re-created.

These would typically only be seen if the consumer had been paused for some time – but pausing a consumer group and starting it periodically is a valid use case, I think. There can also be times when a group has to go offline in order for some change to be made.

For these reasons, it _could_ make sense to have automatic resetting of positions...

dasch:
Okay, I'll try it out and evaluate again before making the API beta.

PR #268, opened 2016-08-22
Allow pausing partitions in the consumer

dasch:
In some cases, a specific message with e.g. a malformed payload will cause a consumer's message processor to raise an exception, e.g.

```ruby
  value = JSON.parse(message.value) # <-- this could fail!
  ...
end
```

There are myriad other ways a processor could fail, and oftentimes the cause will be deterministic and based on a single message containing something unexpected.

For this reason, the official Java client allows [pausing specific partitions](https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#pause(org.apache.kafka.common.TopicPartition...)), making the consumer skip them when fetching messages. An operator would typically have to look into the matter, fix the underlying issues, and resume processing. In the meantime, processing of the _other_ partitions can continue uninterrupted.

This PR replicates that behavior in ruby-kafka. Currently, only explicitly pausing and resuming partitions is supported, but in the future it would be possible to e.g. set timeouts on pauses, allowing a partition to be automatically resumed after a while. This could be a nice fit in cases where the underlying error was not caused directly by the message, but rather by some resource referenced from the message, e.g. a shard of a database being unavailable.

The user would use the API like this:

```ruby
begin
    ...
  end
rescue Kafka::ProcessingError => e
  # Here we elect to pause the partition from which the message was read:
  consumer.pause(e.topic, e.partition)

  # Report the exception to some error aggregator:
  ErrorTracker.report(e.cause)

  # Go back into the processing loop:
  retry
end
```

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -9,6 +9,10 @@ module Kafka
     attr_reader :topic, :partition, :offset
 
     def initialize(topic, partition, offset)
+      @topic = topic
+      @partition = partition
+      @offset = offset
+
       super()
     end
   end
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -4,6 +4,15 @@ module Kafka
   class Error < StandardError
   end
 
+  # There was an error processing a message.
+  class ProcessingError < Error
+    attr_reader :topic, :partition, :offset
+
+    def initialize(topic, partition, offset)
+      super()
+    end
+  end
+
   # Subclasses of this exception class map to an error code described in the
   # Kafka protocol specification.
   #
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -52,6 +52,9 @@ module Kafka
       @session_timeout = session_timeout
       @heartbeat = heartbeat
 
+      # A list of partitions that have been paused, per topic.
+      @paused_partitions = {}
+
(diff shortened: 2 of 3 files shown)


lib/kafka.rb
```diff
@@ -4,6 +4,15 @@ module Kafka
   class Error < StandardError
   end
 
+  # There was an error processing a message.
+  class ProcessingError < Error
+    attr_reader :topic, :partition, :offset
+
+    def initialize(topic, partition, offset)
+      super()
```
andreas: Arguments are being discarded?

lib/kafka.rb
```diff
@@ -4,6 +4,15 @@ module Kafka
...
+    end
+  end
+
```
andreas: Should `ProcessingError` include the original exception?

lib/kafka.rb
dasch: Nice catch.

lib/kafka.rb
dasch: It does so automatically, I believe.

dasch: Added a spec for it.

andreas: You're right: it's been like that since 2.1 apparently. I'm not up to date 😁

dasch:
How to monitor this...

PR #272, opened 2016-08-26
Allow setting a timeout on a partition pause

dasch:
Although pausing a partition is a good way to avoid a single message causing a complete crash of a consumer, it may sometimes make sense to retry processing of the message, even if the user hasn't gone in and deployed a code change to handle the error, as sometimes the exception is due to an external resource referenced by the message that's in a temporary error state.

This change allows pausing a partition while setting a timeout for the pause. After the timeout expires, the partition is resumed.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -137,7 +137,15 @@ module Kafka
         # automatically resumed. When pausing, the timeout is translated to an
         # absolute point in time.
         timeout = partitions.fetch(partition)
-        timeout.nil? || Time.now < timeout
+
+        if timeout.nil?
+          true
+        elsif Time.now < timeout
+          true
+        else
+          resume(topic, partition)
+          false
+        end
       end
     end
 
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -100,12 +100,16 @@ module Kafka
     # idea to simply pause the partition until the error can be resolved, allowing
     # the rest of the partitions to continue being processed.
     #
+    # If the `timeout` argument is passed, the partition will automatically be
+    # resumed when the timeout expires.
+    #
     # @param topic [String]
     # @param partition [Integer]
+    # @param timeout [Integer] the number of seconds to pause the partition for.
     # @return [nil]
-    def pause(topic, partition)
-      @paused_partitions[topic] ||= Set.new
-      @paused_partitions[topic] << partition
+    def pause(topic, partition, timeout: nil)
+      @paused_partitions[topic] ||= {}
+      @paused_partitions[topic][partition] = timeout && Time.now + timeout
     end
 
     # Resume processing of a topic partition.
@@ -115,7 +119,7 @@ module Kafka
     # @param partition [Integer]
     # @return [nil]
     def resume(topic, partition)
-      paused_partitions = @paused_partitions.fetch(topic, [])
+      paused_partitions = @paused_partitions.fetch(topic, {})
       paused_partitions.delete(partition)
     end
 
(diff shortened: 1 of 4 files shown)


lib/kafka/consumer.rb
```diff
@@ -100,12 +100,16 @@ def stop
...
     # @param topic [String]
     # @param partition [Integer]
+    # @param timeout [Integer] the number of seconds to pause the partition for.
     # @return [nil]
-    def pause(topic, partition)
-      @paused_partitions[topic] ||= Set.new
-      @paused_partitions[topic] << partition
+    def pause(topic, partition, timeout: nil)
+      @paused_partitions[topic] ||= {}
+      @paused_partitions[topic][partition] = timeout && timeout + Time.now
```
bquorning: `TypeError: Time can't be coerced into Fixnum`

I think `Time.now + timeout` will work better.

bquorning:
This PR needs a test.

lib/kafka/consumer.rb
dasch: :+1:

dasch:
> This PR needs a test.

Yup.

dasch:
@bquorning added a test.

lib/kafka/consumer.rb
```diff
@@ -126,7 +130,15 @@ def resume(topic, partition)
...
     def paused?(topic, partition)
-      @paused_partitions.fetch(topic, []).include?(partition)
+      partitions = @paused_partitions.fetch(topic, {})
+
+      if partitions.key?(partition)
+        # Users can set an optional timeout, after which the partition is
+        # automatically resumed. When pausing, the timeout is translated to an
+        # absolute point in time.
+        timeout = partitions.fetch(partition)
+        timeout.nil? || Time.now < timeout
```
andreas: How about deleting the entry in the `paused_partitions`-hash if the timeout has passed?

spec/consumer_spec.rb
```diff
@@ -129,5 +131,21 @@
...
+      end
+
+      Timecop.freeze time + 29 do
+        expect(consumer.paused?("greetings", 0)).to eq true
+      end
+
+      Timecop.freeze time + 31 do
+        expect(consumer.paused?("greetings", 0)).to eq false
+      end
+    end
```
andreas: Hmm, I think I would favor a single `freeze`-block with calls  to `travel`.

lib/kafka/consumer.rb
dasch: :+1:

spec/consumer_spec.rb
dasch: When using `travel`, I think time is still moving, so it's more difficult to make assumptions.

dasch: Nah, it works as expected inside `freeze`. I'll change it.

lib/kafka/consumer.rb
```diff
@@ -126,7 +130,23 @@ def resume(topic, partition)
...
-      @paused_partitions.fetch(topic, []).include?(partition)
+      partitions = @paused_partitions.fetch(topic, {})
+
+      if partitions.key?(partition)
+        # Users can set an optional timeout, after which the partition is
+        # automatically resumed. When pausing, the timeout is translated to an
+        # absolute point in time.
+        timeout = partitions.fetch(partition)
+
+        if timeout.nil?
```
bquorning: I’m wondering if `Float::INFINITY` would help here?

dasch: Doesn't work well with `Time.now +`

PR #275, opened 2016-08-30
Automatically reopen idle connections

dasch:
Connections that have been idle for more than five minutes are automatically reopened.

### DIFF
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -1,3 +1,5 @@
+require "timecop"
+
 describe Kafka::Connection do
   let(:logger) { LOGGER }
   let(:host) { "127.0.0.1" }
@@ -35,8 +37,10 @@ describe Kafka::Connection do
 
         begin
           handle_client(client)
+        rescue EOFError
+          # Client disconnected.
         rescue => e
-          puts e
+          puts "Server crashed with error: #{e}"
           break
         ensure
           client.close
@@ -132,6 +136,18 @@ describe Kafka::Connection do
       response = connection.send_request(request)
 
       expect(response).to eq "hello!"
+    end
+
+    it "re-opens the connection if it is idle for more than 5 minutes" do
+      Timecop.freeze do
+        connection.send_request(request)
+
+        Timecop.travel(60 * 6)
+
+        response = connection.send_request(request)
+
+        expect(response).to eq "hello!"
+      end
     end
 
     it "emits a notification" do
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -27,6 +27,7 @@ module Kafka
   class Connection
     SOCKET_TIMEOUT = 10
     CONNECT_TIMEOUT = 10
+    IDLE_TIMEOUT = 60 * 5
 
     # Opens a connection to a Kafka broker.
     #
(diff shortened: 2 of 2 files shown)


spec/connection_spec.rb
```diff
@@ -134,6 +138,18 @@ def handle_client(client)
       expect(response).to eq "hello!"
     end
 
+    it "re-opens the connection if it is idle for more than 5 minutes" do
+      Timecop.freeze do
+        connection.send_request(request)
+
+        Timecop.travel(60 * 6)
+
```
bquorning: Shouldn’t you `expect` that it’s `idle?` somewhere around here?

dasch: I knew you'd say that! ;-)

I'd rather not expose the idleness stuff at this layer, but it's difficult to test otherwise. Especially since this test actually uses the full TCP stack rather than a test double socket...

I'm thinking about splitting this class into two: Session and Connection. The former would handle lifecycle events while the latter would wrap the socket and keep track of correlation ids, last-request timestamps, etc. Connection would expose methods that Session could use to make decisions.

ctrochalakis:
Hello @dasch!

What's the purpose of this change? I mean, is it better compared to handling `EOFError`?

dasch:
@ctrochalakis the problem is that the brokers in Kafka 0.9 automatically close client connections that have idle for more than 10 minutes. If you have producer instances that only seldomly produce messages, they may attempt to use such a connection, getting a network error. There's no way for the client to know the reason why the broker closed the connections, nor the amount of data the broker received first, so indiscriminate retries are probably a bad idea.

This solution simply preempts the brokers by reopening the connection client-side, using a lower timeout that the brokers.

ctrochalakis:
Do we really need to know why the server closed the connection, or can we handle EOFErrors in a uniform way? That's what the other drivers are doing. Also, there are other cases where the server can close the connection before the 10min timeout, like when we do a manual restart.

I think it would be simpler if, when we fail to read a reply from the server, we retry the request as a whole.

dasch:
@ctrochalakis can you point me to an example of how that's done in another client?

ctrochalakis:
See for example here:
https://github.com/Parsely/pykafka/blob/master/pykafka/producer.py#L449

It catches `SocketDisconnectedError` raised by `owned_broker.broker.produce_messages(req)` and retries the request. Note that this can be raised in either the write(request) or read(response) path.

`SocketDisconnectError` is raised on empty reads (server disconnects) in https://github.com/Parsely/pykafka/blob/master/pykafka/utils/socket.py#L43.

dasch:
Hmm. I'll experiment a bit more with retries. I was bit by it in the past.

dasch:
@ctrochalakis okay – ruby-kafka actually does the same kind of retries as pykafka. What this is about is low-level protocol message retries. The problem is that we actually monitor network errors at the protocol level, and the idle connection killer causes very high error rates when produce requests are sparse.

tcrayford:
@dasch interestingly, the Java clients have their own idle connection reaper whose timeout is tuned slightly under the default broker timeout - see https://issues.apache.org/jira/browse/KAFKA-1282 for the original issue.

I think doing something like that would prevent the issue here, and be safer than reopening idle connections. Thoughts? I'd be happy to work on a PR.

dasch:
@tcrayford I like that approach. Would love a PR.

dasch:
@tcrayford still up for working on a PR or should I give it a go?

tcrayford:
Go for it.

PR #276, opened 2016-08-31
Retry requests when the connection has been closed

dasch:
An alternative to #275. Always retries requests, regardless of request type.

### DIFF
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -75,25 +75,27 @@ module Kafka
     #
     # @return [Object] the response.
     def send_request(request)
-      # Default notification payload.
-      notification = {
-        broker_host: @host,
-        api: Protocol.api_name(request.api_key),
-        request_size: 0,
-        response_size: 0,
-      }
+      with_retry do
+        # Default notification payload.
+        notification = {
+          broker_host: @host,
+          api: Protocol.api_name(request.api_key),
+          request_size: 0,
+          response_size: 0,
+        }
 
-      @instrumenter.instrument("request.connection", notification) do
-        open unless open?
+        @instrumenter.instrument("request.connection", notification) do
+          open unless open?
 
-        @correlation_id += 1
+          @correlation_id += 1
 
-        write_request(request, notification)
+          write_request(request, notification)
 
-        response_class = request.response_class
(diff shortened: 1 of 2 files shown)


lib/kafka/connection.rb
```diff
@@ -75,25 +75,27 @@ def close
...
+          write_request(request, notification)
 
-        response_class = request.response_class
-        wait_for_response(response_class, notification) unless response_class.nil?
+          response_class = request.response_class
+          wait_for_response(response_class, notification) unless response_class.nil?
+        end
       end
-    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
+    rescue Errno::ETIMEDOUT => e
```
ale7714: I have a small question about this, why don't rescue all the same errors as before? 

We were seeing in of our service a high amount of connection errors:
```
 Could not connect to broker kafka-3 (node_id=301): Connection error: EOFError
``` 

And that was having a big impact on the service throughput.  I saw your PR and we tried adding this retry and the connection errors dropped completely  ❤️  .  But I kept the old rescue, since i'm not sure about the impact of changing the rescue

```
rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
```

dasch: I think that may have been a rebase artifact.

dasch: Note that those errors are most likely due to the broker killing an idle connection.

dasch: The problem here is that I'm not sure it's always _safe_ to retry a request...

ale7714: @dasch "safe" in the sense that doing this could cause an unexpected error? or something else?

dasch:
As in it could cause a side effect to happen twice.

ale7714:
@dasch what if we add an option for this to the producer? something like `immediate_retry` so by default is false. And we don't have to force this behavior on everyone. I'll be happy to open a PR for this.

dasch:
I still think it's a dangerous option – we might have to remove it in the future if we add commands that must not be retried.

Note that you shouldn't be afraid of those connection errors – they're only reported as error because the client can't see that the broker killed the connection on purpose. The built-in high-level retries should ensure that everything still works.

PR #277, opened 2016-08-31
Rescue NotCoordinatorForGroup during heartbeats

charkost:
It is possible for a group coordinator after a group rebalance to send
a group to another group coordinator. This results in NotCoordinatorForGroup
error in the next heartbeat attempt and the consumer crashes.

This commit gracefully handles this problem by getting the new group
coordinator from the cluster and retrying the heartbeat.


dasch:
:heart:

PR #278, opened 2016-08-31
Beginning of 0.10.x support - timestamps

mrb:
This Pull Request begins the work to support Kafka version 0.10.x, which has protocol changes outlined here: https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message

I'm opening this PR now as per conversations with @dasch

### DIFF
--- a/lib/kafka/fetched_message.rb
+++ b/lib/kafka/fetched_message.rb
@@ -16,12 +16,16 @@ module Kafka
     # @return [Integer] the offset of the message in the partition.
     attr_reader :offset
 
-    def initialize(value:, key:, topic:, partition:, offset:)
+    # @return [Integer] the timestamp associated with the message.
+    attr_reader :timestamp
+
+    def initialize(value:, key:, topic:, partition:, offset:, timestamp:)
       @value = value
       @key = key
       @topic = topic
       @partition = partition
       @offset = offset
+      @timestamp = timestamp
     end
   end
 end
--- a/lib/kafka/protocol/message.rb
+++ b/lib/kafka/protocol/message.rb
@@ -6,17 +6,29 @@ module Kafka
 
     # ## API Specification
     #
-    #     Message => Crc MagicByte Attributes Key Value
-    #         Crc => int32
-    #         MagicByte => int8
-    #         Attributes => int8
-    #         Key => bytes
-    #         Value => bytes
+    #   v1
+    #   Message => Crc MagicByte Attributes Key Value
+    #     Crc => int32
+    #     MagicByte => int8
+    #     Attributes => int8
+    #     Key => bytes
+    #     Value => bytes
+    #
+    #   v2 (supported since 0.10.0)
+    #   Message => Crc MagicByte Attributes Timestamp Key Value
+    #     Crc => int32
+    #     MagicByte => int8
+    #     Attributes => int8
+    #     Timestamp => int64
+    #     Key => bytes
+    #     Value => bytes
     #
     class Message
-      MAGIC_BYTE = 0
+      V1_MAGIC_BYTE = 0
+      V2_MAGIC_BYTE = 1
+      MAGIC_BYTE = V2_MAGIC_BYTE
 
-      attr_reader :key, :value, :codec_id, :offset
+      attr_reader :key, :value, :timestamp, :codec_id, :offset
 
       attr_reader :bytesize, :create_time
 
@@ -25,6 +37,7 @@ module Kafka
         @value = value
         @codec_id = codec_id
         @offset = offset
+        @timestamp = (create_time.to_f*1000.0).to_i
         @create_time = create_time
 
         @bytesize = @key.to_s.bytesize + @value.to_s.bytesize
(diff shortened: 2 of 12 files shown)


pbrisbin:
Nice!

I doesn't look like there are any changes to `FetchRequest`/`FetchResponse` here. Does that mean we'll fall into Phase 2, scenario 2.c [here](https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message#KIP-32-AddtimestampstoKafkamessage-Compatibility,Deprecation,andMigrationPlan):

> When broker sees a fetch request V1 (Supporting MessageAndOffset = V0), because the data format on disk is MessageAndOffset V1, it will NOT use the zero-copy transfer. Instead the broker will read the message from disk, down-convert them to V0 and reply using fetch response V1 with MessageAndOffset V0.

lib/kafka/fetched_message.rb
```diff
@@ -16,12 +16,16 @@ class FetchedMessage
     # @return [Integer] the offset of the message in the partition.
     attr_reader :offset
 
-    def initialize(value:, key:, topic:, partition:, offset:)
+    # @return [Integer] the timestamp associated with the message.
+    attr_reader :timestamp
+
+    def initialize(value:, key:, topic:, partition:, offset:, timestamp:nil)
```
pbrisbin: 💄 `timestamp: nil` (missing space)

dasch:
@mrb should this be closed in favor of #279?

mrb:
@dasch I'll take a look! I don't believe it handles both decoding cases, and I have some more code to push, and then we can decide.

mrb:
@dasch I'm pretty happy with this PR in its current state - going to start testing with our app today! Happy to just use the other PR or combine them somehow if that's easiest. /c @pbrisbin

dasch:
It would be nice to have a single PR. Also, I think this PR needs to handle timestamps being in milliseconds in Kafka.

mrb:
@dasch Gotcha. I've since addressed the milliseconds issue.

lib/kafka/protocol/message.rb
```diff
@@ -81,7 +104,18 @@ def self.decode(decoder)
         new(key: key, value: value, codec_id: codec_id, offset: offset)
       end
 
-      private
+      def self.v2_decode(offset, message_decoder)
+        attributes = message_decoder.int8
+        timestamp = message_decoder.int64
```
dasch: The timestamp should probably be 1) converted from ms to s, and 2) wrapped in a Time object.

spec/spec_helper.rb
```diff
@@ -1,3 +1,10 @@
+require 'simplecov'
+SimpleCov.start do
+  add_filter do |src|
+    src.filename =~ /spec/
+  end
+end
```
dasch: This seems unrelated...

dasch:
@yingnanliu can you and @mrb work out a single, shared PR for this feature? I'm going to end up closing one of them.

:heart:

yingnanliu:
@dasch Of course, we'll discuss how to merge these two PRs.
Hi @mrb, I see a major difference between these two PRs is that #278 supports decoding for both V1 and V2. IIUC, the `FetchRequest` is already set to V2, and then brokers will send V2 responses. V1 decoding is not needed in this case. What do you think?

dasch:
@yingnanliu the reason for that is that the v2 fetch API can still serve up messages written in the v1 format – the "magic byte" of the encoded message dictates the format of the message itself, regardless of the fetch API version used.

mrb:
@dasch @yingnanliu It seemed worth preserving that backward compatibility here. It's very similar to how the official Java client handles it.

pbrisbin:
When performing the update from 0.8/0.9 to 0.10 it's important to keep the log message format at the older version if there are still any older clients around. If you move directly to the 0.10 format, there have been reports of large performance issues on the brokers since they have to down-convert the messages for any older clients and can't use zero-copy transfer.

http://kafka.apache.org/documentation.html#upgrade_10_performance_impact

I _believe_ this is the V1/V2 you're talking about, which means that until we've upgrade all clients and moved the log message format up to 0.10, the clients that _are_ upgraded will have to handle V1 messages.

dasch:
@pbrisbin I believe even after the protocolversion has been set to 0.10 you'll still have to deal with messages in the old format – otherwise Kafka would have to rewrite the entire history, as the message sets are copied straight from disk to the wire under normal operation.

pbrisbin:
@dasch thanks for that clarification, that makes sense. Our data is pretty ephemeral so that's unlikely to be the case for us, but would be for anyone reading their data long-term.

Note: I think we should careful about saying "protocol version" vs "log message version" since those are two independent settings, one for inter-broker communication and one for log storage (which impacts broker-client communication).

lib/kafka/protocol/message.rb
```diff
@@ -81,7 +104,18 @@ def self.decode(decoder)
         new(key: key, value: value, codec_id: codec_id, offset: offset)
       end
 
-      private
+      def self.v2_decode(offset, message_decoder)
+        attributes = message_decoder.int8
+        timestamp = Time.at(message_decoder.int64/1000)
```
asoules: Should this divide by the float `1000.0` to preserve milliseconds?

dasch:
Closing in favor of #376.

Issue #270, opened 2016-08-24
Support for messages timestamps in 0.10.0


pbrisbin:
We are going to start using Kafka Streams and Kafka Connect for some of our message processing. This requires Kafka 0.10.0, and specifically message timestamps. We currently have a number of Ruby producers (using poseidon), which we intended to move over to this library for 0.10.0 support, but we saw that message timestamps are not supported.

Do you have any pointers for adding that functionality? There's a good chance using this library and adding that ourselves is our best option at this time. Any direction on that would be appreciated.

dasch:
As far as I know, Kafka brokers will set the timestamp themselves when the 0.9 API is used – this of course means that you don't directly control the timestamps, but it may be enough for your needs.

Unfortunately, it will be rather tricky to support both 0.9 and 0.10 simultaneously due to that way API compatibility is handled in 0.9. My current plan is to wait for 0.10 to become more widely deployed and then switch over to it completely, abandoning 0.9 support. I'd prefer to do that *after* the library has stabilized, though, in order to leave 0.9 users with something usable. It may not be until version 2.0 of ruby-kafka.

Do you have a hard requirement on user-specified message timestamps?

pbrisbin:
> Kafka brokers will set the timestamp themselves when the 0.9 API is used

Looking at [this document](https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message#KIP-32-AddtimestampstoKafkamessage-Compatibility,Deprecation,andMigrationPlan), I believe the configuration you're referring to would be *Phase 2* (0.10 brokers, 0.10 message.format.version, but 0.9 clients). There is message time-stamping, and it can be configured to be broker-provided, but for older clients, the following is happening:

- At point 1.a, you've got the added overhead of re-compression
- At point 2.c, you're unable to use zero-copy transfer

Confluent has advised us against running this way, because it has been known to cause operational issues (high CPU) on the brokers. Therefore, we're taking the stance that we can't move to the 0.10 message.format.version until we've upgraded our clients to support the 0.10 API. That means we could only get as far as *Phase 1*, and would have no message time-stamping.

We may discover our workload is such that the CPU issues won't bite us, but we decided to investigate updating our clients first -- hence this Issue.

dasch:
That sounds about right – the brokers would have to re-encode every message.

[...]

PR #279, opened 2016-09-01
0.10 Timestamp Support

yingnanliu:
This is our take on v 0.10.x support (message timestamps).

It basically implemented what @dasch mentioned in https://github.com/zendesk/ruby-kafka/issues/270#issuecomment-242332398, plus `FetchRequest` also has `api_version` set to 2.

### DIFF
--- a/lib/kafka/protocol/message.rb
+++ b/lib/kafka/protocol/message.rb
@@ -6,15 +6,16 @@ module Kafka
 
     # ## API Specification
     #
-    #     Message => Crc MagicByte Attributes Key Value
+    #     Message => Crc MagicByte Attributes Timestamp Key Value
     #         Crc => int32
     #         MagicByte => int8
     #         Attributes => int8
+    #         Timestamp => int64, in ms
     #         Key => bytes
     #         Value => bytes
     #
     class Message
-      MAGIC_BYTE = 0
+      MAGIC_BYTE = 1
 
       attr_reader :key, :value, :codec_id, :offset
 
@@ -71,6 +72,7 @@ module Kafka
         end
 
         attributes = message_decoder.int8
+        timestamp = message_decoder.int64
         key = message_decoder.bytes
         value = message_decoder.bytes
 
@@ -78,7 +80,7 @@ module Kafka
         # attributes.
         codec_id = attributes & 0b111
 
-        new(key: key, value: value, codec_id: codec_id, offset: offset)
+        new(key: key, value: value, codec_id: codec_id, offset: offset, create_time: Time.at(timestamp/1000))
       end
 
       private
@@ -102,6 +104,7 @@ module Kafka
 
         encoder.write_int8(MAGIC_BYTE)
         encoder.write_int8(@codec_id)
+        encoder.write_int64((@create_time.to_f*1000).to_i)
         encoder.write_bytes(@key)
         encoder.write_bytes(@value)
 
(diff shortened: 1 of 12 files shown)


dasch:
:heart:

dasch:
I think this looks good. I'd like to park it here until I have a better sense of when ruby-kafka will start officially supporting the new Kafka 0.10 protocol, at which point I will probably abandon support for Kafka 0.9.

lib/kafka/protocol/message.rb
```diff
@@ -71,14 +72,15 @@ def self.decode(decoder)
...
+        timestamp = message_decoder.int64
         key = message_decoder.bytes
         value = message_decoder.bytes
 
         # The codec id is encoded in the three least significant bits of the
         # attributes.
         codec_id = attributes & 0b111
 
-        new(key: key, value: value, codec_id: codec_id, offset: offset)
+        new(key: key, value: value, codec_id: codec_id, offset: offset, create_time: Time.at(timestamp/1000))
```
asoules: This should divide by a float in order to preserve milliseconds (ie. `1000.0`)

For example:
```
$ irb
$ timestamp = 1473867722801 # milliseconds
 => 1473867722801
$ Time.at(timestamp/1000).strftime('%S.%N')
 => "02.000000000"
$ Time.at(timestamp/1000.0).strftime('%S.%N')
 => "02.801000118"
```

yingnanliu: Thanks! It's fixed.

thijsc:
It should be possible to support both 0.9 and 0.10 at the same time right?

dasch:
@thijsc not easily, I'm afraid. Only in version 0.10 did Kafka introduce a way for clients to discover broker versions, and even then it's a bit arduous.

thijsc:
Right. I do know other clients offer a configuration option for this. Is that an option you would consider?

dasch:
Not at the moment - the other clients do this by having the user configure the protocol version manually, which I think is problematic. Once ruby-kafka moved to only support 0.10+ sometime in the future I will revisit this.

niels-s:
@dasch any news on when this can be merged?

dasch:
Once I'm in a position to thoroughly test ruby-kafka against 0.10 I'll create a stable branch for 0.9 support and move active development to 0.10. Hopefully within the next few months.

hoanhp:
can anyone show me the proper config for that??

dasch:
@hoanhp I'm not sure what you mean...

hoanhp:
@dasch  sorry for my poor English.  I mean can you tell me how to custom kafka-broker's config to add timestamp into each message??

dasch:
@hoanhp it's not currently supported by ruby-kafka – it will be in a future release targeting Kafka 0.10+.

dasch:
@yingnanliu can you rebase this on the latest master?

I've switched master development to target Kafka 0.10 – compatibility with Kafka 0.9 will happen in the `v4.0-stable` branch, and only include bug fixes and security patches.

lib/kafka/protocol/produce_response.rb
```diff
@@ -43,13 +45,16 @@ def self.decode(decoder)
               partition: decoder.int32,
               error_code: decoder.int16,
               offset: decoder.int64,
+              timestamp: decoder.int64,
```
dasch: Maybe this should also be converted to a Time object?

dasch:
@yingnanliu I hope it's okay if I take over this branch – I'd like to get this merged sooner rather than later.

Issue #280, opened 2016-09-05
Committed offsets are shifted by one


tulios:
Hi, after using the consumer for a while I've caught a difference from other libraries. In my company, we use [burrow](https://github.com/linkedin/Burrow) to monitor our kafka consumers (a lag checker from linkedin) and with ruby-kafka it reports one message lagging in each partition. I noticed it because we have some topics with seasonal load, topics that receive messages in one period of the day and not constantly, and this is really clear on those.

I have started debugging burrow to see if the problem was there but after some checks I concluded that burrow was correct.

* Version of Ruby:
2.3.1

* Version of Kafka:
0.9.0.1

* Version of ruby-kafka:
0.3.15.beta3

###### Steps to reproduce

Regular ruby-kafka consumer group

```ruby
require 'kafka'
kafka = Kafka.new(seed_brokers: ['localhost:9092'])
consumer = kafka.consumer(group_id: 'test1')
trap('QUIT') { consumer.stop }

consumer.subscribe('test')
consumer.each_message { |message| puts(message.inspect) }
```

###### Expected outcome

Comparing ruby-kafka and kafkacat consumers, for example, we have the following results:

Using a new topic with two partitions and publishing one message (the first message)

burrow topic offsets from `/v2/kafka/local/topic/topic-name`
```json
{
  "error": false,
  "message": "broker topic offsets returned",
  "offsets": [
    0,
    1
  ]
}
```

kafkacat consumer group offsets:
(running with `kafkacat -C -b localhost:9092 -o beginning -G kafkacat topic-name`)
burrow consumer group offsets from `/v2/kafka/local/consumer/kafkacat/topic/topic-name`

```json
{
  "error": false,
  "message": "consumer group topic offsets returned",
  "offsets": [
    -1,
    1
  ]
}
```

###### Actual outcome

ruby-kafka consumer group offsets:
burrow consumer group offsets from `/v2/kafka/local/consumer/ruby-kafka/topic/topic-name`

```json
{
  "error": false,
  "message": "consumer group topic offsets returned",
  "offsets": [
    -1,
    0
  ]
}
```

The kafka documentation says: "The committed offset should always be the offset of the next message that your application will read. Thus, when calling commitSync(offsets) you should add one to the offset of the last message processed."

https://kafka.apache.org/090/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html in "Manual Offset Control"

[...]

PR #281, opened 2016-09-05
Ensures that kafka has the offset off the next message

tulios:
This PR fixes issue #280 

@dasch I tried to follow your recommendation to move +1 to `commit _offsets` but `offset_manager#next_offset_for` is involved in two scenarios:

1) When `@processed_offsets` is not initialized and it reads from kafka
2) While the batch is not committed, thus depending on `@processed_offsets` to get the next offset

+1 is in `offset_manager#mark_as_processed` now to keep all the logic inside the offset manager.

### DIFF
--- a/spec/offset_manager_spec.rb
+++ b/spec/offset_manager_spec.rb
@@ -72,7 +72,7 @@ describe Kafka::OffsetManager do
 
       expected_offsets = {
         "x" => {
-          0 => kind_of(Numeric),
+          0 => 43,
         }
       }
 
--- a/spec/offset_manager_spec.rb
+++ b/spec/offset_manager_spec.rb
@@ -26,8 +26,8 @@ describe Kafka::OffsetManager do
 
       expected_offsets = {
         "greetings" => {
-          0 => 42,
-          1 => 13,
+          0 => 43,
+          1 => 14,
         }
       }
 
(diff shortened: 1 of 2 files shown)


spec/offset_manager_spec.rb
```diff
@@ -80,7 +72,7 @@
 
       expected_offsets = {
         "x" => {
-          0 => 42,
+          0 => kind_of(Numeric),
```
dasch: Why not an explicit value here?

tulios: Can be, fixing

dasch:
:heart:

dasch:
This will mean that existing applications will lose a message per partition upon upgrading ruby-kafka, but I'd prefer to let that happen now rather than later, seeing as the Consumer API is still alpha.

:heart: :heart: :heart:

dasch:
@tulios can you deploy this commit of `master` to your infrastructure and verify that it's working?

tulios:
Sure, I'll do it now. Regarding the "lose one message" it will be fine because the consumer offset is committed without +1 so the consumers will reprocess the last message instead.

dasch:
Ah, you're right.

tulios:
It's working great, before the PR  you can see the flat lines indicating the lag.

before PR281
pre_pr281

after PR281
after_pr281

dasch:
:heart:

dasch:
Is that Librato? If so, would you be interested in adapting the [Datadog metrics reporter](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/datadog.rb) to Librato?

PR #283, opened 2016-09-22
fix ssl socket timeout

tcrayford:
We've been seeing a case where Kafka won't accept new SSL connections
properly. It accepts the TCP one, but doesn't perform the SSL handshake
properly. The SSL socket behaviour in this library doesn't handle this
case properly.

This clears up the logic in this class as well to use a
`select_with_timeout` method (taken from
https://github.com/excon/excon/blob/192bb842e70b550ed6d78c6d621962ee0faca4be/lib/excon/socket.rb)

I've tested this against a broken production Kafka and this does indeed get a proper timeout.

### DIFF
--- a/lib/kafka/ssl_socket_with_timeout.rb
+++ b/lib/kafka/ssl_socket_with_timeout.rb
@@ -23,6 +23,7 @@ module Kafka
       addr = Socket.getaddrinfo(host, nil)
       sockaddr = Socket.pack_sockaddr_in(port, addr[0][3])
 
+      @connect_timeout = connect_timeout
       @timeout = timeout
 
       @tcp_socket = Socket.new(Socket.const_get(addr[0][0]), Socket::SOCK_STREAM, 0)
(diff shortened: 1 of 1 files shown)


dasch:
:heart:

tcrayford:
@dasch can we get a release with this fix in it?

dasch:
Unfortunately I'm on leave and won't be back until December – since I haven't been able to find anyone willing to co-maintain the library in the only one with push access :-/

Not sure if it's possible to add a maintainer without terminal access...

dasch:
@tcrayford I've released v0.3.16.beta1 – I'd love for you guys to give it a spin before I release a final version.

tcrayford:
@dasch we've been running with the commit 39dd97c5a793de81357567936bb1f41f8cf567dc (which includes this PR) in production for aages now, I think it'll be fine.

dasch:
:heart:

dasch:
I'll probably cut a final release this week, then.

PR #284, opened 2016-09-30
don't re-commit offsets on idle.

osheroff:
Since we never clear @processed_offsets, we should check that we're actually moving forward on a partition before bothering to commit it.

### DIFF
--- a/lib/kafka/offset_manager.rb
+++ b/lib/kafka/offset_manager.rb
@@ -50,14 +50,22 @@ module Kafka
     end
 
     def commit_offsets
-      unless @processed_offsets.empty?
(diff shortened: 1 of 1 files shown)


dasch:
Is the rate of commits causing problems?

osheroff:
no, just kinda annoying.  at idle all my consumers re-commit the same offset over and over and over again.  I don't know what the cost is server side, but I'd imagine it's non-zero with a decent amount of consumers.

osheroff:
(man, that was the most wishy-washy statement ever.  "i'd imagine it's non-zero".  sheesh.)

dasch:
I'd rather not mess too much with the checkpointing code while I'm away - it's been a source of many a bug. Let me look at this when I'm back.

PR #291, opened 2016-11-01
Add host and port to BrokerPool cache key

djiang:
### Problem
In cases where a Kafka broker goes down and new broker with a different host/port spins up with the same broker id, we don't want to use the previously cached broker because the connection is still pointing to the old host/port.

### Solution
To fix this, we'll consider the host and port when deciding whether or not to use the cached broker. If the cached broker is no longer valid, we'll disconnect and close the connection.


dasch:
:heart:

Could you add a unit test for this?

djiang:
Hey @dasch, I added some `BrokerPool` specs 😃

.ruby-version
```diff
@@ -0,0 +1 @@
+2.2.3
```
dasch: Can you remove this file from the PR? It doesn't seem related.

dasch:
It would also be great if you could give this branch a test in a production environment for a while before I merge, as I'm unable to do so myself for the time being. This change is rather risky.

dasch:
@osheroff could you also take a look? I won't have much internet the next few days.

.ruby-version
djiang: Sure thing

djiang:
I'll keep my branch running for my project over the weekend.

djiang:
I think I found some issues -- I'll investigate and report back

dasch:
Thanks!

djiang:
The issue seems to lie in the cluster not refreshing the metadata when it gets into certain states. The fix I have in place works, but it's a bit defensive. I want to spend some more time to get to the root cause of the problem.

dasch:
@djiang could that be related to #294?

djiang:
Hmm I don't think #294 is related as it pertains to the producer and the issue here is in the consumer.

I'll dig more into this issue soon -- I haven't had a lot of time recently.

dasch:
@djiang do you have an update on this?

dasch:
@djiang any update, or should I close this issue?

djiang:
We can close this for now -- haven't had a chance to revisit.

dasch:
👍

Issue #293, opened 2016-11-09
Failed to assign partitions


ahsandar:
my concern is related to the error message `Failed to assign partitions to X messages`, I am using fluentd-kafka plugin which uses ruby-kafka , I have an issue explaing it here https://github.com/fluent/fluent-plugin-kafka/issues/97, as the message is not very descriptive its hard to identify the cause being `fluentd`, `kafka` or `ruby-kafka` gem itself , using `kafka 0.10.1.0` compiled for `scala 2.11` the stack trace for the exception is as follows printed by `fluentd ` for `ruby-kafka`

```
2016-11-08 17:10:00 +0800 [warn]: Send exception occurred: Failed to assign partitions to 1 messages
2016-11-08 17:10:00 +0800 [info]: initialized kafka producer: kafka
2016-11-08 17:10:00 +0800 [warn]: temporarily failed to flush the buffer. next_retry=2016-11-08 17:10:00 +0800 error_class="Kafka::DeliveryFailed" error="Failed to as
sign partitions to 1 messages" plugin_id="object:3f931afe1d9c"
  2016-11-08 17:10:00 +0800 [warn]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/ruby-kafka-0.3.15/lib/kafka/producer.rb:323:in `deliver_messages_with_retries'
  2016-11-08 17:10:00 +0800 [warn]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/ruby-kafka-0.3.15/lib/kafka/producer.rb:243:in `block in deliver_messages'
  2016-11-08 17:10:00 +0800 [warn]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/ruby-kafka-0.3.15/lib/kafka/instrumenter.rb:21:in `call'
  2016-11-08 17:10:00 +0800 [warn]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/ruby-kafka-0.3.15/lib/kafka/instrumenter.rb:21:in `instrument'
  2016-11-08 17:10:00 +0800 [warn]: /opt/td-agent/embedded/lib/ruby/gems/2.1.0/gems/ruby-kafka-0.3.15/lib/kafka/producer.rb:236:in `deliver_messages'
[...]

PR #294, opened 2016-11-10
Update producer.rb

ahsandar:
Referring to these 2  issue 
https://github.com/zendesk/ruby-kafka/issues/293
 https://github.com/fluent/fluent-plugin-kafka/issues/97

from the looks need to refresh metadata when failed to assign partition, as producer can't rebalance due to broker imbalance, hence metadata needs to be refreshed.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -320,6 +320,8 @@ module Kafka
       end
 
       unless @pending_message_queue.empty?
+        # Mark the cluster as stale in order to force a cluster metadata refresh.
+        @cluster.mark_as_stale!
         raise DeliveryFailed, "Failed to assign partitions to #{@pending_message_queue.size} messages"
       end
 


dasch:
@ahsandar thanks for this PR, and sorry for the late reply – I just returned from a vacation. This looks good – would you be able to test the code in production before I merge it? There can be tricky interactions when it comes to the metadata refresh cycle.

ahsandar:
would be real difficult to get it check as we moved from staging to production during this time and unless fluentd plugin doesn't updates can't test it in an environment any time soon

ahsandar:
@dasch  can this be merged anytime soon ? would really help in this  scenario (https://github.com/fluent/fluent-plugin-kafka/issues/97) from my understanding 

```
when a broker falls out of the quorm o goes down and the leader on the topic changes but fluentd 
doesn't updated the metadata to resolve this , every time this happens it should update the 
metadata and start sending message to the new leader and once the broker is brought back up and 
leader election is performed the metadata will refresh again following the cycle
```
 what do you think  ?

dasch:
Okay, I'll merge it – but it would be great if the fluent plugin could help test this out.

ahsandar:
@dasch thanks i will send a PR for fluentd plugin and will then update at my end to test our setup.

dasch:
:heart:

ahsandar:
can you please create a tag next in sequence to `0.3.16.beta1` so i can update the plugin accordingly, thanks

dasch:
@ahsandar I've released version 0.3.16.beta2 just now.

ahsandar:
thanks

ahsandar:
happy to report that the fix works 
metadata-refresh

dasch:
:heart:

Cutting a new production version now.

PR #301, opened 2016-12-02
WIP: async-mode for the consumer

osheroff:
FetchOperation will perform multiple requests to brokers in parallel,
returning from #execute as soon as data from one broker is available.


lib/kafka/connection.rb
```diff
@@ -72,9 +72,10 @@ def close
     #
     # @param request [#encode, #response_class] the request that should be
     #   encoded and written.
+    # @param async if true, do not wait for the response
     #
     # @return [Object] the response.
-    def send_request(request)
+    def send_request(request, async: false)
```
dasch: Maybe just split out a separate `send_async_request` that is called by `send_request`, after which it would `wait_for_response`.

lib/kafka/consumer.rb
```diff
@@ -340,12 +340,12 @@ def fetch_batches(min_bytes:, max_wait_time:)
...
           else
             @logger.debug "Fetching batch from #{topic}/#{partition} starting at offset #{offset}"
-            operation.fetch_from_partition(topic, partition, offset: offset, max_bytes: max_bytes)
+            @operation.fetch_from_partition(topic, partition, offset: offset, max_bytes: max_bytes)
           end
         end
       end
 
-      operation.execute
+      @operation.execute
```
dasch: I think memoizing the operation opens us up to a range of new problems. For instance, the arguments to `fetch_batches` would be ignored on all but the first call...

Is there an alternative?

dasch:
@osheroff I've created #303 as a simpler implementation. It still waits for all the fetches to succeed before yielding anything to the user, so it's obviously _less_ efficient, but the simplicity might be worth it, at least for now. What do you think?

osheroff:
if you just kick everything off at once, but still wait for all requests to finish, you've improved your latency from n*max_wait_time to just max_wait_time, but that's about it; ie if you have any broker not receiving messages you're waiting 5 seconds for any data.  The two methods that will actually work:

- async IO
- spawn a synchronous fetch-thread per broker, which pushes into a queue

either way, I don't think there's any avoiding having a "some magic going on in the background" setup.

lib/kafka/consumer.rb
osheroff: nope, having the operation live outside the scope of the single call is utterly key; you want to return data to the client as soon as it's available, meaning the other requests need to live on in the background.

I was shooting for "minimum lines of churn" here, just to show you the idea/approach without going full-bore clean-code; in a real PR we'd probably have to move min_bytes and max_wait_time to the initializer.

dasch:
> if you just kick everything off at once, but still wait for all requests to finish, you've improved your latency from n*max_wait_time to just max_wait_time

I agree – but it's still a big improvement. I'm mostly worried about adding complexity to the already overly complex consumer code. Is this level of performance necessary? Are you running into scalability issues?

lib/kafka/consumer.rb
dasch: In that case I'd probably want to create a new type of operation altogether...

Would you be okay with reviewing the current consumer code and working with me to overhaul it? I'm not completely satisfied with it, and I'd prefer to do this right. The Java client uses futures and async IO, so maybe it would be a good idea to continue down this route.

dasch: Hmm... couldn't the state live on the individual connections rather than the operation? Calling `fetch_batches` would register a request for a specific correlation id on each connection. Blocking on a future would mean waiting until there's a response with that correlation id on the connection. We could make some of the Connection methods public in order to facilitate this.

osheroff:
I mean waiting 5 seconds for a message is pretty much unacceptable in all scenarios.  I can easily imagine a cluster in which someone has setup a topic with 20 partitions, but is only writing to 5 actively (our own maxwell topic has properties similar to this).  In this case the consumer is going to be totally unusable.  So it's not an issue, to me, of trade-offs and added complexity, it's an issue of correctness.

The threaded model of the consumer is probably going to be much cleaner code-wise, although forces the client to start/stop things.

dasch:
I think it depends on the use case – in some cases blocking a bit is fine, and you can configure `max_wait_time` as you want. However, if we have actual use cases where this is not good enough then we need to work on it.

PR #302, opened 2016-12-03
throw proper topic errors when starting to consume a topic

osheroff:
this actually only throws the *first* topic error that we see, but it's close enough; I was getting an odd "couldn't find group co-ordinator" error, when in fact I had mis-configured my cluster and should have been getting a "no leader" error.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -46,6 +46,10 @@ module Kafka
         @target_topics.merge(new_topics)
 
         refresh_metadata!
+
+        cluster_info.topics.each do |topic|
+          Protocol.handle_error(topic.topic_error_code)
+        end
       end
     end
 


lib/kafka/cluster.rb
```diff
@@ -46,6 +46,10 @@ def add_target_topics(topics)
         @target_topics.merge(new_topics)
 
         refresh_metadata!
+
+        cluster_info.topics.each do |topic|
+          Protocol.handle_error(topic.topic_error_code)
+        end
```
dasch: I think this should either go in `refresh_metadata!` or only check the topics that are being added.

dasch: Also, this could potentially break user code in some cases where the Kafka error is transient... I think the current error handling code handles recovery, but this snippet does not.

dasch: Indeed, it does break in the specs. LeaderNotAvailable is raised until the leader is, well, available. But the user code should not break because of that.

osheroff: sure, I can buy that, but presumably some other bit of the code should rescue transient errors and retry?

dasch: Yeah, but that currently happens in the consumer loop, not the setup step, as far as I remember.

dasch: The problem is that "no leader available" is _not_ necessarily an error – it could just be a transient issue. Logging is really the only proper way to handle that situation, aside from using e.g. a timeout to crap out after some time, which I'd prefer not to do.

PR #303, opened 2016-12-05
Fetch messages concurrently from all brokers

dasch:
The consumer will write a request to each broker it needs messages from before reading any responses. It will then block on the socket of the first broker in the list until there's a response, after which it will start processing messages; then continue with the next broker; etc.

This is by no means the optimal design, but it balances simplicity and efficiency.

### DIFF
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -2,10 +2,40 @@ require "stringio"
 require "kafka/socket_with_timeout"
 require "kafka/ssl_socket_with_timeout"
 require "kafka/protocol/request_message"
+require "kafka/protocol/null_response"
 require "kafka/protocol/encoder"
 require "kafka/protocol/decoder"
 
 module Kafka
+
+  # An asynchronous response object allows us to deliver a response at some
+  # later point in time.
+  #
+  # When instantiating an AsyncResponse, you provide a response decoder and
+  # a block that will force the caller to wait until a response is available.
+  class AsyncResponse
+    # Use a custom "nil" value so that nil can be an actual value.
+    MISSING = Object.new
+
+    def initialize(decoder, &block)
+      @decoder = decoder
+      @block = block
+      @response = MISSING
+    end
+
+    # Block until a response is available.
+    def call
+      @block.call if @response == MISSING
+      @response
+    end
+
+    # Deliver the response data.
+    #
+    # After calling this, `#call` will returned the decoded response.
+    def deliver(data)
+      @response = @decoder.decode(data)
+    end
+  end
 
   # A connection to a single Kafka broker.
   #
@@ -75,6 +105,18 @@ module Kafka
     #
     # @return [Object] the response.
     def send_request(request)
+      # Immediately block on the asynchronous request.
+      send_async_request(request).call
+    end
+
+    # Sends a request over the connection.
+    #
+    # @param request [#encode, #response_class] the request that should be
+    #   encoded and written.
+    #
+    # @return [AsyncResponse] the async response, allowing the caller to choose
+    #   when to block.
+    def send_async_request(request)
       # Default notification payload.
       notification = {
         broker_host: @host,
@@ -83,15 +125,41 @@ module Kafka
         response_size: 0,
       }
 
-      @instrumenter.instrument("request.connection", notification) do
-        open unless open?
+      @instrumenter.start("request.connection", notification)
 
-        @correlation_id += 1
+      open unless open?
 
-        write_request(request, notification)
+      @correlation_id += 1
 
-        response_class = request.response_class
-        wait_for_response(response_class, notification) unless response_class.nil?
+      write_request(request, notification)
+
+      response_class = request.response_class
+      correlation_id = @correlation_id
+
+      if response_class.nil?
+        async_response = AsyncResponse.new(Protocol::NullResponse) { nil }
+
+        # Immediately deliver a nil value.
+        async_response.deliver(nil)
+
+        @instrumenter.finish("request.connection", notification)
+
+        async_response
+      else
+        async_response = AsyncResponse.new(response_class) {
+          # A caller is trying to read the response, so we have to wait for it
+          # before we can return.
+          wait_for_response(correlation_id, notification)
+
+          # Once done, we can finish the instrumentation.
+          @instrumenter.finish("request.connection", notification)
+        }
+
+        # Store the asynchronous response so that data can be delivered to it
+        # at a later time.
+        @pending_async_responses[correlation_id] = async_response
+
+        async_response
       end
     rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
       close
@@ -115,6 +183,9 @@ module Kafka
 
       # Correlation id is initialized to zero and bumped for each request.
       @correlation_id = 0
+
+      # The pipeline of pending response futures must be reset.
+      @pending_async_responses = {}
     rescue Errno::ETIMEDOUT => e
       @logger.error "Timed out while trying to connect to #{self}: #{e}"
       raise ConnectionError, e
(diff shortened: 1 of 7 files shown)


lib/kafka/connection.rb
```diff
@@ -99,6 +99,34 @@ def send_request(request)
...
+        broker_host: @host,
+        api: Protocol.api_name(request.api_key),
+        request_size: 0,
+        response_size: 0,
+      }
+
+      @instrumenter.instrument("async_request.connection", notification) do
+        open unless open?
+
+        @correlation_id += 1
```
dasch: There's an obvious race condition here, of course. We'd need to register the response futures on the connection and deliver the actual responses to them if the blocking reads happen in the wrong order...

dasch:
@osheroff I've implemented very basic pipelining support in Connection.

lib/kafka/connection.rb
```diff
@@ -172,23 +230,30 @@ def read_response(response_class, notification)
...
+        correlation_id, response = read_response(response_class, expected_correlation_id, notification)
+
+        if correlation_id < expected_correlation_id
+          # There may have been a previous request that timed out before the client
+          # was able to read the response. In that case, the response will still be
+          # sitting in the socket waiting to be read. If the response we just read
+          # was to a previous request, we deliver it to the pending async response
+          # future.
+          async_response = @pending_async_responses.delete(correlation_id)
+          async_response.deliver(response) if async_response
```
osheroff: raise here if no async_response was found?

osheroff: hmmm, you can pipeline requests, but you have to block on them in the order you sent them in; maybe there's a bit deeper on an abstraction here where you send multiple requests and then get back a single handle that will resolve each request in turn?  Something like:

```
batch = broker.start_batch
topics_by_broker.each { |t|
   batch.fetch_messages_async(...)
}
batch.each_response { |response|

}
```

just  a thought.

dasch: > raise here if no async_response was found?
The original request may not have been async...

> you can pipeline requests, but you have to block on them in the order you sent them in
I thought this design fixed that limitation – if you asynchronously send A and B, then block on B, the connection read loop will deliver A first, then B. Calling A after B should still work, no?

dasch: The next step would be to not call the futures in-order, but rather loop through them a look for ones that are `ready?` (same as your code, using `select`) a call them first.

dasch: Hmm... I think it could actually be possible to use `IO.select` to pick the first future that's available:

```ruby
class IoFuture
  def initialize(io, &block)
    @io, @block = io, block
  end

  def to_io # IO.select calls this
    @io
  end

  def value
    @value ||= @block.call
  end
end

futures = [IoFuture.new(socket1) { ... }, IoFuture.new(socket2) { ... }]

until futures.empty?
  ready, _, _ = IO.select(futures, nil, nil, timeout)
  ready.each do |future|
    futures.delete(future)
    response = future.value
    ...
  end
end
```

dasch:
There's a bug in the implementation – when there's a response pending and you do another request, the latter request's decoder will be used to decode the response to the first request... I probably need to separate "wait for a response to this correlation id" and the decoding step...

lib/kafka/connection.rb
osheroff: > I thought this design fixed that limitation – if you asynchronously send A and B, then block on B, the connection read loop will deliver A first, then B. Calling A after B should still work, no?

i think that you'd block on B, it'd deliver A, then you'd block on A, but when it went to deliver B it'd fail the `correlation_id > expected_correlation_id` check.  I *think*.

osheroff: that code looks like it'd work, though I'm not clear on the whole context of it. 

I'd be tempted to try to leave `Connection` mostly as is, except break the link between send/recv.  Then maybe a `MultiConnection` or `MultiRequest` class that deals with juggling a bunch of in-flight requests, including selecting whichever one is ready. 

errrrrr not sure how that'd mix in to the synchronous world though.  Seems very tough to mix/match async/sync requests.

dasch: Yeah – there needs to be a direct mapping of a connection's correlation ids to the pending response future, and that number is reset when the connection is reopened. I think it's better to keep everything in Connection for now. As for the blocking – there should be no blocking if the future has already been delivered, so you wouldn't block on A at all.

dasch:
@osheroff since `FetchOperation#execute` returns an Enumerator that lazily blocks, this should represent a big improvement.

What as see as further improvements to this branch:

1. Rather than going through (and blocking on) responses in-order, pick responses that are ready first.
2. After completing processing of a response's batches, asynchronously send a request to the broker for the next batches before continuing with the other responses.

Does this sound reasonable to you?

dasch:
I'm still thinking about doing a re-write of the entire Consumer fetch code, but I'd like to explore a bit first.

osheroff:
> After completing processing of a response's batches, asynchronously send a request to the broker for the next batches before continuing with the other responses.

Would that outbound request be eligible to respond inside the same loop?  In other words, if I have 1 busy broker and 2 empty brokers, can I get 4 messages back from the busy broker while still waiting for the "no data" response from the 2 empty brokers?

dasch:
@osheroff in theory, yes. This is sort of a hack on top of the existing design – I imagine a optimal solution would be more "streaming" in nature: the consumer would send messages to the fetcher over a queue, the "fetcher" would select connections with responses ready, read the responses, and write them back over another queue. The fetcher could `select` with a small timeout, allowing new pending connections to be added to the IOs being selected from.

dasch:
Actually, this needn't be consumer specific – the producer has the same issue right now, with each produce request blocking sequentially. I think some abstraction on top of Connection would be best.

dasch:
The Java client seems to use these classes:

* https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
* https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/InFlightRequests.java
* https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/network/Selectable.java
* https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/network/Selector.java

osheroff:
yeah it's tricky to get just right; hard to predict the client's use cases.  The java client does roughly the following, fwiw:

- both producer and consumer are backed by hidden threads, requiring `.close` to be called.  
- the producer gives you an on-complete callback
- the consumer used to be of a style where you ceded flow-control to it.  Now they have a separate thread that does the fetching, but it's up to the caller to tick the state machine along by calling `.poll` when it's ready for more rows.

dasch:
The ruby-kafka async producer is already using a hidden thread, but it does add some complexity. I'd prefer to keep things single-threaded for now, especially since the Kafka processing model is synchronous.

A sprinkle of pipelining and async IO should be enough. I'll try to see if I can copy the concept of Selector to ruby-kafka – it seems like a decent abstraction. I'd have to change the operation API, though, as we currently call synchronous methods on Broker.

osheroff:
yeah.  if you can make it work, 1 thread is better than 2.

dasch:
@osheroff would your team have time to work on this or are you swamped? I'm currently rather busy with other stuff.

lib/kafka/connection.rb
```diff
@@ -156,8 +227,8 @@ def write_request(request, notification)
     #   a given Decoder.
     #
     # @return [nil]
-    def read_response(response_class, notification)
-      @logger.debug "Waiting for response #{@correlation_id} from #{to_s}"
+    def read_response(expected_correlation_id, notification)
```
tcrayford: this correlation id stuff means that we can still send consumer group related requests to the broker over the same channel, right?

lib/kafka/connection.rb
```diff
@@ -75,6 +105,18 @@ def close
     #
     # @return [Object] the response.
     def send_request(request)
+      # Immediately block on the asynchronous request.
+      send_async_request(request).call
```
tcrayford: I really like how the sync request pattern is just the same as the async, with an immediate blocking call. Nice!

dasch: ❤️

lib/kafka/connection.rb
dasch: Yup, at least that's the design. I ran this in production for a good hour and didn't see any problems, but I'm going to be doing at least one beta release before a final.

Issue #238, opened 2016-07-08
Specs don't run on OS X 10.11.5: docker error


natesholland:
If this is a bug report, please fill out the following:
- Version of Ruby: 2.2.3
- Version of Kafka: 0.9.0.0
- Version of ruby-kafka: 0.3.9 (master)

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.
###### Steps to reproduce

On OS X with the new docker beta running execute the command `rake spec` or `rake`
###### Expected outcome

According to the README the specs should run without a problem.
###### Actual outcome

Instead I get this error trace:

``` ruby
Users/nateh/.rvm/rubies/ruby-2.2.3/bin/ruby -I/Users/nateh/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.4.1/lib:/Users/nateh/.rvm/gems/ruby-2.2.3/gems/rspec-support-3.4.1/lib /Users/nateh/.rvm/gems/ruby-2.2.3/gems/rspec-core-3.4.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
ifconfig: interface docker0 does not exist
/Users/nateh/.rvm/gems/ruby-2.2.3/gems/docker-api-1.25.0/lib/docker/connection.rb:42:in `rescue in request': 400 Bad Request: malformed Host header (Docker::Error::ClientError)
    from /Users/nateh/.rvm/gems/ruby-2.2.3/gems/docker-api-1.25.0/lib/docker/connection.rb:38:in `request'
    from /Users/nateh/.rvm/gems/ruby-2.2.3/gems/docker-api-1.25.0/lib/docker/connection.rb:65:in `block (2 levels) in <class:Connection>'
    from /Users/nateh/.rvm/gems/ruby-2.2.3/gems/docker-api-1.25.0/lib/docker/container.rb:277:in `create'
    from /Users/nateh/code/open/ruby-kafka/spec/test_cluster.rb:209:in `create'
    from /Users/nateh/code/open/ruby-kafka/spec/test_cluster.rb:19:in `initialize'
    from /Users/nateh/code/open/ruby-kafka/spec/spec_helper.rb:63:in `new'
    from /Users/nateh/code/open/ruby-kafka/spec/spec_helper.rb:63:in `<top (required)>'
[...]

PR #306, opened 2016-12-16
Get functional tests working with Docker for Mac

jturkel:
This should fix https://github.com/zendesk/ruby-kafka/issues/238


dasch:
:heart:

PR #309, opened 2016-12-23
Only fetch messages for subscribed topics

tjwp:
We ran into an issue where a ConsumerGroup was assigned partitions for a topic that the Consumer was not subscribed to. This was a topic that the consumer group used to consume, but that we were no longer using.

This led to an exception in the OffsetManager because there was no default offset for the topic that the consumer no longer subscribes to: https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/offset_manager.rb#L104

The proposed fix here addresses the issue by only fetching from partitions for the topics that the consumer is subscribed to.

I tried to write a functional test for the ConsumerGroup to demonstrate the problem, but was unable to reproduce it using the test cluster.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -319,11 +319,11 @@ module Kafka
     def fetch_batches(min_bytes:, max_wait_time:)
       join_group unless @group.member?
 
-      assigned_partitions = @group.assigned_partitions
+      subscribed_partitions = @group.subscribed_partitions
 
       @heartbeat.send_if_necessary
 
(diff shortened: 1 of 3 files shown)


dasch:
This happened during a deploy? Sounds like the group leader was still using the old code base that included the topic and then assigned partitions in that topic to the rest of the group.

There's also another scenario: you add a new topic subscription, but the leader is still using the old code which doesn't. In this case, none of the members will be assigned partitions from that topic.

However, I think this should sort itself out: when the leader is updated, the group re-syncs and a new round of assignment takes place.

Thanks for the PR!

Issue #233, opened 2016-06-28
Allow configuring the offset commit retention time


dasch:
Currently, the broker default is always used. When a consumer group is stopped and started again after a while, the committed offsets may have expired, causing the group to re-process the topics.

The API could look like this:

``` ruby
consumer = kafka.consumer(group_id: "...", offset_commit_retention_time: 120)
```

I'm not sure whether to surface the nomenclature around "offset commits" or whether to use the more intuitive concept of "checkpointing".

dasch:
What is the maximum value allowed by Kafka?

tjwp:
The maximum appears to be `Integer.MAX_VALUE` (2_147_483_648), over 4000 years.

dasch:
OK, no problems then. I've run into weird arbitrary timeout limits in the past.

PR #316, opened 2017-01-27
Expose offset retention time in the Consumer API

tjwp:
This change implements a version of the enhancement described in https://github.com/zendesk/ruby-kafka/issues/233.

Our use case for this is a very low throughput topic where some partitions do not receive new messages within the broker's retention period setting (1 day). Previously committed offsets would be purged and the next day the partition would be read from the beginning again.

Exposing this setting to the consumer provides more flexibility than changing the broker's setting for all consumers.


dasch:
:heart:

Issue #317, opened 2017-01-27
Ensure that committed offsets are never purged for active consumer groups


tjwp:
We have some very low throughput topics where partitions may not receive a new message within the retention period for committed offsets.

Increasing the retention period via the broker setting or per consumer (as represented by https://github.com/zendesk/ruby-kafka/pull/316) helps, but there are limits to how high we want to raise this. We still want committed offsets for topics and consumer groups that are no longer used to be purged within some reasonable time period.

We want to be able to re-commit the previously committed offsets (not just the processed offsets in the offset manager) for consumers groups that are still in use to ensure that the committed offsets never get stale.

We have three basic approaches to suggest:

1. A new, public method in the Consumer API to explicitly re-commit the current offsets. This would combine the processed offsets and the committed offsets held by the offset manager. The intent here is not to open up an interface to manually commit arbitrary offsets. We have already implemented a form of this locally and call it prior to initiating a consumer loop. Our processes are restarted at least every 24 hours, so this ensures that offsets are re-committed at least once a day.
2. Periodically re-commit the previously committed offsets. This would require something like another offset commit interval, and when committing offsets the committed offsets would be combined with the processed offsets if the timeout for the re-commit was reached.
3. Always re-commit the previously committed offsets when committing the processed offsets. There could be some flag specified to the consumer to determine if this is behavior enabled. This approach avoids the need to track multiple intervals but may add overhead to already frequent re-commits of processed offsets (https://github.com/zendesk/ruby-kafka/pull/284).

Please let us know which the of approaches above, or some variant/combination of them, you’d be open to and we’d be happy to produce a PR.

dasch:
If there's an upper limit to the retention forced by Kafka I would be open to _always_ re-committing the offsets within that interval.

tjwp:
It looks like the upper limit on the `offsets.retention.minutes` setting is `Integer.MAX_VALUE` (2_147_483_647), which is over 4000 years.

We would want to re-commit offsets within the configured `offsets.retentions.minutes`. Perhaps offsets should be re-committed at the first commit, and within _half_ of the `offset_retention_time`? If unset, the broker's default of 1 day could be assumed, so offsets would be re-committed twice a day.

dasch:
Sounds reasonable. I'd :heart: a PR!

PR #318, opened 2017-02-01
Periodically re-commit previously committed offsets

tjwp:
This change addresses Issue https://github.com/zendesk/ruby-kafka/issues/317.

Previously committed offsets are re-committed at the first commit and periodically after that at an interval that is half the _offset retention time_. If the _offset retention time_ is not specified for the consumer, then half of the broker's default setting (1 day) is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,16 @@ Changes and additions to the library will be listed here.
 
 ## Unreleased
 
+- Re-commit previously committed offsets periodically with an interval of half
+  the offset retention time, starting with the first commit (#318).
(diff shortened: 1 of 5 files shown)


dasch:
:heart:

dasch:
Can you test this commit out in production before I cut a release with it?

tjwp:
> Can you test this commit out in production before I cut a release with it?

Yes, I'll do that today.

tjwp:
@dasch We've been running with this commit in staging and production today and it's behaving as expected.

dasch:
Awesome! I'll cut a new release within the next few weeks.

PR #320, opened 2017-02-13
Allow ssl_ca_cert to be an array of PEM encoded certificates

wjoel:
`ssl_ca_cert` adds a single certificate to the OpenSSL:X509::Store, but some configurations have longer certificate chains than just one level.

This commit exposes`add_file` of OpenSSL:X509::Store to allow adding all certificates contained in a file.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -454,12 +454,8 @@ module Kafka
 
       if ca_cert
         store = OpenSSL::X509::Store.new
-        if ca_cert.kind_of?(Array)
-          ca_cert.each do |cert|
-            store.add_cert(OpenSSL::X509::Certificate.new(cert))
-          end
-        else
-          store.add_cert(OpenSSL::X509::Certificate.new(ca_cert))
+        Array(ca_cert).each do |cert|
+          store.add_cert(OpenSSL::X509::Certificate.new(cert))
         end
         ssl_context.cert_store = store
       end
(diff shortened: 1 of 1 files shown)


tcrayford:
@wjoel -1 - can you have this not use a file, but instead use `add_cert`? The current mechanism is explicitly not coupled to files, and should remain so.

Maybe we could make `ssl_ca_cert` optionally an array of certs and add all of them?

wjoel:
@tcrayford, I've updated the PR. Is this what you had in mind?

wjoel:
I doubt the failing test is related to my changes, but I could be wrong.

wjoel:
@tcrayford, does this look ok now? Not sure how to re-trigger testing, but the failure seems unrelated to my changes.

lib/kafka/client.rb
```diff
@@ -454,7 +454,13 @@ def build_ssl_context(ca_cert, client_cert, client_cert_key)
 
       if ca_cert
         store = OpenSSL::X509::Store.new
-        store.add_cert(OpenSSL::X509::Certificate.new(ca_cert))
+        if ca_cert.kind_of?(Array)
+          ca_cert.each do |cert|
```
tcrayford: mind refactoring this to:

```ruby
Array(ca_cert).each do |cert|
  store.add_cert(OpenSSL::X509::Certificate.new(cert)
end
```

Thanks!

wjoel: Nice. Done.

tcrayford:
Looks good to me. I don't have commit rights - @dasch can you merge this when you get the time? Thanks.

dasch:
:heart:

Issue #314, opened 2017-01-23
Passphrase for private key files 


iam404:
Hi

I was looking for passing passphrase of a private key file into below code snippet : 

```
kafka = Kafka.new(
  ssl_ca_cert: File.read('my_ca_cert.pem'),
  ssl_client_cert: File.read('my_client_cert.pem'),
  ssl_client_cert_key: File.read('my_client_cert_key.pem'),
)
```
Any help ?

dasch:
@iam404 sorry for the late response. I'm unfortunately not familiar with how you'd do that – @tcrayford would you know this?

tcrayford:
@dasch it looks like http://stackoverflow.com/questions/235759/opening-an-rsa-private-key-from-ruby is the right thing - we'd need another parameter for the passphrase, that's all. @iam404 want to make a PR for that?

iam404:
Ya sure , will test the above solution and create a PR. Thanks.

PR #323, opened 2017-03-01
Adding support for passphrase for private key files 

iam404:
Fix Issue #314

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -41,12 +41,12 @@ module Kafka
     #   SSL connection. Must be used in combination with ssl_client_cert.
     #
     # @return [Client]
-    def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_timeout: nil, socket_timeout: nil, ssl_ca_cert: nil, ssl_client_cert: nil, ssl_client_cert_key: nil)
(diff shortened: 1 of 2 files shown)


lib/kafka/client.rb
```diff
@@ -436,15 +436,16 @@ def initialize_cluster
...
+    def build_ssl_context(ca_cert, client_cert, client_cert_key, client_cert_key_passphrase)
       return nil unless ca_cert || client_cert || client_cert_key
 
       ssl_context = OpenSSL::SSL::SSLContext.new
 
       if client_cert && client_cert_key
         ssl_context.set_params(
           cert: OpenSSL::X509::Certificate.new(client_cert),
-          key: OpenSSL::PKey.read(client_cert_key)
+          key: OpenSSL::PKey.read(client_cert_key, client_cert_key_passphrase)
```
tcrayford: what happens here if you run it past a client cert key that doesn't have a passphrase?

tcrayford:
I'd like to see tests for this

lib/kafka/client.rb
iam404: If `ssl_client_cert_key_passphrase` is not pass then default  value `nil` is saved. Also openssl tend to use passphrase only when key is encrypted. If it is not encrypted then passphrase provided is not used.
https://github.com/ruby/openssl/blob/7b319d95ed46a9ffb1d9962d5d4de8bff94cb28c/ext/openssl/ossl_pkey_rsa.c#L210

dasch:
@iam404 do you expect to work more on this, or should I close it?

dasch:
Closed due to lack of activity. @iam404 I can reopen if you'd like to continue to work on this.

PR #324, opened 2017-03-09
Fix TypeError nil can't be coerced into Fixnum

eldan-klarna:
When @buffer[topic][partition].count was zero a "TypeError nil can't be coerced into Fixnum" was thrown
because [].reduce(:+) returns nil (line 59)
and Fixnum - nil raises this error


dasch:
❤️

Issue #119, opened 2016-03-04
Authentication using SASL


dasch:
http://kafka.apache.org/documentation.html#security_sasl

kzk:
@dasch Hi, we're running the project called Fluentd (http://www.fluentd.org/), an open source data collector. Currently Fluentd's kafka plugin is contributed by community (https://github.com/htgc/fluent-plugin-kafka), but it uses Poseidon. Apparently it doesn't work with v0.9, and we're exploring the possibility of rewriting it with ruby-kafka. One requirements we hear from Fluentd users are, they want Kerberos support to publish events to Kafka. I just saw this ticket but are you now working on this right now? If not, are you open to merge the changes we'll make? I'm just curious how important this is for you at this point. cc/ @repeatedly

dasch:
@kzk that sounds great! I'm not working on Kerberos support, so by all means go ahead :-)

If you need help we can set up a call – otherwise, a work-in-progress PR could be a good starting point.

tmeinlschmidt:
@dasch I'm on adding SASL (kerberos) to ruby-kafka right now, as we need it. It seems to be working, so I'll create some pull request soon. Have to add support and tests for SASL_SSL then too.

jpaskhay:
@tmeinlschmidt We are also in need of this feature for a project we're working on, so we'd be more than happy to help in whatever capacity possible. We were literally about to start doing this ourselves, so we are very thankful to have seen your comment :) We don't have a strong Ruby background on the team but willing to do what we can to help.

We are not currently using SSL, so we may not be able to assist much there, unfortunately.

tmeinlschmidt:
Hey all, pushed some initial code for SASL/GSSAPI authentication. Working on adding some more tests and docs.

dasch:
@tmeinlschmidt I think that code is being overly intrusive - can you make a change within the confines of the existing design?

Feel free to create a PR with design changes separately, although I won't promise that I'll merge it.

tmeinlschmidt:
The problem is - SASL needs to send some kafka messages before any auth can be made. that's the reason I cut ConnectionOperation out from Connection class. Is there any specific part you want to change/rewrite?

dasch:
@tmeinlschmidt can you start by opening a PR? That makes commenting a lot easier.

tmeinlschmidt:
@dasch opened #334.

huhongda:
[...]

PR #334, opened 2017-03-29
Initial commit of SASL/GSSAPI support (#119)

tmeinlschmidt:
- some refactoring of connection, as it needs to send messages prior to GSSAPI authentication
- gssapi gem required
- using constants instead of plain integers in api keys


lib/kafka/client.rb
```diff
@@ -40,8 +40,13 @@ class Client
...
     #   SSL connection. Must be used in combination with ssl_client_cert.
     #
+    # @param principal [String, nil] a KRB5 principal
+    #
+    # @param keytab [String, nil] a KRB5 keytab filepath
+    #
     # @return [Client]
-    def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_timeout: nil, socket_timeout: nil, ssl_ca_cert: nil, ssl_client_cert: nil, ssl_client_cert_key: nil)
+    def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_timeout: nil, socket_timeout: nil,
+                   ssl_ca_cert: nil, ssl_client_cert: nil, ssl_client_cert_key: nil, principal: nil, keytab: nil)
```
dasch: First of all, we'd need to prefix the new keywords with something that makes sense, e.g. `krb5_`.

lib/kafka/connection.rb
```diff
@@ -88,10 +93,10 @@ def send_request(request)
 
         @correlation_id += 1
 
-        write_request(request, notification)
+        @operation.write_request(request, notification, @correlation_id)
 
         response_class = request.response_class
-        wait_for_response(response_class, notification) unless response_class.nil?
+        @operation.wait_for_response(response_class, notification, @correlation_id) unless response_class.nil?
```
dasch: I don't see why we need to wrap Connection here – why not just instantiate the handshake connection and execute it here? Why does `write_request` and `wait_for_response` have to move?

tmeinlschmidt: I can't make new instance of a Connection inside SaslConnectionWithTimeout. It call's `open` etc. So that was the reason I moved out just simple read/write and wrote own `send_request` in SaslConnection class. The other (dirty) way is to compose kafka message (SaslHandshakeRequest) out of the box and do not use this at all. But when protocol changes, it must be changes here as well, so it's really ugly and I didn't do that.

The reason why it in separate class - Kafka supports sasl plain auth too (and other sasl methods), which are not here as well and then encryption with SSL (after SASL auth). So I think it's better to have separate class, where I can do the SASL protocol negotiation (plain vs gssapi or others), do the auth and simply return the socket.

lib/kafka/client.rb
tmeinlschmidt: okay. no prob. just think about more sasl methods, eg. plain, then there's a request to support SCRAM (scram-sha256), so perhaps use something like `sasl_krb5_principal` or `sasl_plain_username`...

dasch: Makes sense.

lib/kafka/connection.rb
dasch: Would it be possible to do the handshaking using only calls to `write_request` and `wait_for_response` or do you _need_ lower level writes and reads? I think it would be better to handle this at the client level if possible.

tmeinlschmidt: @dasch unf. I need to send one kafka message (apikey 17), then read the response, and then send and receive low-level data...

lib/kafka/client.rb
tmeinlschmidt: great. Maybe - better to use `gssapi` instead of `krb5`, as it's commonly used in kafka docs and configs. so I'd change `principal` to `sasl_gssapi_principal` and `keytab` to `sasl_gssapi_keytab`

lib/kafka/connection.rb
dasch: I would rather we add low-level `write` and `read` methods to the existing Connection and handle the SASL stuff in Client – ideally using a separate SaslOperation that does the full thing.

lib/kafka/protocol/api_key.rb
```diff
@@ -0,0 +1,27 @@
...
+      JOIN_GROUP      = 11
+      HEARTBEAT       = 12
+      LEAVE_GROUP     = 13
+      SYNC_GROUP      = 14
+      DESCRIBE_GROUPS = 15
+      LIST_GROUPS     = 16
+      SASL_HANDSHAKE  = 17
+      API_VERSIONS    = 18
+      CREATE_TOPICS   = 19
+      DELETE_TOPICS   = 20
```
dasch: I don't think this is actually _required_ for this PR. I'd prefer to do these kinds of changes outside of feature PRs.

tmeinlschmidt:
@dasch I did some changes, can you take a look again, pls? Thinking how to add more tests.

lib/kafka/connection_builder.rb
```diff
@@ -1,25 +1,44 @@
...
         host: host,
         port: port,
         client_id: @client_id,
         connect_timeout: @connect_timeout,
         socket_timeout: @socket_timeout,
         logger: @logger,
         instrumenter: @instrumenter,
-        ssl_context: @ssl_context,
+        ssl_context: @ssl_context
       )
```
dasch: The connection builder needs to be able to build _multiple_ connections, so we can't store a single connection in instance state.

lib/kafka/connection_builder.rb
```diff
@@ -1,25 +1,44 @@
...
+      return @connection.socket unless sasl_gssapi_auth?
```
dasch: The method still needs to return the full Connection.

lib/kafka/sasl_gssapi_authenticator.rb
```diff
@@ -0,0 +1,75 @@
...
+      msg = @decoder.read(@decoder.int32)
+      raise Kafka::Error, "GSSAPI negotiation failed." unless msg
+      # unwrap with integrity only
+      msg_unwrapped = @gssapi_ctx.unwrap_message(msg, GSSAPI_CONFIDENTIALITY)
+      msg_wrapped = @gssapi_ctx.wrap_message(msg_unwrapped + @principal, GSSAPI_CONFIDENTIALITY)
+      @encoder.write(msg_with_size(msg_wrapped))
+    end
+
+    def msg_with_size(msg)
+      [msg.size].pack('l>') + msg
```
dasch: Isn't this the same as [`Encoder#write_bytes`](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/protocol/encoder.rb#L90-L97)?

lib/kafka/sasl_gssapi_authenticator.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      # read wrapped message and return it back with principal
+      handshake_messages
+
+      # no error, return @socket
+      @socket
+    end
+
+    def handshake_messages
+      msg = @decoder.read(@decoder.int32)
```
dasch: You should be able to just do `msg = @decoder.bytes`.

lib/kafka/sasl_gssapi_authenticator.rb
```diff
@@ -0,0 +1,75 @@
...
+      raise Kafka::Error, "GSSAPI negotiation failed." unless msg
+      # unwrap with integrity only
+      msg_unwrapped = @gssapi_ctx.unwrap_message(msg, GSSAPI_CONFIDENTIALITY)
+      msg_wrapped = @gssapi_ctx.wrap_message(msg_unwrapped + @principal, GSSAPI_CONFIDENTIALITY)
+      @encoder.write(msg_with_size(msg_wrapped))
```
dasch: This can just be `@encoder.write_bytes(msg_wrapped)`, I think.

ruby-kafka.gemspec
```diff
@@ -25,6 +25,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_runtime_dependency 'gssapi', '>=1.2.0'
```
dasch: Does this require any special C libraries or is it fully portable?

lib/kafka/connection_builder.rb
tmeinlschmidt: changed

ruby-kafka.gemspec
tmeinlschmidt: gssapi requires `ffi`, to map C functions to ruby... I think there's no simple way to rewrite it entirely in ruby, as it's still system dependent (based on system kerberos implementation)

tmeinlschmidt:
Tests are failing here, weird, on my local it works correctly - with the same docker.

````
$ bundle exec rspec --tag functional
...
Killing broker 0
Starting broker 0
  handle a broker going down after the initial discovery
Stopping cluster...

Finished in 8 minutes 30 seconds (files took 0.37519 seconds to load)
25 examples, 0 failures
````

File: ruby-kafka.gemspec (hunk 9 above)
Comment (dasch) [inline ruby-kafka.gemspec]: Does that mean that it'll only work on systems with the Kerberos dev packages installed?

**Comment by dasch** [comment]
Sorry for not getting back to this, I've been on vacation these last few weeks.

File: lib/kafka/connection.rb [hunk 10]
```diff
@@ -28,6 +28,10 @@ class Connection
     SOCKET_TIMEOUT = 10
     CONNECT_TIMEOUT = 10
 
+    attr_reader :socket
+    attr_reader :encoder
+    attr_reader :decoder
+
```
dasch: ~~I don't think these attributes are used anywhere...~~ Looks like only `socket` is unused.

ruby-kafka.gemspec
tmeinlschmidt: nope. It doesn't require dev versions.. check this
https://github.com/zenchild/gssapi/blob/master/lib/gssapi/lib_gssapi_loader.rb
so whenever you are using kerberos, you already have these libs installed. And these libs are required/loaded only when gssapi is required.

dasch: Yeah – but won't they be required by all `ruby-kafka` installations, since you're unconditionally requiring the library?

dasch: Basically: can you install ruby-kafka without installing some system package first?

tmeinlschmidt: yes. No need to install anything else if you don't plan to use gssapi sasl authentication.

dasch:
Looks good! I'll merge this tomorrow.

Thanks for contributing and sorry for the long wait ❤️

tmeinlschmidt:
let's wait until Monday, I want to try against some other version.. just to be sure. thanks

tmeinlschmidt:
fixed one small issue and tested. You could merge it. Thanks!

jeremywu0127:
@tmeinlschmidt this PR only support `SASL/GSSAPI`?  Does it support `SASL/PLAIN`?

tmeinlschmidt:
@jeremywu0127 only SASL/GSSAPI. Would like to add support for /PLAIN then too.

dasch:
@tmeinlschmidt would you like to do that in this PR?

tmeinlschmidt:
@dasch nope. Gonna create another PR then. It's a bit different than GSSAPI.

dasch:
@tmeinlschmidt okay, I'll merge this when I have time then – I'm a bit swamped this week.

dasch:
@tmeinlschmidt ❤️

dasch:
@tmeinlschmidt would you like to get PLAIN support included in the next release? If so, I'll probably need it merged into master within one or two weeks.

tmeinlschmidt:
@dasch I'll try to finish it before the release.

PR #336, opened 2017-04-04
Update topic metadata when leader not available

### LINKED ISSUE #335: Stuck on "Leader not available" for auto-created topics

* Version of Ruby: 2.3.1
* Version of Kafka: 0.10.2.0
* Version of ruby-kafka: master (02169e9f1656c988a030e82466af4f9ad5e80e45)

###### Steps to reproduce

Start a Kafka server with `num.partitions` > 8 (or adjust the number of threads created with this sample code).

```ruby
require 'kafka'
require 'logger'

logger = Logger.new(STDOUT)
logger.level = Logger::DEBUG

brokers = ['127.0.0.1:9092']

# topic-xxxxxxxx
topic = "test-#{(0...8).map { (97 + rand(26)).chr }.join}"

threads = []

8.times do
  threads << Thread.new {
    client = Kafka.new(seed_brokers: brokers, logger: logger)
    consumer = client.consumer(group_id: 'group')

    consumer.subscribe(topic)

      puts "#{message.topic}: #{message.value}"
    end
  }
end

threads.each do |t|
  t.join
end
```

###### Expected outcome

Each of the threads should be assigned partitions and start waiting to consume messages.

###### Actual outcome

Some threads get stuck, printing `Leader not available; waiting 1s before retrying`

---

I'm fairly sure this is happening because the topic doesn't initially exist (but is auto-created). It looks like after the first metadata query, where Kafka returns an error code of 5 for the topic, no further metadata queries are made.

It seems that the following change can fix this, but I'm hesitant to open a PR because it might introduce unwanted behaviour.

```diff
diff --git a/lib/kafka/consumer.rb b/lib/kafka/consumer.rb
index e699341..8b627ec 100644
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -286,6 +286,7 @@ module Kafka
           @cluster.mark_as_stale!
         rescue LeaderNotAvailable => e
           @logger.error "Leader not available; waiting 1s before retrying"
+          @cluster.mark_as_stale!
           sleep 1
         end
       end
```

anthonynguyen:
Fixes #335

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -286,6 +286,7 @@ module Kafka
           @cluster.mark_as_stale!
         rescue LeaderNotAvailable => e
           @logger.error "Leader not available; waiting 1s before retrying"
+          @cluster.mark_as_stale!
           sleep 1
         end
       end


dasch:
🎉

NeQuissimus:
@dasch Hate to be a pest but is there any chance @anthonynguyen and I could get a release with this change soon(ish)?

dasch:
@NeQuissimus I just released v0.3.17 :-)

NeQuissimus:
Sweet, thanks!

PR #340, opened 2017-04-12
Retry final offsets commit if timed out

Nattfodd:
It turns out it's a pretty common situation when we get `ConnectionError` right after we stopped consumer.
Retrying offsets commitment allows us to make sure it is commited properly to avoid situation when large batch of messages will be processed again after consumer's restart.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -293,9 +293,22 @@ module Kafka
     ensure
       # In order to quickly have the consumer group re-balance itself, it's
       # important that members explicitly tell Kafka when they're leaving.
-      @offset_manager.commit_offsets rescue nil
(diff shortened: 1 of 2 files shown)


lib/kafka/consumer.rb
```diff
@@ -293,11 +293,24 @@ def consumer_loop
     ensure
       # In order to quickly have the consumer group re-balance itself, it's
       # important that members explicitly tell Kafka when they're leaving.
-      @offset_manager.commit_offsets rescue nil
+      make_final_offsets_commit!
       @group.leave rescue nil
       @running = false
     end
 
+    def make_final_offsets_commit!(attemtps = 5)
```
dasch: There's a typo in the argument name – it should be `attempts`.

PR #341, opened 2017-04-19
shuffle seed brokers before fetching metadata

tcrayford:
Yesterday, a component of ours had vastly increased latency during an
outage, to the point where callers were timing out and seeing issues.

Upon tracking this down, we found that the cause of the latency was that
broker 0 was replaced. This component uses the sync producer, and as a
result of broker 0 being replaced, any produce request that fetched
metadata stalled for the default connect timeout (10s).

Instead, fetch metadata from seed brokers in a random order. This means
that a single broker going down will slow down at most 1/NBROKERS
metadata fetches, rather than every single one.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -190,7 +190,7 @@ module Kafka
     # @raise [ConnectionError] if none of the nodes in `seed_brokers` are available.
     # @return [Protocol::MetadataResponse] the cluster metadata.
     def fetch_cluster_info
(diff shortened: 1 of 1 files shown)


tcrayford:
@dasch I think the circleci failure is unrelated, is it?

dasch:
It seems so, but restarting the build doesn't seem to fix it...

dasch:
Merging this, will fix CI tomorrow.

PR #342, opened 2017-04-20
Allow committing offsets manually 

ale7714:
This PR exposes `#commit_offsets` on Consumer and adds an option to avoid auto commit offsets on `#each_batch`

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -230,9 +230,11 @@ module Kafka
     #   is ignored.
     # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
     #   returning messages from the server, in seconds.
+    # @param auto_commit_offsets [Boolean] if true, it will commit offsets when necessary.
+    #   Otherwise it won't commit offsets.
     # @yieldparam batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
     # @return [nil]
-    def each_batch(min_bytes: 1, max_wait_time: 5)
+    def each_batch(min_bytes: 1, max_wait_time: 5, auto_commit_offsets: true)
       consumer_loop do
         batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
 
@@ -263,7 +265,7 @@ module Kafka
             mark_message_as_processed(batch.messages.last)
           end
 
-          @offset_manager.commit_offsets_if_necessary
+          @offset_manager.commit_offsets_if_necessary if auto_commit_offsets
 
           @heartbeat.send_if_necessary
 
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -272,6 +272,10 @@ module Kafka
       end
     end
 
+    def commit_offsets
+      @offset_manager.commit_offsets
+    end
+
     private
 
     def consumer_loop
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -200,7 +200,7 @@ module Kafka
             end
 
             mark_message_as_processed(message)
-            @offset_manager.commit_offsets_if_necessary
+            @offset_manager.commit_offsets_if_necessary if @offset_manager.offset_commit_enabled
 
             @heartbeat.send_if_necessary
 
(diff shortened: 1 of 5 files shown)


lib/kafka/consumer.rb
```diff
@@ -230,9 +230,11 @@ def each_message(min_bytes: 1, max_wait_time: 5)
     #   is ignored.
     # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
     #   returning messages from the server, in seconds.
+    # @param auto_commit_offsets [Boolean] if true, it will commit offsets when necessary.
+    #   Otherwise it won't commit offsets.
     # @yieldparam batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
     # @return [nil]
-    def each_batch(min_bytes: 1, max_wait_time: 5)
+    def each_batch(min_bytes: 1, max_wait_time: 5, auto_commit_offsets: true)
```
ale7714: I'm thinking on adding this option to `each_message` too.

dasch: I think the option should go in the initializer instead and be exposed by https://github.com/ale7714/ruby-kafka/blob/ab128e762d30601867c67bc0934c1bbf613bc934/lib/kafka/client.rb#L222 – that's where the current offset commit policy options are.

dasch: And maybe call it `offset_commit_enabled` in order to keep things consistent.

lib/kafka/consumer.rb
```diff
@@ -263,7 +263,7 @@ def each_batch(min_bytes: 1, max_wait_time: 5)
             mark_message_as_processed(batch.messages.last)
           end
 
-          @offset_manager.commit_offsets_if_necessary
+          @offset_manager.commit_offsets_if_necessary if @offset_manager.offset_commit_enabled
```
ale7714: Another option is to update `#commit_offsets_if_necessary` to do nothing when `offset_commit_enabled` is false. And i think it wouldn't be a bad option because the `#commit_offsets` is exposed on consumer. What do you think?

dasch: I agree that's a better option.

ale7714:
@dadah89 thank you for your input, I just updated the PR. It looks better now :)

dadah89:
I think you meant @dasch =)

ale7714:
oh yep sorry!

ale7714:
Another side effect of avoiding auto-offset-commits, is that we don't want to do this https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L296. Consumer may stop before actually doing the expected processing but it will commit the offset, causing loss messages. 

I'm thinking changing this line for:
```
make_final_offsets_commit! if @offset_manager. commit_enabled
```

WDYT?

dasch:
@ale7714 I agree.

ale7714:
All set :), let me know if I need to add anything else.

dasch:
Great! I'll take a look later today.

dasch:
:heart: 

I have a few nitpicks that I'll fix myself.

dasch:
Actually, I'm having second thoughts on this design. I think a better and safer way to handle the problem would be to allow disabling the automatic marking of messages as processed – it's really not the commits themselves that are the problem, it's the advancement of the offset pointer.

How about this API?

```ruby
# We would still like the consumer to control the commit rate – both values can be set to
# 0 to disable those automated triggers.
consumer = kafka.consumer(..., commit_interval: 10, commit_threshold: 5)

  buffer = Buffer.new # imaginary buffer thingie

    buffer.add(message)

    if buffer.ready?
      buffer.do_that_thing # e.g. complete transaction
      buffer.clear
      consumer.commit_offsets # We can optionally trigger a synchronous commit.
    end
  end
end
```

dasch:
I've reverted this PR – I think controlling message marking is the right approach. @ale7714 would you be up for a second PR?

ale7714:
@dasch i like this approach, it's clearer what's happening inside of the loop. But one little detail if that the buffer may not be ready on one `each_batch` loop. Which mean the offset pointer won't move and it will retrieve the send batch? in that case should we have something like offset buffer?

dasch:
There are two offsets in play: the offset of the last message fetched from the broker, and the offset of the last message successfully processed. We'll still increment the former, so we'll get new batches as long as the consumer doesn't lose its in-memory state.

ale7714:
Oh, that's right! Will open another PR. Thanks for the support.

dasch:
Thanks for all the work! :D

PR #344, opened 2017-04-21
Add auto commit option to each batch 

ale7714:
This is a follow up to https://github.com/zendesk/ruby-kafka/pull/342

* Exposes `#commit_offsets` and `#automatically_mark_as_processed ` on Consumer
* It adds `automatically_mark_as_processed` option to `#each_batch` and `#each_message`.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -272,6 +272,10 @@ module Kafka
       end
     end
 
+    def commit_offsets
+      @offset_manager.commit_offsets
+    end
+
     private
 
     def consumer_loop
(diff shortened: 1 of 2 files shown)


ale7714:
PR on `ruby-kafka` gem https://github.com/zendesk/ruby-kafka/pull/344

lib/kafka/consumer.rb
```diff
@@ -199,7 +201,7 @@ def each_message(min_bytes: 1, max_wait_time: 5)
               end
             end
 
-            mark_message_as_processed(message)
+            mark_message_as_processed(message) if automatically_mark_as_processed
```
dasch: I think it would make sense to only add this option to `#each_batch` – I'd like to see an actual use case where it would make sense to use it with `#each_message` before adding more API surface.

ale7714:
@dasch yeah it looks cleaner 😄. I updated the PR too

PR #350, opened 2017-05-04
Retrieve end offests for topics (issue 311)

ttamimi:


lib/kafka/client.rb
```diff
@@ -414,6 +414,32 @@ def last_offset_for(topic, partition)
       @cluster.resolve_offset(topic, partition, :latest) - 1
     end
 
+
+    # Retrieve the offset of the last message in each partition of the specfied topics.
```
dasch: Typo in "specfied".

lib/kafka/client.rb
```diff
@@ -414,6 +414,32 @@ def last_offset_for(topic, partition)
...
+    #
+    # @example
+    #   last_offsets_for ['topic-1', 'topic-2']  ->
+    #   {
+    #     'topic-1' => { 0 => 100, 1 => 100 },
+    #     'topic-2' => { 0 => 100, 1 => 100 },
+    #   }
+    def last_offsets_for(topics = nil)
+      topics   = [topics] if !topics.nil? && topics.is_a?(String)
+      topics ||= self.topics
```
dasch: I think this can be simplified a bit:

1) is there a reason to support fetching _all_ offsets? if not, let's not allow that for now.
2) you can use the splat operator to accept multiple arguments: `def last_offsets_for(*topics)`

ttamimi: Thanks for the quick feedback!
Re: #1, sure, I will change it.
Re: #2, great suggestion!

PR #352, opened 2017-05-12
Add NoPartitionsAssignedError

ale7714:
* Raise a more meaningfull error when no partitions are
  assigned

This small change is to make easier/cleaner handling this error on the client side.


dasch:
❤️

PR #353, opened 2017-05-13
Manual offset management 

ale7714:
This PR intends to extend the support on the manual offset management previously introduced by https://github.com/zendesk/ruby-kafka/pull/344.  

Proposed workflow: 

```ruby
buffer = []

consumer.each_message(automatically_mark_as_processed: false) do |message|
  # Our messages are JSON with a `type` field and other stuff.
  event = JSON.parse(message.value)

  case event.fetch("type")
  when "add_to_cart"
    buffer << event
  when "complete_purchase"
    # We've received all the messages we need, time to save the transaction.
    save_transaction(buffer)

    # We must be sure we're committing/processing the offset only if we're subscribed
    # to that partition. This prevents a message on the buffer from a previously  
    # subscribed partition from committing a wrong/out of date offset
    if consumer.subscribed_partitions[message.topic].include?(message.partition)
      consumer.commit_offsets
   end

    buffer.clear
  end
  
  # We want to keep moving forward the offset pointer so we can keep 
  # fetching without committing the offset. 
  consumer.mark_message_as_buffered(message)
end
```

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -177,7 +177,7 @@ module Kafka
     #   {Kafka::ProcessingError} instance.
     # @return [nil]
     def each_message(min_bytes: 1, max_wait_time: 5, automatically_mark_as_processed: true)
-      consumer_loop do
+      consumer_loop(clear_offsets_on_join: automatically_mark_as_processed) do
         batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
 
         batches.each do |batch|
@@ -241,7 +241,7 @@ module Kafka
     # @yieldparam batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
     # @return [nil]
     def each_batch(min_bytes: 1, max_wait_time: 5, automatically_mark_as_processed: true)
-      consumer_loop do
+      consumer_loop(clear_offsets_on_join: automatically_mark_as_processed) do
         batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
 
         batches.each do |batch|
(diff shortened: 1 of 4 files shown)


lib/kafka/consumer.rb
```diff
@@ -327,12 +335,12 @@ def make_final_offsets_commit!(attempts = 3)
...
     end
 
-    def join_group
+    def join_group(clear_offsets_on_join: false)
       old_generation_id = @group.generation_id
 
       @group.join
 
-      if old_generation_id && @group.generation_id != old_generation_id + 1
+      if (old_generation_id && @group.generation_id != old_generation_id + 1) || clear_offsets_on_join
```
ale7714: i can moved this proposed change into a separate PR, that's why I kept it as the last commit. We would like to have the option of always clearing the offset but this change is not strictly required on this PR.

dasch:
I think this is probably adding too much complexity considering the relative infrequency with which users will want to manually manage offsets.

Try reading the section called Manual Offset Control in [the JVM client's documentation](https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html) – would it make more sense to provide a polling-based API that doesn't do automatic message marking and let users with those needs use that instead of `each_message`?

Alternatively, we could provide a callback system that allows users to e.g. clear buffers when partitions are reassigned – but that's also rather complex.

ale7714:
@dasch the `poll` option could be a good approach. But i guess to be able to cross the batch boundaries, you will need to be able to start polling from a particular offset. Would that make sense?

dasch:
I guess it still wouldn't solve the problem of clearing buffers when partition assignment changes. I think a simple callback would perhaps do the trick – would that work for you? Something like

```ruby
consumer.on_partition_assignment_change do
  buffer.clear
end
```

ale7714:
I'll close this PR on favor of https://github.com/zendesk/ruby-kafka/pull/369. And also will work on a separate PR to add `on_partition_assignment_change` concept.

Issue #357, opened 2017-05-25
Connection to broker shouldn't be terminated when kafka returns `MessageSizeTooLarge` error


jasantunes:
I recently came across the `MessageSizeTooLarge` error response from kafka, which I think it's not being properly handled by the `ruby-kafka`, at least in the async call from the producer. Currently it seems to be disconnecting the producer, thus triggering a bunch on other issues and ultimately 500s. The synchronized call seems to be handling the error fine.

* Version of Ruby: 2.2.3, 2.3.3
* Version of Kafka: 0.10.1.0
* Version of ruby-kafka: 0.3, master

###### Steps to reproduce

When calling it synchronously, the exception is raised properly without affecting the kafka connection:

```ruby
2.2.3 (main):0 > k = Kafka.new(logger: Logger.new(STDOUT), seed_brokers: [ 'localhost'], client_id: "my-application")
2.2.3 (main):0 > k.deliver_message("A"*1500000, topic: "development_events")
I, [2017-05-24T17:36:14.573805 #70030]  INFO -- : New topics added to target list: development_events
I, [2017-05-24T17:36:14.573922 #70030]  INFO -- : Fetching cluster metadata from kafka://localhost:9092
D, [2017-05-24T17:36:14.574013 #70030] DEBUG -- : Opening connection to localhost:9092 with client id my-application...
D, [2017-05-24T17:36:14.576325 #70030] DEBUG -- : Sending request 1 to localhost:9092
D, [2017-05-24T17:36:14.576523 #70030] DEBUG -- : Waiting for response 1 from localhost:9092
D, [2017-05-24T17:36:14.580413 #70030] DEBUG -- : Received response 1 from localhost:9092
I, [2017-05-24T17:36:14.580479 #70030]  INFO -- : Discovered cluster metadata; nodes: localhost:9092 (node_id=0)
D, [2017-05-24T17:36:14.580559 #70030] DEBUG -- : Closing socket to localhost:9092
D, [2017-05-24T17:36:14.580825 #70030] DEBUG -- : Current leader for development_events/0 is node localhost:9092 (node_id=0)
I, [2017-05-24T17:36:14.580913 #70030]  INFO -- : Sending 1 messages to localhost:9092 (node_id=0)
D, [2017-05-24T17:36:14.580956 #70030] DEBUG -- : Opening connection to localhost:9092 with client id my-application...
[...]

PR #358, opened 2017-05-25
Handle MessageSizeTooLarge errors #357

jasantunes:
When calling the `produce` (async version) on kafka and it  responds with `MessageSizeTooLarge`, the exception is being raised and the connection is abruptly terminated (see [issue 357](https://github.com/zendesk/ruby-kafka/issues/357)). This is not the behavior for other types of errors (e.g.,`CorruptMessage`) or even for the non-async version (`Kafka#deliver_messages`).

This PR adds support to handle `MessageSizeTooLarge` errors (w/o terminating the connection). Also, it clears this message from the buffer to prevent it from being retried (it's not a transient error).

### DIFF
--- a/lib/kafka/produce_operation.rb
+++ b/lib/kafka/produce_operation.rb
@@ -27,7 +27,7 @@ module Kafka
   # * `:sent_message_count` – the number of messages that were successfully sent.
   #
   class ProduceOperation
-    def initialize(cluster:, buffer:, compressor:, required_acks:, ack_timeout:, logger:, instrumenter:)
+    def initialize(cluster:, buffer:, compressor:, required_acks:, ack_timeout:, logger:, instrumenter:, error_handler: nil)
       @cluster = cluster
       @buffer = buffer
       @required_acks = required_acks
@@ -35,6 +35,7 @@ module Kafka
       @compressor = compressor
       @logger = logger
       @instrumenter = instrumenter
+      @error_handler = error_handler
     end
 
     def execute
@@ -123,7 +124,7 @@ module Kafka
               topic: topic,
               exception: [e.class.to_s, e.message],
             })
-
+            @error_handler.on_error(e, topic, partition, messages) if @error_handler
             raise e
           end
 
@@ -139,7 +140,6 @@ module Kafka
           end
         rescue Kafka::MessageSizeTooLarge
           @logger.error "Message is too large when writing to #{topic}/#{partition}"
-          @buffer.clear_messages(topic: topic, partition: partition)
         rescue Kafka::CorruptMessage
           @logger.error "Corrupt message when writing to #{topic}/#{partition}"
         rescue Kafka::UnknownTopicOrPartition
--- a/lib/kafka/produce_operation.rb
+++ b/lib/kafka/produce_operation.rb
@@ -137,6 +137,9 @@ module Kafka
               delay: ack_time - message.create_time,
             })
           end
+        rescue Kafka::MessageSizeTooLarge
+          @logger.error "Message is too large when writing to #{topic}/#{partition}"
+          @buffer.clear_messages(topic: topic, partition: partition)
         rescue Kafka::CorruptMessage
           @logger.error "Corrupt message when writing to #{topic}/#{partition}"
         rescue Kafka::UnknownTopicOrPartition
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -89,6 +89,11 @@ module Kafka
       @timer = Timer.new(queue: @queue, interval: delivery_interval)
     end
 
+    def with_error_handler(error_handler)
+      @worker.producer.with_error_handler(error_handler)
+      self
+    end
+
     # Produces a message to the specified topic.
     #
     # @see Kafka::Producer#produce
@@ -180,6 +185,7 @@ module Kafka
     end
 
     class Worker
+      attr_reader :producer
       def initialize(queue:, producer:, delivery_threshold:)
         @queue = queue
         @producer = producer
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -60,6 +60,12 @@ module Kafka
       @cluster = initialize_cluster
     end
 
+    # Override error handler.
+    def with_error_handler(error_handler)
+      @error_handler = error_handler
+      self
+    end
+
     # Delivers a single message to the Kafka cluster.
     #
     # **Note:** Only use this API for low-throughput scenarios. If you want to deliver
@@ -117,6 +123,7 @@ module Kafka
         compressor: compressor,
         logger: @logger,
         instrumenter: @instrumenter,
+        error_handler: @error_handler
       )
 
       operation.execute
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -150,6 +150,15 @@ module Kafka
 
       # Messages added by `#produce` but not yet assigned a partition.
       @pending_message_queue = PendingMessageQueue.new
+
+      # Object/class that implements handle_error(status_code)
+      @error_handler = nil
+    end
+
+    # Override error error_handler.
+    def with_error_handler(error_handler)
+      @error_handler = error_handler
+      self
     end
 
     # Produces a message to the specified topic. Note that messages are buffered in
@@ -288,6 +297,7 @@ module Kafka
         compressor: @compressor,
         logger: @logger,
         instrumenter: @instrumenter,
+        error_handler: @error_handler,
       )
 
       loop do


lib/kafka/produce_operation.rb
```diff
@@ -137,6 +137,9 @@ def handle_response(response)
               delay: ack_time - message.create_time,
             })
           end
+        rescue Kafka::MessageSizeTooLarge
+          @logger.error "Message is too large when writing to #{topic}/#{partition}"
+          @buffer.clear_messages(topic: topic, partition: partition)
```
dasch: This clears _all_ messages, which causes more data loss than is strictly necessary... however, I think it's a good enough solution for now. Can you add a description to the log message indicating that you're clearing _n_ messages?

tjwp: The data loss implications of this are terrifying for us. Could the buffer clearing at least be configurable? Even `CorruptMessage` does not cause the buffer to be cleared when it is raised.

dasch: I'm not sure I've ever seen CorruptMessage in the wild, though – I imagine you'd incur the same message loss since the producer wouldn't be able to deliver the buffered messages to the partition, and forcibly restarting it would clear those messages anyway.

Unfortunately, the Kafka protocol doesn't indicate _which_ message caused the problem, making it rather difficult to only clear the problematic one.

One option would be to go through each buffered message for the partition and remove the _largest one_. But it's sort of a hack...

dasch: At any rate, the policy should be set at the producer level, not the operation level. I think it's preferable to raise the error further up that stack and handle it specifically in AsyncProducer, which doesn't guarantee delivery anyway. The sync producer should never silently lose data.

jasantunes: My rationale to _clear the buffer_ was to follow the same pattern as if we'd had no errors (see below in https://github.com/jasantunes/ruby-kafka/blob/b9745a7e1a3080c225e91ebf09cbedfba7c98af1/lib/kafka/produce_operation.rb#L164).

If `@buffer.clear_messages` is clearing unsent messages from the queue, then maybe we shouldn't be clearing the _entire_ buffer on no errors either, right?

Is there a way to clear only this particular message?

dasch: I'm not 100% sure what the behavior is on the broker side when one of the messages in a batch is too large (or is corrupt) – do the rest of the messages get appended to the partition, only those up until the corrupt one, or none of them?

It's only safe to clear the buffer when the broker has acknowledged that the messages have all been appended, otherwise we're losing data. Whether we can assume that the only message left unappended will never succeed anyway and can therefore be cleared depends on how the brokers handle this situation. We'd need to look into that.

Note that I think the CorruptMessage behavior in ruby-kafka is wrong – we shouldn't just log when that happens and then clear the buffer afterwards, I'd prefer it if we crash.

dasch:
See also #124 for how to induce the error in a functional spec.

jasantunes:
Tried a different approach: Forward the exception to an external handler that can handle the error more appropriately. Since the producer has access to the failed message, there's no data loss. In the following example, I'm just removing the message from the collection so that it won't be retried:

```ruby
class CustomErrorHandler
  def on_error(e, topic, partition, messages)
    puts "Handling error #{e.message} on topic: #{topic}, partition #{partition}"
    case e.class.to_s
    when ::Kafka::MessageSizeTooLarge.to_s
      message = messages.shift # remove first message
      puts "First message size too large? #{message.bytesize} bytes"
    else
      puts "Some other error"
    end
  end
end
```

Calling the `async_producer` with a large message:
```ruby
load 't.rb'
k = Kafka.new(logger: Logger.new(STDOUT), seed_brokers: [ 'localhost'], client_id: "my-application")
handler = CustomErrorHandler.new
p = k.async_producer.with_error_handler(handler)
p.produce("A"*1500000, topic: "development_events")
p.deliver_messages
```

Output:
```
I, [2017-06-17T21:44:54.096241 #57768]  INFO -- : New topics added to target list: development_events
I, [2017-06-17T21:44:54.096335 #57768]  INFO -- : Fetching cluster metadata from kafka://localhost:9092
D, [2017-06-17T21:44:54.096468 #57768] DEBUG -- : Opening connection to localhost:9092 with client id my-application...
D, [2017-06-17T21:44:54.098587 #57768] DEBUG -- : Sending request 1 to localhost:9092
=> nil
2.2.3 (main):0 > D, [2017-06-17T21:44:54.101532 #57768] DEBUG -- : Waiting for response 1 from localhost:9092
D, [2017-06-17T21:44:54.106506 #57768] DEBUG -- : Received response 1 from localhost:9092
I, [2017-06-17T21:44:54.106574 #57768]  INFO -- : Discovered cluster metadata; nodes: localhost:9092 (node_id=0)
D, [2017-06-17T21:44:54.106604 #57768] DEBUG -- : Closing socket to localhost:9092
D, [2017-06-17T21:44:54.106863 #57768] DEBUG -- : Current leader for development_events/0 is node localhost:9092 (node_id=0)
I, [2017-06-17T21:44:54.106940 #57768]  INFO -- : Sending 1 messages to localhost:9092 (node_id=0)
D, [2017-06-17T21:44:54.107000 #57768] DEBUG -- : Opening connection to localhost:9092 with client id my-application...
D, [2017-06-17T21:44:54.109235 #57768] DEBUG -- : Sending request 1 to localhost:9092
D, [2017-06-17T21:44:54.132856 #57768] DEBUG -- : Waiting for response 1 from localhost:9092
D, [2017-06-17T21:44:54.155442 #57768] DEBUG -- : Received response 1 from localhost:9092
Handling error Kafka::MessageSizeTooLarge on topic: development_events, partition 0
First message size too large? 1500000 bytes
E, [2017-06-17T21:44:54.155718 #57768] ERROR -- : Message is too large when writing to development_events/0
W, [2017-06-17T21:44:54.155781 #57768]  WARN -- : Failed to send all messages; attempting retry 1 of 2 after 1s
D, [2017-06-17T21:44:55.161025 #57768] DEBUG -- : Current leader for development_events/0 is node localhost:9092 (node_id=0)
I, [2017-06-17T21:44:55.161156 #57768]  INFO -- : Sending 0 messages to localhost:9092 (node_id=0)
D, [2017-06-17T21:44:55.161264 #57768] DEBUG -- : Sending request 2 to localhost:9092
D, [2017-06-17T21:44:55.161638 #57768] DEBUG -- : Waiting for response 2 from localhost:9092
D, [2017-06-17T21:44:55.167220 #57768] DEBUG -- : Received response 2 from localhost:9092
D, [2017-06-17T21:44:55.167610 #57768] DEBUG -- : Successfully appended 0 messages to development_events/0
W, [2017-06-17T21:44:55.167697 #57768]  WARN -- : Failed to send all messages; attempting retry 2 of 2 after 1s
E, [2017-06-17T21:44:56.169854 #57768] ERROR -- : Failed to send all messages; keeping remaining messages in buffer
```

This approach has the following two advantages:
- The connection to the broker is not abruptly closed (since we're not propagating the exception internally),
- There's no data loss (since we're forwarding the exception to a custom handler, along with the messages)

dasch:
I feel that, if we were to go down this route, we'd need to limit the handler to a predefined set of policies – e.g. `:continue`, `:crash`, and whatever else people might want to happen. So you'd maybe do:

```ruby
producer = kafka.async_producer

producer.on_invalid_message do |exception, message|
  logger.error exception.message
  dead_letter_box.write(message)
  :continue
end
```

jasantunes:
I like the block approach. The custom handler code doesn't even need to return a specific symbol (like `:crash:` or `:continue`); it could just `raise exception` at the end (or not raise it).

Is my assumption that the first message in the [message collection](https://github.com/zendesk/ruby-kafka/pull/358/files#diff-f6b2e51b589e2f77d6ca3ef981be7e62R116) in handle_response would be the invalid message?

Also, I tried the functional testing but couldn't get the docker-api to run on my macos. Any idea how to use an existing kafka server for those tests?

dasch:
> I like the block approach. The custom handler code doesn't even need to return a specific symbol (like `:crash:` or `:continue`); it could just `raise exception` at the end (or not raise it).

The main with exceptions is that it would be raised in a background thread – so it's not obvious how to communicate e.g. a failure to the foreground. But yeah, and exception is probably the simplest approach.

> Is my assumption that the first message in the [message collection](https://github.com/zendesk/ruby-kafka/pull/358/files#diff-f6b2e51b589e2f77d6ca3ef981be7e62R116) in handle_response would be the invalid message?

That's not a valid assumption. We don't know which message caused the error.

> Also, I tried the functional testing but couldn't get the docker-api to run on my macos. Any idea how to use an existing kafka server for those tests?

How you tried the Mac installer for Docker? It works for me on Mac OS.

jasantunes:
> That's not a valid assumption. We don't know which message caused the error.

So instead of
```ruby
producer.on_invalid_message do |exception, message|
```

we'd have
```ruby
producer.on_invalid_messages do |exception, messages|
```

or more generic
```ruby
producer.on_error do |exception, messages|
```

> How you tried the Mac installer for Docker? It works for me on Mac OS.

For some reason it works on my work laptop but not my personal one. Thanks!

I'll change the PR to use a block `|exception, messages|` to handle errors. Does that sound good?

dasch:
@jasantunes I think it's an interesting approach, but I can't promise that I'll merge it – the cost of maintaining code needs to be considered and held against the problem being solved.

dasch:
Closing this – want to avoid going in this direction.

Issue #355, opened 2017-05-18
Expose heartbeat?


michaelsauter:
I have a consumer which reads messages in batches, and needs to call an external API a few times for each batch (there is a limit on the external API, and the Kafka batch size might exceed that).

In that case, I've seen the consumer session timeout and when the consumer tries to commit the offsets, the broker thinks that the consumer died already.

I see a couple of options here:

1. Reduce the amount of bytes to fetch from a partition
2. Increase the session timeout
3. Send a heartbeat while processing the batch

My question if it makes sense to do (3) ? I have already done (1), and that helps, but potentially the external API is slow and/or rate limited so I don't really know how long each API call can take so I might still have to do a heartbeat. I could do (2) but that would then increase the time until Kafka detects a consumer died. (3) sounds good to me because I can keep the timeout short, and in my code I know if I'm still alive, just not done yet.

Does that make sense? If so, should it be possible to do e.g. `consumer.send_heartbeat_if_necessary`? I experimented with this by doing `consumer.instance_variable_get("@heartbeat").send_if_necessary`, and that has the desired effect.

What do you think?

dasch:
What does your processing code look like? Is your API call synchronous?

There's a risk in decoupling the heartbeat from the actual processing, in that you may end up with a zombie consumer holding on to partitions. That's why I'm wary of exposing the heartbeat trigger directly, as well as why it doesn't happen on a background thread.

michaelsauter:
Yes, the API calls are asynchronous :/

Good point about the zombie processes. However, I agree only partially.

It makes a lot of sense not to trigger heartbeats automatically in a background thread because a) it might not be obvious that this is happening and b) as a user I cannot control it so in the end it is likely that some users would run into problems. 

But exposing an API to trigger a heartbeat on demand is a different matter I think. It is completely obvious (because it's an explicit opt-in) and as a user, I have control over it (e.g. I could have some check to allow only 3 manual heartbeats before giving up).

In our case, there is actually just one consumer in the group right now, and I know that doing one API call while processing a batch is completely fine. However, there are situations where I might need to do  more API calls, and then I want to heartbeat after each API call.
If the API I'm talking to is really really slow, it's okay to timeout though and process the messages again, that is actually desired as I want to basically spin until the API is back up.
But I don't want to spin forever because I get a batch which triggers more API calls then what the session timeout will allow me to do.

[...]

PR #361, opened 2017-06-13
Add method to send heartbeat if necessary on consumer

headshota:
As discussed in #355 implemented a method for exposing heartbeat from a consumer.

@michaelsauter FYI

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -288,6 +288,10 @@ module Kafka
       @offset_manager.mark_as_processed(message.topic, message.partition, message.offset)
     end
 
+    def send_heartbeat_if_necessary
+      @heartbeat.send_if_necessary
+    end
+
     private
 
     def consumer_loop
--- a/spec/consumer_spec.rb
+++ b/spec/consumer_spec.rb
@@ -197,4 +197,11 @@ describe Kafka::Consumer do
       end
     end
   end
+
+  describe "#send_heartbeat_if_necessary" do
+    it "sends heartbeat if necessary" do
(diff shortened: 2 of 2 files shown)


dasch:
I think the idea was to have a separate `heartbeat` method that actually triggers a heartbeat.

ale7714:
@headshota @dasch any plans on merging this? :)

lib/kafka/consumer.rb
```diff
@@ -288,6 +288,10 @@ def mark_message_as_processed(message)
       @offset_manager.mark_as_processed(message.topic, message.partition, message.offset)
     end
 
+    def send_heartbeat_if_necessary
```
dasch: I think we should probably simplify the API in order to avoid confusion – maybe it would make more sense to only allow clients to force a heartbeat manually?

dasch:
@ale7714 sorry, lost track of this. I think we need to consider what clients would expect in terms of behavior – the "send if necessary" API is not very clear from the outside. I'd prefer an API that just forces the heartbeat to be sent.

michaelsauter:
@dasch Hmm, interesting question.

Right now, the gem seems to only have this "send if necessary" behaviour though, so that was the easiest to expose without changing anything. 

Would you send the heartbeat then also if it doesn't make sense? E.g. if they happen in close succession?

The use case we have (see #355) is best served by "send if necessary" as a forced heartbeat would result in more heartbeats than necessary ...

dasch:
@michaelsauter if you have an existing use case for "send if necessary" then I think that's a better indicator than my intuition – I'll merge this as-is.

Thanks! ❤️

PR #362, opened 2017-06-19
Block on async responses in the order they become ready

dasch:
Currently, we send a fetch request to each broker before blocking on the first one while we're waiting for a response. This is inefficient when the first broker in the list is not the first to respond – which will often be the case when there's an uneven data load.

This change orders the reads by when the async responses become _ready_, as indicated by calling `IO.select` on their backing socket. After sending the requests, we'll wait for a response to become ready. If there's a timeout, we'll start waiting again. If one or more of the responses become ready, we select the first one, read it and return the data to the client, then move on to the rest of the pending responses.

This allows us to immediately process messages returned by brokers that see a lot of throughput without having to wait for the other brokers to "time out" while waiting for messages to arrive (due to `max_wait_time`).

The next logical step would be to pre-fetch messages after reading a response, but before letting the client process the messages. That's for another PR, though.

### DIFF
--- a/lib/kafka/ssl_socket_with_timeout.rb
+++ b/lib/kafka/ssl_socket_with_timeout.rb
@@ -163,6 +163,10 @@ module Kafka
       @tcp_socket.closed? || @ssl_socket.closed?
     end
 
+    def to_io
+      @tcp_socket.to_io
+    end
+
     def set_encoding(encoding)
       @tcp_socket.set_encoding(encoding)
     end
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -17,10 +17,15 @@ module Kafka
     # Use a custom "nil" value so that nil can be an actual value.
     MISSING = Object.new
 
-    def initialize(decoder, &block)
+    def initialize(decoder, io, &block)
       @decoder = decoder
+      @io = io
       @block = block
       @response = MISSING
+    end
+
+    def to_io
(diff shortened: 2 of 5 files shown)


lib/kafka/ssl_socket_with_timeout.rb
```diff
@@ -163,6 +163,10 @@ def closed?
       @tcp_socket.closed? || @ssl_socket.closed?
     end
 
+    def to_io
```
tcrayford: I am not sure this works. Just because the backing tcp socket is ready doesn't mean the ssl socket is ready. It probably works well enough in practice, but we would need to test that it does.

dasch: True. If only I could afford an SSL-enabled Kafka cluster! 😂

I think I'll hold off on merging this until I've at least had time to test that the current master works as expected in production, so as to not introduce too many new moving parts at a time.

tcrayford: seems good. Also we can probably swing you a testing cluster somehow, if you're interested.

dasch: @tcrayford that would be lovely! I'm actually thinking about running the CI suite against a more stable cluster...

dasch:
Dropping this in favor of threads. Don't let anybody tell you that async IO is easy!

PR #363, opened 2017-06-21
Batch Statsd events

dasch:
In order to minimize the performance cost of instrumenting the operation of Kafka clients we can batch up individual Statsd event rather than sending them separately.

### DIFF
--- a/lib/kafka/datadog.rb
+++ b/lib/kafka/datadog.rb
@@ -56,6 +56,10 @@ module Kafka
         end
       end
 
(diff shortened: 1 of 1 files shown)


dasch:
This doesn't seem to have any effect in production.

mensfeld:
Switch if you can to UDP, for me it was a huge difference

dasch:
That's what I think I'm using – this PR didn't have any effect on performance.

PR #369, opened 2017-07-08
Add Consumer#poll

ale7714:
Proposed usage: 

```
loop do 
  buffer = []

  fetched_batches = consumer.poll() 
  
  fetched_batches.each do |batch| 
    
      # We want to send a heartbeat if necessary
      consumer.send_heartbeat_if_necessary 
      
      event = JSON.parse(message.value)
      
      case event.fetch("type")
      when "add_to_cart"
       buffer << event
      when "complete_purchase"
        # We've received all the messages we need, time to save the transaction.
        save_transaction(buffer)
        consumer.commit_offsets
        buffer.clear
      end
    end
  end 
end
```

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -378,6 +378,10 @@ module Kafka
       @heartbeat.send_if_necessary
     end
 
+    def subscribed_partitions
+      @group.subscribed_partitions
+    end
+
     private
 
     def consumer_loop
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -170,7 +170,7 @@ module Kafka
     # subscribes to.
     #
     # Each message is yielded to the provided block. If the block returns
-    # without raising an exception, the message will be considered successfully
+    # without raising an exeption, the message will be considered successfully
     # processed. At regular intervals the offset of the most recent successfully
     # processed message in each partition will be committed to the Kafka
     # offset store. If the consumer crashes or leaves the group, the group member
@@ -359,22 +359,15 @@ module Kafka
     #   is ignored.
     # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
     #   returning messages from the server, in seconds.
+    # @param max_retries [Integer] the maximum numbers of retries to fetch messages
+    #   before raising an error
     # @return batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
-    def poll(min_bytes: 1, max_wait_time: 5)
-      batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
-      send_heartbeat_if_necessary
-      batches
-    rescue HeartbeatError, OffsetCommitError
-      join_group
-      retry
-    rescue FetchError, NotLeaderForPartition, UnknownTopicOrPartition
-      @cluster.mark_as_stale!
-      retry
-    rescue LeaderNotAvailable => e
-      @logger.error "Leader not available; waiting 1s before retrying"
-      @cluster.mark_as_stale!
-      sleep 1
-      retry
+    def poll(min_bytes: 1, max_wait_time: 5, max_retries: 20)
+      retry_or_raise(retries: max_retries) do
+        batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
+        send_heartbeat_if_necessary
+        batches
+      end
     end
 
     def commit_offsets
@@ -547,6 +540,35 @@ module Kafka
 
     def pause_for(topic, partition)
       @pauses[topic][partition]
+    end
+
+    RETRYABLE_POLL_EXCEPTIONS=[
+      HeartbeatError,
+      OffsetCommitError,
+      FetchError,
+      NotLeaderForPartition,
+      UnknownTopicOrPartition,
+      LeaderNotAvailable,
+    ]
+
+    def retry_or_raise(retries:)
+      try = 0
+      begin
+        @logger.debug "Attempt ##{try} of #{retries}"
+        yield
+      rescue *RETRYABLE_POLL_EXCEPTIONS => ex
+        if [HeartbeatError, OffsetCommitError].include?(ex.class)
+          join_group
+        elsif [FetchError, NotLeaderForPartition, UnknownTopicOrPartition].include?(ex.class)
+          @cluster.mark_as_stale!
+        elsif LeaderNotAvailable == ex.class
+          @logger.error "Leader not available; waiting 1s before retrying"
+          @cluster.mark_as_stale!
+          sleep 1
+        end
+        try += 1
+        try <= retries ? retry : raise
+      end
     end
   end
 end
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -170,7 +170,7 @@ module Kafka
     # subscribes to.
     #
     # Each message is yielded to the provided block. If the block returns
-    # without raising an exeption, the message will be considered successfully
+    # without raising an exception, the message will be considered successfully
     # processed. At regular intervals the offset of the most recent successfully
     # processed message in each partition will be committed to the Kafka
     # offset store. If the consumer crashes or leaves the group, the group member
@@ -564,11 +564,12 @@ module Kafka
         @logger.debug "Attempt ##{try} of #{retries}"
         yield
       rescue *RETRYABLE_POLL_EXCEPTIONS => ex
-        if [HeartbeatError, OffsetCommitError].include?(ex.class)
+        case ex
+        when HeartbeatError, OffsetCommitError
           join_group
-        elsif [FetchError, NotLeaderForPartition, UnknownTopicOrPartition].include?(ex.class)
+        when FetchError, NotLeaderForPartition, UnknownTopicOrPartition
           @cluster.mark_as_stale!
-        elsif LeaderNotAvailable == ex.class
+        when LeaderNotAvailable
           @logger.error "Leader not available; waiting 1s before retrying"
           @cluster.mark_as_stale!
           sleep 1
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -357,14 +357,21 @@ module Kafka
     # @param min_bytes [Integer] the minimum number of bytes to read before
     #   returning messages from the server; if `max_wait_time` is reached, this
     #   is ignored.
+    # @param max_bytes [Integer] the maximum number of bytes to read before
+    #   returning messages from each broker.
     # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
     #   returning messages from the server, in seconds.
     # @param max_retries [Integer] the maximum numbers of retries to fetch messages
     #   before raising an error
     # @return batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
-    def poll(min_bytes: 1, max_wait_time: 5, max_retries: 20)
+    def poll(min_bytes: 1, max_bytes: 10485760, max_wait_time: 5, max_retries: 20)
       retry_or_raise(retries: max_retries) do
-        batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
+        batches = fetch_batches(
+          min_bytes: min_bytes,
+          max_bytes: max_bytes,
+          max_wait_time: max_wait_time,
+          automatically_mark_as_processed: false
+        )
         send_heartbeat_if_necessary
         batches
       end
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -371,16 +371,16 @@ module Kafka
         max_wait_time: max_wait_time,
       )
 
-      @running = true
-      @fetcher.start
+      unless @running
+        @running = true
+        @fetcher.start
+      end
 
       retry_or_raise(retries: max_retries) do
         batches = fetch_batches
         send_heartbeat_if_necessary
         batches
       end
-    ensure
-      @fetcher.stop
     end
 
     def commit_offsets
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -354,6 +354,18 @@ module Kafka
       @offset_manager.seek_to(topic, partition, offset)
     end
 
+    # @param min_bytes [Integer] the minimum number of bytes to read before
+    #   returning messages from the server; if `max_wait_time` is reached, this
+    #   is ignored.
+    # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
+    #   returning messages from the server, in seconds.
+    # @return batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
+    def poll(min_bytes: 1, max_wait_time: 5)
+      batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
+      send_heartbeat_if_necessary
+      batches
+    end
+
     def commit_offsets
       @offset_manager.commit_offsets
     end
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -352,6 +352,7 @@ module Kafka
     # @return [nil]
     def seek(topic, partition, offset)
       @offset_manager.seek_to(topic, partition, offset)
+      @fetcher.seek(topic, partition, offset)
     end
 
     # @param min_bytes [Integer] the minimum number of bytes to read before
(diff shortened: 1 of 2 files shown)


lib/kafka/consumer.rb
```diff
@@ -280,6 +280,21 @@ def each_batch(min_bytes: 1, max_wait_time: 5, automatically_mark_as_processed:
...
 
+    # @param min_bytes [Integer] the minimum number of bytes to read before
+    #   returning messages from the server; if `max_wait_time` is reached, this
+    #   is ignored.
+    # @param max_wait_time [Integer, Float] the maximum duration of time to wait before
+    #   returning messages from the server, in seconds.
+    # @param topics_offsets [Hash] the offsets to fetch from. By default it will
+    #   fetch from where the offset manager defaults too.
+    # @return batch [Kafka::FetchedBatch] a message batch fetched from Kafka.
+    def poll(min_bytes: 1, max_wait_time: 5, offsets: {})
```
ale7714: i don't feel that `offsets` is the right name for what i'm going for. But my goal is to be able to fetch given a particular offset. 
I followed the docs on [`#seek`](https://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#seek) to achieve that.
i could also expose a `seek_to` method on the consumer and move this responsibility out of the 
poll method

ale7714:
I need to do some more testing on this but would like to have some initial thoughts.

ale7714:
@dasch forgot to mention you :)

dasch:

lib/kafka/consumer.rb
dasch: A `#seek` approach sounds intriguing. How would that work?

ale7714: Consumer will do something like:
```
consumer.seek_to(desired_offset)
fetched_batches = consumer.poll 
```

Then i can remove this `offsets` idea from poll. Now that i think more about this, it feels cleaner than giving the responsibility of find out where to fetch from to `#poll`.

dasch: I've been thinking about adding seeking to the normal consumer API as well, e.g.

```ruby
offsets = MyDB.find_offsets_for("my-topic")

offsets.each do |partition, offset|
  # Seeks the specified topic/partition to `offset`
  consumer.seek("my-topic", partition, offset)
end

consumer.each_message {|message| ... }
```

The main issue is whether _all_ consumers in a group need to do this or if we should propagate the specified offsets somehow – maybe using the `Sync` Kafka API.

dasch: This should also allow operators to reset consumer positions using a CLI tool, ideally.

ale7714: @dasch with `Sync` you refer to `commitSync`? 
On our case we want to be able to seek without propagating that offset until we receive the done message and ready to actually commit offsets. 
The downside of not propagating is that if the consumer goes down it could/will double process some messages. I see how using `Sync` would be preferable on most cases. 
 i think we could make `sync` should be optional. WDYT?

ale7714: also you think on this consumer could seek to any given partition/offset or only on the partitions they're subscribed to?

dasch: I mean the [SyncGroup](https://kafka.apache.org/protocol#The_Messages_SyncGroup) API that synchronizes partition assignments – basically, one of the consumers in a group is chosen as Leader and is asked to assign partitions to the others. This assignment can include custom metadata that's currently left blank in ruby-kafka, so it would be possible to pass the offset targets in there. Of course only the Leader would get to actually send the resets, and we won't know ahead of time which consumer will be chosen.

Given the fact that consumer groups need to be relatively small anyway (more than 32 would be a stretch IMO) I think it's probably not a huge deal if each consumer separately seeks _every_ partition, as long as they follow the same procedure. This would also mean that we don't _need_ to coordinate, as a consumer will just use the local seek targets with the partitions assigned to it. The assignments may change at any time, so it needs access to all the offsets (or be able to coordinate, e.g. via SyncGroup).

ale7714: @dasch yeah i see what you mean. I think as a first implementation every consumer using the local seek could be enough. 

> Given the fact that consumer groups need to be relatively small anyway (more than 32 would be a stretch IMO)

i'm a little bit intrigued this? have you experienced instability or issues with groups larger than 32?

dasch: Mostly in testing, but really there's an upper practical limit around that number. Any one consumer behaving erratically impacts the entire group, and that'll happen from time to time – the more consumers, the more frequently it'll happen. As ruby-kafka has itself become better at stabilizing groups this is perhaps not as big a problem as it used to be.

dasch:
Just to post the current status here at the bottom: I think we should pursue a `seek` based API instead.

ale7714:
@dasch i updated the PR, so now `#poll` is only responsible for fetching messages. Client can `#seek` to a specific offset is desired. 

```ruby
topic = "purchase-started"
offsets = { "0" => 42, "1" =>  7 }

offsets.each do |partition, offset|
  consumer.seek(topic, partition, offset)
end

loop do 
  buffer = []

  fetched_batches = consumer.poll
  
  fetched_batches.each do |batch| 
    
      # We want to send a heartbeat if necessary
      consumer.send_heartbeat_if_necessary 

      # Keep track of current TopicPartition offsets 
      offsets[message.topic] ||= {}
      offsets[message.topic][message.partition] = message.offset 
      
      event = JSON.parse(message.value)
      
      case event.fetch("type")
      when "add_to_cart"
       buffer << event
      when "complete_purchase"
        # We've received all the messages we need, time to save the transaction.
        save_transaction(buffer)
        consumer.commit_offsets
        buffer.clear
      end
    end
  end 
end
```

dasch:
This would have to go through a lot of testing – offset management is really tricky...

ale7714:
@dasch yes definitely! i will test it on my side. Any suggestion on how I can help to test this on the gem context?

dasch:
@ale7714 the existing functional tests _should_ cover most of the typical problems, but nothing beats running things in production for a few days...

ale7714:
@dasch do you think is valuable moving `Consumer#seek` into a separate PR? i feel `#poll` is a more risky change. I haven't yet test this fully on production will let you know when i do that.

dasch:
@ale7714 I do – I think `seek` makes sense on its own.

ale7714:
@dasch this PR is only the `Consumer#poll` change and a I opened separate PR  for `Consumer#seek` here https://github.com/zendesk/ruby-kafka/pull/386. We've been using on our fork `Consumer#seek`  in production for a while.

ale7714:
@dasch definitely! I'm planning to test this on prod this week hopefully or next week. Will let you know! Thanks

ale7714:
@dasch :wave: little update on this. Other things have come to my plate that have prevented me from testing this. But i will try to make it a priority. Just wanted to let you know that this is not abandon.

lib/kafka/consumer.rb
```diff
@@ -309,6 +321,10 @@ def send_heartbeat_if_necessary
       @heartbeat.send_if_necessary
     end
 
+    def subscribed_partitions
+      @group.subscribed_partitions
+    end
```
dasch: Remind me again why this is necessary?

ale7714: @dasch because we hold offsets in a buffer for longer than a fetch, which means that the consumer could have been assigned a different partition. We need to check if this consumer instance is still subscribe to that partition and i use this to do this.

```
      when "complete_purchase"
        # We've received all the messages we need, time to save the transactions
        save_transaction(buffer)
        if consumer.subscribed_partitions.include?(message.partition)
          consumer.commit_offsets
        end
        buffer.clear
      end
```

Maybe it's better to this on some kind of callback?

dasch: Hmm. Nah, I think it's OK.

ale7714:
just wanted to give a quick update that i'm actually testing this change in-prod and will be tweaking

lib/kafka/consumer.rb
```diff
@@ -307,6 +307,17 @@ def poll(min_bytes: 1, max_wait_time: 5)
       batches = fetch_batches(min_bytes: min_bytes, max_wait_time: max_wait_time)
       send_heartbeat_if_necessary
       batches
+    rescue HeartbeatError, OffsetCommitError
+      join_group
+      retry
+    rescue FetchError, NotLeaderForPartition, UnknownTopicOrPartition
+      @cluster.mark_as_stale!
+      retry
+    rescue LeaderNotAvailable => e
```
dasch: `e` is not used for anything here.

lib/kafka/consumer.rb
```diff
@@ -307,6 +307,17 @@ def poll(min_bytes: 1, max_wait_time: 5)
...
+      @logger.error "Leader not available; waiting 1s before retrying"
+      @cluster.mark_as_stale!
+      sleep 1
+      retry
```
dasch: Do we want to retry indefinitely?

ale7714: that's a good question. the reason i was thinking to retry is to let the consumer rebalance while they're trying to poll. But maybe having a max number of retries would be a better approach. What do you think about 10 retries?

ale7714:
Running this on-prod, so far is behaving as expected :). Let me know what do you think about the retry strategy.

ale7714:
Update: after rebasing with master and trying on production, we got a lot of 
> Invalid offset for some-topic/1, resetting to default offset
i think something may have a changed slightly on the offset management side.

lib/kafka/consumer.rb
```diff
@@ -474,5 +501,34 @@ def fetch_batches(min_bytes:, max_bytes:, max_wait_time:, automatically_mark_as_
...
+      rescue *RETRYABLE_POLL_EXCEPTIONS => ex
+        if [HeartbeatError, OffsetCommitError].include?(ex.class)
+          join_group
+        elsif [FetchError, NotLeaderForPartition, UnknownTopicOrPartition].include?(ex.class)
+          @cluster.mark_as_stale!
+        elsif LeaderNotAvailable == ex.class
+          @logger.error "Leader not available; waiting 1s before retrying"
+          @cluster.mark_as_stale!
+          sleep 1
+        end
```
dasch: This could be done more elegantly as a `case` expression:

```ruby
case ex
when HeartbeatError, OffsetCommitError
  join_group
when FetchError, NotLeaderForPartition, UnknownTopicOrPartition
  @cluster.mark_as_stale!
when LeaderNotAvailable
  # ...
end
```

dasch:
I'll take a look when I get the chance.

ale7714:
It looks like `FetchedBatch` is reporting the wrong `highwater_mark_offset`: 
```
some_service_1  | D, [2017-11-26T19:23:06.389813 #1] DEBUG -- : Fetching batch from test-topic/0 starting at offset 20
some_service_1  | D, [2017-11-26T19:23:06.390418 #1] DEBUG -- : Sending fetch API request 11 to kafka:9092
some_service_1  | D, [2017-11-26T19:23:06.391145 #1] DEBUG -- : Waiting for response 11 from kafka:9092
some_service_1  | D, [2017-11-26T19:23:07.992063 #1] DEBUG -- : Received response 11 from kafka:9092
some_service_1  | D, [2017-11-26T19:23:07.992407 #1] DEBUG -- : Fetched Batch start #<Kafka::FetchedBatch:0x00561fad736040 @topic="test-topic", @partition=0, @highwater_mark_offset=24, @messages=[#<Kafka::FetchedMessage:0x00561fad7364c8 @value="\x87\x01\x00\x00\x02type\x00\t\x00\x00\x00document\x00\x02", @key=nil, @topic="test-topic", @partition=0, @offset=23, @create_time=nil>, #<Kafka::FetchedMessage:0x00561fad7363d8 @value="\xD9\x01\x00\x00\x02type\x00\t\x00\x00\x00document\x00\x020", @key=nil, @topic="test-topic", @partition=0, @offset=24, @create_time=nil>, #<Kafka::FetchedMessage:0x00561fad736270 @value="\x9C\x01\x00\x00\x02type\x00\t\x00\x00\x00document\x00\x02", @key=nil, @topic="test-topic", @partition=0, @offset=25, @create_time=nil>, #<Kafka::FetchedMessage:0x00561fad736108 @value="\xEC\x01\x00\x00\x02type\x00\t\x00\x00\x00document\x00\x02", @key=nil, @topic="test-topic", @partition=0, @offset=26, @create_time=nil>]>
```

Or i wonder if `FetchedMessage` are parsing the wrong offset. Maybe related to this https://github.com/zendesk/ruby-kafka/pull/486. So i'm gonna tweak this and report back

dasch:
Could be – this looks fishy. `highwater_mark_offset` is read directly off the API response, so hopefully it's correct.

ale7714:
Looking at the logs, actually I think `highwater_mark_offset` is correct but the offset on the messages is not on some cases. When I fetch a batch with just one message the `FetchedMessage` has the right offset. But when I fetch a batch with more than one message the offset on the `FetchedMessage`  is off.

dasch:
Could be because of this:

https://github.com/zendesk/ruby-kafka/blob/fbc4bbf7057c789df3a0582ce4c25602abcb1759/lib/kafka/protocol/message.rb#L53-L73

dasch:
Are you using compression?

ale7714:
Thanks. Yes we're using gzip compression. I'm going to add some debugging around it to understand what's happening.

ale7714:
@dasch after adding a lot of debugging and trying multiple times, it seems like the offset we're using on https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/protocol/message.rb#L64 
it's the offset of the last message instead of the offset for the first message.

ale7714:
@dasch after reading a bit more. I think this is intended that the MessageSet is suppose to hold the offset of the last message on the offset field. 

Some references: 
- https://issues.apache.org/jira/browse/KAFKA-506?focusedCommentId=13464469
- http://grokbase.com/t/kafka/dev/131ywhrj10/question-about-offsets#top

dasch:
Those are old posts – my understanding was that in Kafka 0.10+, the message that wraps a compressed message set (it's all a bit confusing) would dictate the starting offset, and each inner message would get that plus its index in the message set. That may be wrong.

The Kafka protocol is very poorly documented, unfortunately – you may have to dig into the official client's code base.

ale7714:
@dasch I think that's true for messages using the new format. But even looking at the current implementation, for "legacy batch" it still looks like the offset of the the outer record is the last offset: 

https://github.com/apache/kafka/blob/6f1d4c693307d1c5d1a178d56e403bb7084d537f/clients/src/main/java/org/apache/kafka/common/record/AbstractLegacyRecordBatch.java#L40-L54

dasch:
ruby-kafka only supports magic values 0 and 1, though...

dasch:
I don't see how the comment relates to the offsets...

ale7714:
@dasch i guess what i was trying to point out if that on the java client for magic values 0 and 1, the offset on the "MessageSet" is the last offset.

ale7714:
What do you think would be the best approach to move this forward? Even without my changes, but looping trough the offset using the master branch they seem to have the wrong offset.

dasch:
If you could replicate this problem with a functional test in this repo, it would be easier to see how Kafka behaves in various versions.

dasch:
It looks like this class is doing the magic stuff in the Java client: https://github.com/apache/kafka/blob/6f1d4c693307d1c5d1a178d56e403bb7084d537f/clients/src/main/java/org/apache/kafka/common/record/AbstractLegacyRecordBatch.java#L310-L387

So there's some work involved in figuring out the right "base" offset for a message set, but then you can iterate through each message in the set and do `offset = base_offset + index`. 

It furthermore looks like the base offset must be non-positive (negative or zero). It's all a bit weird.

ale7714:
@dasch yeah i agree, my PR is very naive about figuring out the base offset. I'm wrapping up some stuff before the holidays and will try to get to this soon.

dasch:
@ale7714 should I still consider this open?

ale7714:
@dasch I still would like to merge this upstream but need to update my fork. I can reopen the PR if you prefer this one to be close. Feel free to close it :).

dasch:
@ale7714 would you be interested in rebasing this?

ale7714:
Yes, I'll do. Thanks for the ping :)

ale7714:
I rebased and doing some testing on our service. But looking mostly good.

lib/kafka/consumer.rb
```diff
@@ -352,6 +352,36 @@ def each_batch(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automaticall
...
+        max_wait_time: max_wait_time,
+      )
+
+      unless @running
+        @running = true
+        @fetcher.start
+      end
+
+      retry_or_raise(retries: max_retries) do
+        batches = fetch_batches
```
dasch: Won't this risk dropping messages if `send_heartbeat_if_necessary` raises an exception? At the very least you'd need to stop and reset the fetcher, as it would keep pushing new fetch responses into the queue.

bai:
Hey 👋Is this still something you intent to merge?

ale7714:
@bai I definitely would love to but it's not really my focus right now. I think I may need to close this PR and pick up when I'm ready. If anyone wants to pick up though, I welcome the help.

PR #370, opened 2017-07-11
Working SASL_PLAIN

trthomps:
This solves #119

### DIFF
--- a/lib/kafka/sasl_plain_authenticator.rb
+++ b/lib/kafka/sasl_plain_authenticator.rb
@@ -28,9 +28,7 @@ module Kafka
       begin
         msg = @decoder.bytes
         raise Kafka::Error, 'SASL PLAIN authentication failed: unknown error' unless msg
-      rescue Errno::ETIMEDOUT => e
-        raise Kafka::Error, "SASL PLAIN authentication failed: #{e.message}"
-      rescue EOFError => e
+      rescue Errno::ETIMEDOUT, EOFError => e
         raise Kafka::Error, "SASL PLAIN authentication failed: #{e.message}"
       end
       @logger.debug 'SASL PLAIN authentication successful.'
--- a/lib/kafka/sasl_plain_authenticator.rb
+++ b/lib/kafka/sasl_plain_authenticator.rb
@@ -23,15 +23,17 @@ module Kafka
       # SASL PLAIN
       msg = [@authzid.to_s,
              @username.to_s,
-             @password.to_s].join("\000").force_encoding("utf-8")
-      @encoder.write([msg.bytesize].pack("l>")+msg)
+             @password.to_s].join("\000").force_encoding('utf-8')
+      @encoder.write_bytes(msg)
       begin
         msg = @decoder.bytes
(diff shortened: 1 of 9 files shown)


dasch:

dasch:
@trthomps can you rebase on the latest master and force push?

lib/kafka/sasl_plain_authenticator.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      unless response.error_code == 0 && response.enabled_mechanisms.include?(PLAIN_IDENT)
+        raise Kafka::Error, "#{PLAIN_IDENT} is not supported."
+      end
+
+      # SASL PLAIN
+      msg = [@authzid.to_s,
+             @username.to_s,
+             @password.to_s].join("\000").force_encoding("utf-8")
+      @encoder.write([msg.bytesize].pack("l>")+msg)
```
dasch: Would `@encoder.write_bytes(msg)` not work here?

lib/kafka/sasl_plain_authenticator.rb
```diff
@@ -0,0 +1,37 @@
...
+      begin
+        msg = @decoder.bytes
+        raise Kafka::Error, "SASL PLAIN authentication failed." unless msg
+      rescue Errno::ETIMEDOUT
+        raise Kafka::Error, "SASL PLAIN authentication failed." unless msg
```
dasch: The last error message should probably include the reason why it failed, e.g.

> SASL PLAIN authentication timed out

lib/kafka/sasl_plain_authenticator.rb
trthomps: I had that but for whatever reason it wouldn't work.

lib/kafka/sasl_plain_authenticator.rb
trthomps: Good call.

lib/kafka/sasl_plain_authenticator.rb
dasch: `#write_bytes` should do exactly what you're doing here (https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/protocol/encoder.rb#L90), with the only difference being when `msg` is `nil` – which it definitely shouldn't be here. Can you try again?

lib/kafka/sasl_plain_authenticator.rb
```diff
@@ -0,0 +1,37 @@
+module Kafka
+  class SaslPlainAuthenticator
```
dasch: I'd like to see a test that exercises this code, maybe against a fake SASL server implementation. Currently none of this code is run during testing.

trthomps: Agreed.  I broke the `FakeServer` class out and added support for SASL PLAIN authentication.

trthomps:
It doesn't look like the failed functional tests are related to this PR.

tmeinlschmidt:
@trthomps good job!

PR #372, opened 2017-07-12
Retry subsequent messages for a topic after failing to assign a partition

tjwp:
When a message fails to receive a partition, all subsequent messages for the same topic should be queued for retry because it is possible that they should follow the message that failed to receive a partition and we do not want to lose ordering.

We were intermittently getting messages out of order in our tests due to this bug.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -334,11 +334,19 @@ module Kafka
 
     def assign_partitions!
       failed_messages = []
+      topics_with_failures = Set.new
 
       @pending_message_queue.each do |message|
         partition = message.partition
 
         begin
+          # If a message for a topic fails to receive a partition all subsequent
+          # messages for the topic should be retried to preserve ordering
+          if topics_with_failures.include?(message.topic)
+            failed_messages << message
+            next
+          end
+
           if partition.nil?
             partition_count = @cluster.partitions_for(message.topic).count
             partition = Partitioner.partition_for_key(partition_count, message)
@@ -357,6 +365,7 @@ module Kafka
             exception: [e.class.to_s, e.message],
           })
 
+          topics_with_failures << message.topic
           failed_messages << message
         end
       end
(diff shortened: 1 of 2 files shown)


tjwp:
I'm unable to reproduce the Circle test failure locally.

dasch:

lib/kafka/producer.rb
```diff
@@ -357,6 +365,7 @@ def assign_partitions!
             exception: [e.class.to_s, e.message],
           })
 
+          topics_with_failures << message.topic
           failed_messages << message
```
dasch: I don't see how this fixed the problem – you'd have to wait with writing the messages to `@buffer` until a partition has been assigned for each message in a topic.

dasch:
@tjwp is this something you've seen with real workloads? What was the cause of the partition assignment failure?

lib/kafka/producer.rb
tjwp: Other messages for the same topic where a message failed to receive a partition are kept in `failing_messages` which is ultimately copied back to `@pending_message_queue`.

In the loop in `#deliver_messages_with_retries`, the `@pending_message_queue` size is included in `#buffer_size`. Since the size is still non-zero the loop is retried (if still below the max retries configuration).

So the partition assignment will be retried for the originally failing message and all subsequent messages for the topic.

tjwp:
I believe that we have only seen this failure in our integration specs that use ruby-kafka. Some of those specs rely on the topic being deleted in a before hook, and then auto-created during the example. It appears that the topic is not immediately available so the first partition assignment fails.

I think it is unlikely that we would be affected by this problem in a production environment, but it seems like an edge case worth handling to prevent out-of-order delivery, particularly if any other intermittent problem could prevent a partition assignment.

lib/kafka/producer.rb
dasch: Ah, now I get it.

dasch:
:+1: :heart:

PR #382, opened 2017-07-26
Experimental thread based Consumer

dasch:
Work in progress.


mensfeld:
Just a question: wouldn't it be easier to use celluloid to manage thread based consumers?

dasch:
@mensfeld not really – this is pretty simple stuff, and it's already being done in the async producer. Also, I'm wary of adding big dependencies to a low-level library.

The biggest challenge will be designing the damn thing and refactoring the code base.

nguyenquangminh0711:
The threaded fetcher is already implemented and available in newest stable branch. This PR is obsoleted. So, I'll close it.

Issue #288, opened 2016-10-20
busy error log spamming with unreachable kafka node


MartinNowak:
If this is a bug report, please fill out the following:
- Version of Ruby: 2.2.4
- Version of Kafka: 0.9.0.1
- Version of ruby-kafka: 0.3.15

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.
###### Steps to reproduce

``` ruby
kafka = Kafka.new(
  seed_brokers: ['kafka1:9092', 'kafka2:9092', 'kafka3:9092'],
  logger: logger)
consumer.subscribe('topic')
consumer.each_batch(min_bytes: 64 * 1024, max_wait_time: 5) do |kafka_batch|
  kafka_batch.messages.each do |msg|
    # process message
  end
end
```
###### Expected outcome

When one of the Kafka nodes went down, the consumer should have logged an error and tried to reconnect to the cluster/find the current coordinator.
###### Actual outcome

While restarting one of the Kafka replicas (might have been the leader for the partition), the consumer spammed the error log w/ about 2 messages / 4 ms (500 msg / s).

```
2016-10-20 02:17:40.753 cassandra-writer
Connection error while fetching messages: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.749 cassandra-writer
Failed to connect to kafka3:9092: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.749 cassandra-writer
Connection error while fetching messages: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.745 cassandra-writer
Failed to connect to kafka3:9092: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.745 cassandra-writer
Connection error while fetching messages: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.741 cassandra-writer
Failed to connect to kafka3:9092: Connection refused - connect(2) for 10.1.24.5:9092
2016-10-20 02:17:40.741 cassandra-writer
Connection error while fetching messages: Connection refused - connect(2) for 10.1.24.5:9092
```

The error log end with this before the consumer apparently managed to reconnect to the restarted server or another one.

```
2016-10-20 02:17:46.167 cassandra-writer
Failed to find coordinator for group `cassandra_writer`; retrying...
2016-10-20 02:17:46.138 cassandra-writer
[...]

PR #383, opened 2017-08-03
#288 - busy error log spamming with unreachable kafka node

mensfeld:
Related to #288 
AS IS: When we kill last Kafka process in a cluster with kill -9, the connection is being retried constantly, spamming about unreachable kafka cluster in the error log. When a cluster is unreachable for a longer period of time, it results in 10s of GB in the error log.

TO BE: With a simple sleep, we will still be notified and it will still retry to connect, but the frequency will be much lower, which means that it won't spam the error log.


dasch:
:heart:

PR #385, opened 2017-08-07
show the underlying error on metadata fetching issues

tcrayford:
Right now ruby-kafka reports `ConnectionError` when initializing
metadata from a cluster. This can happen just because the metadata
response timed out being read from the broker.

Instead, as well as logging the error and which broker it was regarding,
actually output that in the `ConnectionError` exception

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -193,6 +193,7 @@ module Kafka
     # @raise [ConnectionError] if none of the nodes in `seed_brokers` are available.
     # @return [Protocol::MetadataResponse] the cluster metadata.
     def fetch_cluster_info
+      errors = []
       @seed_brokers.shuffle.each do |node|
         @logger.info "Fetching cluster metadata from #{node}"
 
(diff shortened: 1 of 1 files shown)


lib/kafka/cluster.rb
```diff
@@ -206,13 +207,14 @@ def fetch_cluster_info
...
+          broker_error = "Failed to fetch metadata from #{node}: #{e}"
+          @logger.error broker_error
         ensure
           broker.disconnect unless broker.nil?
         end
       end
 
-      raise ConnectionError, "Could not connect to any of the seed brokers: #{@seed_brokers.join(', ')}"
+      raise ConnectionError, "Could not connect to any of the seed brokers: #{@seed_brokers.join(', ')}: #{broker_error}"
     end
```
dasch: Would it make more sense to include _all_ the errors, e.g.

```ruby
errors = @seed_brokers.shuffle.map {|node|
  begin
    # ...
  rescue Error => e
    @logger.error "..."
    [node, e]
  else
    return cluster_info
  ensure
    # ...
  end
}

error_description = errors.map {|node, exception| "- #{node}: #{exception}" }.join("\n")

raise ConnectionError, "Could not connect to any of the seed brokers:\n#{error_description}"
```

tcrayford: ah nice! We use ruby-kafka for healthchecking individual brokers, but you're right

tcrayford: @dasch sorry I was out of the loop here a bit. Pushed now.

PR #386, opened 2017-08-07
Add Consumer#seek

ale7714:

PR #388, opened 2017-08-09
Fix OffsetManager#seek_to_default

ale7714:
* Add tests to OffsetManager

Follow up to https://github.com/zendesk/ruby-kafka/pull/386, while adding tests to `OffsetManager`  I realized the my change broke `#seek_to_default`  😓.

### DIFF
--- a/spec/offset_manager_spec.rb
+++ b/spec/offset_manager_spec.rb
@@ -203,4 +203,34 @@ describe Kafka::OffsetManager do
       expect(group).to have_received(:commit_offsets).with(expected_offsets)
     end
   end
+
+  describe "#seek_to" do
+    it "seeks to given topic-partition/offset" do
+      topic = "greetings"
+      partition = 0
+      seek_offset = 14
+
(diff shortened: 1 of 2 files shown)


spec/offset_manager_spec.rb
```diff
@@ -203,4 +203,29 @@ def partition_offset_info(offset)
...
+      expect(offset).to eq seek_offset
+    end
+  end
+
+  describe "#seek_to_default" do
+    it "delegates to seek_to" do
+      topic = "greetings"
+      partition = 0
+
+      expect(offset_manager).to receive(:seek_to).with(topic, partition, -1)
```
dasch: I'd rather avoid setting call expectations – this is easily testable by calling `next_offset_for`.

ale7714: @dasch good point! updated the test accordingly

dasch:
:heart:

ale7714:
Not sure that the spec fail relates to my code change

dasch:
Restarting the tests fixed the issue. They're a bit flaky.

PR #391, opened 2017-08-13
Allow to lazily load "gssapi" gem

ale7714:
* Do not require gssapi as a runtime gem

### DIFF
--- a/lib/kafka/sasl_gssapi_authenticator.rb
+++ b/lib/kafka/sasl_gssapi_authenticator.rb
@@ -1,5 +1,3 @@
-require 'gssapi'
-
 module Kafka
   class SaslGssapiAuthenticator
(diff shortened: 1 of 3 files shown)


lib/kafka/sasl_gssapi_authenticator.rb
```diff
@@ -59,6 +58,15 @@ def send_and_receive_sasl_token
       @decoder.bytes
     end
 
+    def load_gssapi
+      begin
+        require "gssapi"
+      rescue LoadError
+        $stderr.puts "In order to use GSSAPI authentication you need to install the `gssapi` gem."
```
dasch: This should probably go to the logger.

ale7714: @dasch done :)

dasch:
:heart:

PR #394, opened 2017-08-15
higher level libraries listed in the readme proper section

mensfeld:
Hello, I've noticed, that more and more often people ask for things not directly related to ruby-kafka itself (as a lower level driver). I propose a proper section, where all the tools that are ruby-kafka dependent and that have stable development are listed.

Please move it to the top / bottom accordingly to your view on where section like that should be.

AS IS: There is no section that would point people to solutions that are ruby-kafka dependent, but that provide a higher level abstractions.

TO BE: A ```Higher level libraries``` section with all the frameworks and tools that are based on ruby-kafka (that I know), but that provide some additional functionalities and that can be used "out of the box" without having to understand how things work on a ruby-kafka and kafka level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Although parts of this library work with Kafka 0.8 – specifically, the Produce
     3. [Consumer Design](#consumer-design)
 5. [Development](#development)
 6. [Roadmap](#roadmap)
+7. [Higher level libraries](#higher-level-libraries)
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


mensfeld:
@dasch  I highly doubt, that the ci fail is related to my changes :D

dasch:
@mensfeld the CI is a bit flaky, unfortunately.

:+1:

Issue #395, opened 2017-08-15
EOFError when the connection was closed by broker because of idle client


Elyahou:
client connections are being closed by broker after some time, defined by `connections.max.idle.ms` config on the broker (default to 9 min).

I noticed that such closed connections are still being used by ruby-kafka when producing message, and this cause EOFError exceptions, and sync retry after backoff time.

I saw that an attempt to fix this had already been added here https://github.com/zendesk/ruby-kafka/pull/242, but it seems this fix don't work, since the `closed?`return false for these half closed socket.

Maybe the client also need to check this socket expiration and implicitly close the connection when time is expired

dasch:
@Elyahou the problem is that the timeout is configurable on the broker side, but we won't know the actual value. I've thought about sending some kind of keep-alive request periodically, but I'm not sure if that's the best solution, as it would add unnecessary traffic.

Elyahou:
I know Java client don't have this issue since 0.9 version. Maybe we should
mimic the same behavior as this client ?

dasch:
Can you investigate what they're doing in order to avoid the problem? They may do client-side socket cleanup with a lower timeout than the brokers, but I'm not sure if it's bullet-proof.

dasch:
Fixed by #399!

PR #399, opened 2017-08-25
Automatically reopen idle connections

dasch:
Connections that have been idle for more than five minutes are automatically reopened. This is done in order to preempt the broker-side connection killer, which by default kicks in after 10 minutes of idleness.


Elyahou:
cool ! it will surely fix https://github.com/zendesk/ruby-kafka/issues/395

dasch:
@Elyahou it did fix it!

PR #401, opened 2017-08-30
Try to authenticate after opening a new socket connection 

sparrovv:
After idle time of 5min ruby kafka tries to reconnect as the connection might be closed by the broker. 
Since it's not calling authentication logic after opening a new socket, sending messages is failing for any client that needs to authenticate:

``` 
I, [2017-08-29T09:46:48.447161 #8528]  INFO -- : Sending 1 messages to kafka-staging-0.aws:9093 (node_id=100) 
E, [2017-08-29T09:46:48.473428 #8528] ERROR -- : Could not connect to broker kafka-staging-0.aws:9093 (node_id=100): Connection error: end of file reached 
```

@dasch Let me know what do you think about it

### DIFF
--- a/lib/kafka/authenticator.rb
+++ b/lib/kafka/authenticator.rb
@@ -1,0 +1,56 @@
+require 'kafka/sasl_gssapi_authenticator'
+require 'kafka/sasl_plain_authenticator'
+
+module Kafka
+  class Authenticator
+    def initialize(logger:, sasl_gssapi_principal:, sasl_gssapi_keytab:, sasl_plain_authzid:, sasl_plain_username:, sasl_plain_password:)
+      @logger = logger
+      @sasl_gssapi_principal = sasl_gssapi_principal
+      @sasl_gssapi_keytab = sasl_gssapi_keytab
+      @sasl_plain_authzid = sasl_plain_authzid
+      @sasl_plain_username = sasl_plain_username
+      @sasl_plain_password = sasl_plain_password
+    end
+
+    def authenticate!(connection)
+      if authenticate_using_sasl_gssapi?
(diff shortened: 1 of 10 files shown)


lib/kafka/authenticator.rb
```diff
@@ -0,0 +1,56 @@
+require 'kafka/sasl_gssapi_authenticator'
+require 'kafka/sasl_plain_authenticator'
+
+module Kafka
+  class Authenticator
```
dasch: Let's call this e.g. SaslAuthenticator.

lib/kafka/connection_builder.rb
```diff
@@ -10,11 +9,14 @@ def initialize(client_id:, logger:, instrumenter:, connect_timeout:, socket_time
...
-      @sasl_plain_username = sasl_plain_username
-      @sasl_plain_password = sasl_plain_password
+      @authenticator = Authenticator.new(
+        sasl_gssapi_principal: sasl_gssapi_principal,
+        sasl_gssapi_keytab: sasl_gssapi_keytab,
+        sasl_plain_authzid: sasl_plain_authzid,
+        sasl_plain_username: sasl_plain_username,
+        sasl_plain_password: sasl_plain_password,
+        logger: @logger
+      )
```
dasch: Can you move this initialization of Authenticator to Client and pass in the object to ConnectionBuilder as a parameter? That way we don't need to ferry all these arguments around.

sparrovv:
@dasch I've addressed the comments in fa2f5a2

dasch:
I'll try to add more testing for this later in the week, will probably release a new version on Friday or Monday.

PR #405, opened 2017-09-07
Add Topic to Datadog Metrics

djiang:
I noticed that the topic was extracted from the event, but it wasn't passed into the tags for datadog. Just adding them in the places where it was missing.

### DIFF
--- a/lib/kafka/datadog.rb
+++ b/lib/kafka/datadog.rb
@@ -249,7 +249,6 @@ module Kafka
       def drop_messages(event)
         tags = {
           client: event.payload.fetch(:client_id),
-          topic: event.payload.fetch(:topic),
         }
 
         message_count = event.payload.fetch(:message_count)
--- a/lib/kafka/datadog.rb
+++ b/lib/kafka/datadog.rb
@@ -145,6 +145,7 @@ module Kafka
 
         tags = {
           client: client,
+          topic: topic,
         }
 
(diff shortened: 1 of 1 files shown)


lib/kafka/datadog.rb
```diff
@@ -247,6 +249,7 @@ def buffer_overflow(event)
       def drop_messages(event)
         tags = {
           client: event.payload.fetch(:client_id),
+          topic: event.payload.fetch(:topic),
```
dasch: This event doesn't currently include the topic, since that would entail introspecting the buffer. I'd skip it for now, or update this part of the code to figure out the per-topic counts and instrument an event for each topic:

https://github.com/zendesk/ruby-kafka/blob/5d46d9eba3b7dd7852ed6d02c3c6e8b648d06db3/lib/kafka/async_producer.rb#L210-L212

djiang: I'll just remove this for now -- I was mainly interested in recording the topic for `enqueue_message`

lib/kafka/datadog.rb
```diff
@@ -145,6 +145,7 @@ def produce_message(event)
 
         tags = {
           client: client,
+          topic: topic,
```
djiang: Thought I'd add it for `produce_message` also since the variable was available.

PR #410, opened 2017-09-08
use SizedQueue for AsyncProducer

nerdrew:


dasch:
I'm not sure this is an improvement – now you have to learn about SizedQueue, and the second argument to `push` is relatively new, I think. I may revisit this if we ever need to block the calling thread on `produce`, but until then I'd prefer to keep things simple.

nerdrew:
:/

nerdrew:
I have another change that adds a `ShuttingDownError` and then raises in `ensure_threads_running!` if the process is shutting down. Is it worth opening a PR for it or will it be rejected as well?

nerdrew:
The other commit is on top of this one, so the changes are mixed together, but here's the other change: https://github.com/zendesk/ruby-kafka/compare/master...nerdrew:nerdrew/shutdown-error?expand=1

dasch:
Is there a bug that is fixed by that?

nerdrew:
It gives a more meaningful error. We are using kafka for logging at square. If a process is shutting down and tries to log in an ensure block, it gets an unhelpful `ThreadError: cannot alloc thread` (or something like that). If there are lots of things that continue trying to log, it will try and fail to spawn many threads. This short circuits the failed thread creations and returns a meaningful error.

dasch:
In that case, can you make that change as a separate PR, or is it dependant on SizedQueue? I'd prefer to avoid using exception for flow control whenever possible, but I definitely think SizedQueue could be used at some point to provide a blocking `produce` method.

nerdrew:
Nah, the shutting down error doesn't depend on the `SizedQueue`. The `SizedQueue` was just me fixing an-almost-always-harmless concurrency bug in the max queue size.

Agreed re: exceptions for flow control, but I don't know if we want to get too clever checking if the process is shutting down preemptively. I'll open a PR and we can discuss that there.

nerdrew:
Discussion for shutdown specific errors: https://github.com/zendesk/ruby-kafka/pull/412

PR #412, opened 2017-09-11
WIP add ShuttingDownError

nerdrew:
Thoughts on how to test this accurately? I'm thinking of testing with `fork` and checking exit statuses.

cc @zanker @dasch

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -78,6 +78,7 @@ module Kafka
       @max_queue_size = max_queue_size
       @instrumenter = instrumenter
       @logger = logger
+      @shutting_down = false
 
       @worker = Worker.new(
         queue: @queue,
@@ -131,6 +132,7 @@ module Kafka
     # @see Kafka::Producer#shutdown
     # @return [nil]
     def shutdown
+      @shutting_down = true
       @timer_thread && @timer_thread.exit
       @queue << [:shutdown, nil]
       @worker_thread && @worker_thread.join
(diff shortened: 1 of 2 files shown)


lib/kafka/async_producer.rb
```diff
@@ -141,11 +143,17 @@ def shutdown
     private
 
     def ensure_threads_running!
+      raise ShuttingDownError if @shutting_down
```
nerdrew: What behavior do we want here? I think that `@shutting_down` should get set in `shutdown`, but then maybe only raise `ShuttingDownError` if any of the threads are dead?

dasch: Do you have a call stack when the shutdown problem happens? Maybe it would be enough to just `return` without starting the threads.

lib/kafka/async_producer.rb
```diff
@@ -141,11 +143,17 @@ def shutdown
...
+
       @worker_thread = nil unless @worker_thread && @worker_thread.alive?
       @worker_thread ||= start_thread { @worker.run }
 
       @timer_thread = nil unless @timer_thread && @timer_thread.alive?
       @timer_thread ||= start_thread { @timer.run }
+    rescue ThreadError
+      main_status = Thread.main.status
+      raise ShuttingDownError if (@shutting_down = !main_status || main_status == "aborting")
```
dasch: I'm not sure exactly what this does – can you add a comment in the code?

dasch:
Any update on this?

nerdrew:
My wife just had a baby :) I'll get back to this in a few weeks. I was distracted at work with other projects.

dasch:
@nerdrew congrats! 🎉 no worries, you've got more important things to attend to ;)

Issue #415, opened 2017-09-19
Request content-type


alexandrapersea:
When i'm writing to the kafka topic using the api like so ```curl -X POST -H "Content-Type: application/vnd.kafka.json.v2+json"  --data '{"records":[{"value":{"foo":"bar"}}]}' "http://kafka:port/topics/abc```  it works.

 When i'm trying to write to a topic using the ruby-kafka client i get 
`Failed to fetch metadata from http://kafka:port: Connection error: EOFError`.

My guess would be that the request fails because the content-type headers are not what the cluster expects. 

What are the ruby-kafka client default request content-type headers?
Could this be the reason why te request fails? If not,  am i doing something wrong, is there something I can do to make it work?

Elyahou:
ruby-kafka do not use the http protocol, but kafka protocol, and there is content-type to send when using the kafka protocol.

which seed broker do you use with ruby-kafka ?

mensfeld:
Hey. There were several similar issues. Ruby Kafka is agnostic and does not support headers.

Elyahou:
@alexandrapersea You should use `18.220.190.71:9092` in place of `http://18.220.190.71:9092`

alexandrapersea:
I have changed to protocol because of this error:

`Kafka::ConnectionError: Could not connect to any of the seed brokers:
- kafka://18.220.190.71:9092: Operation timed out`

Elyahou:
it seems your broker is not available, you can check if maybe some firewall block this port

dasch:
@alexandrapersea can you start by following these steps in the README? https://github.com/zendesk/ruby-kafka#setting-up-the-kafka-client

PR #417, opened 2017-09-21
Validate the URI scheme of seed brokers

dasch:
Due to #415 I think it would be an improvement to UX if we validate these URIs at setup time.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -14,6 +14,7 @@ require "kafka/sasl_authenticator"
 
 module Kafka
   class Client
+    URI_SCHEMES = ["kafka", "kafka+ssl"]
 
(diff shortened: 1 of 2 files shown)


dasch:
@tcrayford is this the correct scheme for Heroku?

mensfeld:
@dasch shouldn't it be a higher level lib responsibility like racecar or Karafka to validate things like that?

dasch:
They can rescue the exception and provide a nice error message, but ruby-kafka should catch the misconfiguration early.

dasch:
Note that, if more schemes were allowed in the future, the high-level libraries shouldn't have to care – it's the responsibility of ruby-kafka to turn those connection strings into actual connections.

PR #418, opened 2017-09-21
Provide access to the list of messages failed in DeliveryFailed

obrie:
In our application, we are consuming messages from an internal queueing system and attempting to publish them to Kafka through ruby-kafka.  If only a subset of those messages succeed, then we need to make sure that only those messages get deleted from our queueing system.  Since we have multiple processes running on this internal queue, we want to know which messages are still remaining (i.e. which ones failed) so that we don't delete those from our internal queue.

As an example, Amazon SQS consumers must explicitly mark messages as successful and any messages that failed will get picked up by another consumer at some point in the future.

There are a few ways we could address this:

(1) Change `deliver_messages` to return the list of messages that were successfully delivered. This doesn't really align with how the current API behaves and raises exceptions on partial failures.

(2) Provide access to the list of remaining messages that need to get sent after `deliver_messages` is invoked. e.g.

```ruby
begin
  producer.produce("hello1", topic: "greetings", partition: 0)
  producer.produce("hello1", topic: "greetings", partition: 1)
  producer.deliver_messages
rescue Kafka::DeliveryFailed
  producer.buffer_messages # Only failed messages will be here
end
```

(3) Provide access to the list of remaining messages that need to get sent in `DeliveryFailed` exceptions.  e.g.

For example:

```ruby
begin
  producer.produce("hello1", topic: "greetings", partition: 0)
  producer.produce("hello1", topic: "greetings", partition: 1)
  producer.deliver_messages
rescue Kafka::DeliveryFailed => ex
  ex.failed_messages
end
```

Since failures are the exceptional case, supporting a new API on `Producer` doesn't make quite as much sense -- we really only want to expose lists of failed messages when a failure has actually occurred.  Additionally, exposing the list of buffer messages when delivery hasn't even been attempted might result in misuse of the feature.  For these reasons, (3) seems like the most appropriate option.

In order to support this feature, 2 additional changes were made:

1. `Kafka::PendingMessage` now supports equality checks with other pending messages.  This is the object that's returned from `Kafka::DeliveryFailed#failed_messages` and providing equality checks helps us write specs against it.  If we would prefer to move that functionality into the spec itself, we can do that as well.  Open to thoughts / feedback here.

2. `Kafka::Producer#deliver_messages` had scenarios in which it could raise a `Kafka::ConnectionError` instead of `Kafka::DeliveryFailed` exception which would make it impossible to get the list of failed messages.  `#deliver_messages` will now always raise a `Kafka::DeliveryFailed` exception.

### DIFF
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -320,7 +320,11 @@ module Kafka
 
         notification[:attempts] = attempt
 
-        @cluster.refresh_metadata_if_necessary!
+        begin
+          @cluster.refresh_metadata_if_necessary!
+        rescue ConnectionError => e
+          raise DeliveryFailed, e
+        end
 
         assign_partitions!
         operation.execute
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -257,6 +257,32 @@ module Kafka
       @pending_message_queue.bytesize + @buffer.bytesize
     end
 
+    # The messages waiting to get delivered.
+    #
+    # @return [Array<Kafka::PendingMessage>]
+    def buffer_messages
+      messages = []
+
+      @pending_message_queue.each do |message|
+        messages << message
+      end
+
+      @buffer.each do |topic, partition, messages_for_partition|
+        messages_for_partition.each do |message|
+          messages << PendingMessage.new(
+            message.value,
+            message.key,
+            topic,
+            partition,
+            nil,
+            message.create_time
+          )
+        end
+      end
+
+      messages
+    end
+
     # Deletes all buffered messages.
     #
     # @return [nil]
--- a/lib/kafka/producer.rb
+++ b/lib/kafka/producer.rb
@@ -257,32 +257,6 @@ module Kafka
       @pending_message_queue.bytesize + @buffer.bytesize
     end
 
-    # The messages waiting to get delivered.
-    #
-    # @return [Array<Kafka::PendingMessage>]
-    def buffer_messages
-      messages = []
-
-      @pending_message_queue.each do |message|
-        messages << message
-      end
-
-      @buffer.each do |topic, partition, messages_for_partition|
-        messages_for_partition.each do |message|
-          messages << PendingMessage.new(
-            message.value,
-            message.key,
-            topic,
-            partition,
-            nil,
-            message.create_time
-          )
-        end
-      end
-
-      messages
-    end
-
     # Deletes all buffered messages.
     #
     # @return [nil]
@@ -323,7 +297,7 @@ module Kafka
         begin
           @cluster.refresh_metadata_if_necessary!
         rescue ConnectionError => e
-          raise DeliveryFailed, e
+          raise DeliveryFailed.new(e, buffer_messages)
         end
 
         assign_partitions!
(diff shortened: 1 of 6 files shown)


dasch:
An alternative would be to include the still-buffered messages in the DeliveryFailed exception. I'd rather have that, since I don't particularly want people messing with the internal buffer too much.

lib/kafka/producer.rb
```diff
@@ -257,6 +257,25 @@ def buffer_bytesize
...
+      messages = []
+
+      @buffer.each do |topic, partition, messages_for_partition|
+        messages_for_partition.each do |message|
+          messages << message.value
+        end
+      end
+
+      @pending_message_queue.each do |message|
+        messages << message.value
```
dasch: It's not enough to return the `value`. This PR would necessitate making some sort of message object part of the public API, which I'm not too happy about.

obrie: :+1: That makes sense

obrie:
@dasch Sure that's something I can explore.  In that scenario (where we're including it in the exception), do you think we'd still need to have a message object as part of the public API?  If so, is that something you're open to?

mensfeld:
@dasch i would strongly advise against using exceptions as it will force users to design their workflow based on exceptions (exceptions should not be used for flow control and should not be expected). Also from the ruby perspective, they are heavy as they contain stack trace and other data.

dasch:
@mensfeld This is exceptional flow, though – it only happens when delivery fails. The exception is already raised in this case. As such, I think it makes sense to include the list of failed messages in the exception itself.

@obrie PendingMessage (https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/pending_message.rb) is currently being used in the pending message queue, but Protocol::Message is being used in the buffer. I think the former is the closest thing to a publicly consumable API we have, so maybe casting the buffer messages back to PendingMessage?

obrie:
Are there cases where we the producer does *not* raise a `DeliveryFailed` exception even if it successfully produced some messages.  For example, I'm specifically looking at this line:

https://github.com/zendesk/ruby-kafka/blob/e9b220b9099f73d126fab90d9f9c241685e8fe77/lib/kafka/producer.rb#L297

If I'm not mistaken, the following can happen:

1. On first loop, some messages succeed, some fail
2. From failed messages, cluster gets marked as stale
3. On second loop, `refresh_metadata_if_necessary!` gets called
4. If `refresh_metadata_if_necessary!` fails, then a `ConnectionError` exception gets raised

I'll validate that scenario, but if someone else more familiar with the library is able to verify that, that'd be great.  If that scenario *is* valid, then I think we'd either need to expose the failed messages as-is in the current PR or we would need to catch `ConnectionError` in `Producer#deliver_messages_with_retries` and raise `DeliveryFailed` instead.

dasch:
Those are good points. It would be great if you could keep investigating.

Note that I'll be travelling the next three weeks and will be without access to my laptop, so I won't be able to merge anything during that time.

dasch:
I probably won't check in here too much, either. I think the best course of action would be for you to write a patch and deploy that version to production (using e.g. a git ref in your Gemfile). If it works I'll take a look when I'm back.

obrie:
Good timing -- I'll also be out for the next 3 weeks :smile: 

Will likely update the PR today and chat more when you're back

dasch:
Have a nice one, then! ⛱

obrie:
I've updated the PR to include the failed messages on the `DeliveryFailed` exception and to ensure that `DeliveryFailed` exceptions are always raised when calling `Producer#deliver_messages` with a partial failure.

I'll update the PR description to reflect the approaches taken and the current iteration of this feature.

obrie:
PR description updated.  Ready for another round of feedback :smile:

dasch:
This looks good! I’ll do a more thorough review when I’m back.

dasch:
Looks good! 👍

obrie:
:heart: Thanks @dasch! (and welcome back)

dasch:
Thanks!

PR #419, opened 2017-09-29
Add the ability to load client cert/cert key from file

jasondoc3:
If using `yaml` and `erb` to load a configuration for `ruby-kafka`, it is much easier to use a file path for the `ssl_client_cert` and `ssl_client_cert_key` .

Related PR: https://github.com/zendesk/ruby-kafka/pull/371

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -40,10 +40,20 @@ module Kafka
     #   to use with an SSL connection.
     #
     # @param ssl_client_cert [String, nil] a PEM encoded client cert to use with an
-    #   SSL connection. Must be used in combination with ssl_client_cert_key.
+    #   SSL connection. Must be used in combination with ssl_client_cert_key or
+    #   ssl_client_cert_key_file_path.
+    #
+    # @param ssl_client_cert_file_path [String, nil] a path on the filesystem to a PEM encoded
+    #   client cert to use with an SSL connection. Must be used in combination with
+    #   ssl_client_cert_key_file_path or ssl_client_cert_key.
     #
     # @param ssl_client_cert_key [String, nil] a PEM encoded client cert key to use with an
-    #   SSL connection. Must be used in combination with ssl_client_cert.
+    #   SSL connection. Must be used in combination with ssl_client_cert
+    #   or ssl_client_cert_file_path.
+    #
+    # @param ssl_client_cert_key_file_path [String, nil] a PEM encoded client cert key to use
+    #   with an SSL connection. Must be used in combination with ssl_client_cert_file_path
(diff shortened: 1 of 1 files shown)


dasch:
Thank you for the PR! I’m currently on vacation, but will take a look when I’m back.

lib/kafka/client.rb
```diff
@@ -477,9 +495,18 @@ def initialize_cluster
...
+      return nil unless [
+        ca_cert_file_path,
+        client_cert_file_path,
+        client_cert_key_file_path,
+        ca_cert,
+        client_cert,
+        client_cert_key
+      ].compact.any?
+
+      client_cert     ||= File.read(client_cert_file_path)
```
tcrayford: I'd recommend using the openssl library's features for reading these files off of disk instead of using `File.read`. Those methods allow e.g. reading multiple certs from a single file in a way that the mechanism used here does not.

dasch: @jasondoc3 I agree with @tcrayford. Would you be able to update your PR?

jasondoc3: Sure. @tcrayford can you elaborate a little more? I usually see the certs read like this:

```rb
OpenSSL::X509::Certificate.new File.read 'certificate.pem'
```

dasch: @jasondoc3 I can't find anything in the documentation for this, either. I'll merge this as-is.

@tcrayford please add a comment if there's an API we've missed, support can be added in a separate PR.

dasch: @jasondoc3 wait, here it is: http://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL/X509/Store.html

dasch: https://github.com/zendesk/ruby-kafka/blob/07614b0cf7fafe332db3cf055a0132921f991300/lib/kafka/client.rb#L496-L505

jasondoc3: @dasch Sorry if I am missing something, but would this be entirely necessary (or even possible) for specifying the  `ssl_client_cert` and `ssl_client_cert_key`?

From the code it appears that these are set on the `OpenSSL::SSL::SSLContext` object directly via the `cert` and `key` attributes. I think it is separate from the `cert_store`?
```rb
if client_cert && client_cert_key
  ssl_context.set_params(
    cert: OpenSSL::X509::Certificate.new(client_cert),
    key: OpenSSL::PKey.read(client_cert_key)
  )
elsif client_cert && !client_cert_key
  # ...
end
```

dasch: @tcrayford is there a suggested way to do this?

Issue #312, opened 2017-01-10
Async producer causing main process to exit


prestonwrthompson:
I have a question about the design of the async producer. 

For context, in our set up, we have our job processing code running in a main process, and most of the jobs result in at least one message that we want to produce to Kafka. From this main process, we call `produce` on the async producer.

The async producer spawns two threads, the timer thread and the worker thread. The problem we are having is that whenever there is an uncaught exception from the synchronous producer in the worker thread (specifically, we received `Kafka::Error, "Correlation id mismatch"` from [here](https://github.com/zendesk/ruby-kafka/blob/2d7812cf1dd0e18130412c9518e400347c4e5a53/lib/kafka/connection.rb#L190)), the main process crashes. We've tracked it down to the async producer setting `abort_on_exception` to true on the worker thread [here](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/async_producer.rb#L151) in `async_producer.rb`.

Here is the documentation for `abort_on_exception` for an instance of a thread:
> When set to true, causes all threads (including the main program) to abort if an exception is raised in `thr`. The process will effectively `exit(0)`.

And here is the documentation for `abort_on_exception` on the `Thread` class:
>Returns the status of the global `abort on exception` condition. The default is false. When set to true, will cause all threads to abort (the process will `exit(0)`) if an exception is raised in any thread. See also `Thread.abort_on_exception=` .

From the documentation, it seems like there's no way to avoid ruby-kafka causing the main process to`exit(0)` if there's an uncaught exception in the worker thread. Is that true or is there a way to prevent it? Can you explain the design decision behind setting `abort_on_exception` to true?

dasch:
Let me look at it Monday - this is definitely not the intended behavior. Are you using the latest version?

prestonwrthompson:
Thanks. Yes, I just tested now on 0.3.15 and master. Here are simple steps to reproduce.

```
AB-C02N4MLLG3QC:ruby-kafka preston$ irb
irb(main):001:0> require "ruby-kafka"
=> true
irb(main):002:0> kafka = Kafka::Client.new(:seed_brokers => ["kafka:9092"])
=> #<Kafka::Client:0x007ff87c57e640 [removed for readability]>
irb(main):003:0> async_producer = kafka.async_producer(:delivery_interval => 5)
=> #<Kafka::AsyncProducer:0x007ff87f00f5f8 [removed for readability]>
[...]

PR #422, opened 2017-10-04
Remove abort_on_exception in favor of error handler

film42:
This is an attempt to solve https://github.com/zendesk/ruby-kafka/issues/312 and https://github.com/zendesk/ruby-kafka/issues/357.

Instead of aborting the process when an error is raised, we can provide an error handler that allows us to do something with it. This lib will already ensure the async producer is running each time we call `#produce`, so with an error handler, there's no need to use `abort_on_exception`.

If no custom error handler is provided, a default error handler is used that will log the error and its backtrace. If you're using a service like honey badger (which is what we use), you can provide your own error handler like the following:

```ruby
error_handler = lambda { |error, _payload| ::Honeybadger.notify(error) }
producer = kafka.async_producer(error_handler: error_handler)
...
```

Now you can receive alerts when an error within your async producer occurs without your app crashing.

Here are a few questions:
1. The structure for the error handler is `lambda { |error|, payload| }` where `payload` may be `nil`. I figured it was worth having if you wanted to extract some information from it when logging errors, but it might not be very clear that the `payload` is not always available. Should I leave it or change it? Thoughts?
2. Since the logger is assigned at the instance level (not library/ class level), I create the error handler in the constructor. Don't love it, but I think it's better than making `logger` a param for the error handler. Thoughts?
3. Future idea: It's a trade-off, but you could add a supervisor thread that ensures the async producer worker thread is running. It's an additional thread per async producer, but it would mean you don't need to wait for someone to call `#produce` before restarting any workers that crashed. Thoughts?

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,20 @@ When calling `#shutdown`, the producer will attempt to deliver the messages and 
 
 **Note:** if the calling thread produces messages faster than the producer can write them to Kafka, you'll eventually run into problems. The internal queue used for sending messages from the calling thread to the background worker has a size limit; once this limit is reached, a call to `#produce` will raise `Kafka::BufferOverflow`.
 
+You may also provide a custom error handler to connect with any error handling service you might use.
+
+```ruby
+# Our custom error handler. Any errors that happen in the background thread will end up here.
+error_handler = lambda do |error, _logger, payload|
+  MyErrorHandlingService.handle(error, payload)
+end
+
+# `#async_producer` will create a new asynchronous producer.
+producer = kafka.async_producer(error_handler: error_handler)
+
+# ...
+```
+
 #### Serialization
 
 This library is agnostic to which serialization format you prefer. Both the value and key of a message is treated as a binary string of data. This makes it easier to use whatever serialization format you want, since you don't have to do anything special to make it work with ruby-kafka. Here's an example of encoding data with JSON:
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ You may also provide a custom error handler to connect with any error handling s
 
 ```ruby
 # Our custom error handler. Any errors that happen in the background thread will end up here.
-error_handler = lambda do |error, _logger, payload|
+error_handler = lambda do |error, payload|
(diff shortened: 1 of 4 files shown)


film42:
Looks like that functional test is flaky. I don't have permissions to restart the CI. That last commit only adds rubydoc comments.

dasch:
I think an exception tracker is the wrong place to track these errors – they're not exceptional, and they may happen in very large numbers from time to time. Before proceeding with this, are you using a monitoring system like Grafana, Datadog, or Librato?

film42:
@dasch That may be true for errors during a shutdown, but if there's an error that occurs that threatens to abort the VM, then I think there should be a mechanism to handle it. Since a lot of users have different use-cases, it's easier to let the user provide their own handler vs bake something rigid into the lib.

For our rabbitmq tooling we send metrics to grafana (timers, counters, etc) and exceptions to honeybadger.

dasch:
What I'm saying is that the async producer should auto-heal whenever possible – I think restarting after a sleep would be the best fallback in cases where there's no way to gracefully handle the error. However, I _don't_ think the user should have to configure a callback for this. Errors are _already_ being sent to Statsd, and I _strongly_ suggest you use that to set up monitoring.

film42:
@dasch I see what you mean about the error handling via instrumentation:
```ruby
        rescue Kafka::Error => e
          @instrumenter.instrument("topic_error.producer", {
            topic: message.topic,
            exception: [e.class.to_s, e.message],
          })
```
But this doesn't give access to the backtrace of the error. It's a simple fix though. Would that be an OK change to make?

Additionally, would you prefer it if I used the same instrumenter in the async producer to be able to remove the `abort_on_exception`? The async producer only reports simple count metrics like number of dropped messages, etc. As long as I can subscribe to those messages and report them out, I'm 👍 . Relying on `abort_on_exception` is a kill switch waiting to cause a service outage from my point of view.

dasch:
Let me look into an alternative to `abort_on_exception`.

dasch:
Would you care to take a look at https://github.com/zendesk/ruby-kafka/pull/432?

PR #428, opened 2017-10-11
Consumers should send heartbeats when no messages are available

jturkel:
This PR updates consumers to send heartbeats even if there are no messages available on the server. Previously consumers would get kicked out of the consumer group if no data became available on the server before the session timeout.


jturkel:
Looks like there's already heartbeat in `fetch_batches` that I missed so this change isn't necessary.

PR #432, opened 2017-10-17
Restart the async producer on errors

dasch:
This is an attempt to fix the cause of #422.

When there's an unexpected Kafka error, we should not crash the host application – that goes against what the async producer is there to do, which is insulate the host from Kafka problems.

Instead, we sleep for 10 seconds before restarting the background worker. This should both ensure that we don't burn crazy amounts of CPU if we run into a loop, but also that the user can look at the logs and take action.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -142,16 +142,10 @@ module Kafka
 
     def ensure_threads_running!
       @worker_thread = nil unless @worker_thread && @worker_thread.alive?
-      @worker_thread ||= start_thread { @worker.run }
+      @worker_thread ||= Thread.new { @worker.run }
 
       @timer_thread = nil unless @timer_thread && @timer_thread.alive?
-      @timer_thread ||= start_thread { @timer.run }
-    end
-
-    def start_thread(&block)
-      thread = Thread.new(&block)
-      thread.abort_on_exception = true
-      thread
+      @timer_thread ||= Thread.new { @timer.run }
     end
 
     def buffer_overflow(topic)
@@ -224,6 +218,9 @@ module Kafka
 
         sleep 10
         retry
+      rescue Exception => e
+        @logger.error "Unexpected Kafka error #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}"
+        @logger.error "Async producer crashed!"
       ensure
         @producer.shutdown
       end
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -218,6 +218,12 @@ module Kafka
             raise "Unknown operation #{operation.inspect}"
           end
         end
+      rescue Kafka::Error => e
+        @logger.error "Unexpected Kafka error #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}"
+        @logger.info "Restarting in 10 seconds..."
+
+        sleep 10
+        retry
       ensure
         @producer.shutdown
       end


lib/kafka/async_producer.rb
```diff
@@ -218,6 +218,12 @@ def run
             raise "Unknown operation #{operation.inspect}"
           end
         end
+      rescue Kafka::Error => e
+        @logger.error "Unexpected Kafka error #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}"
+        @logger.info "Restarting in 10 seconds..."
+
+        sleep 10
+        retry
```
film42: Should we have a retry limit here in case the error is related to a problem with the message and not the connection? Example: https://github.com/abrandoned/protobuf-nats/blob/master/lib/protobuf/nats/client.rb#L96-L105

film42: Also, 10 seconds might be a bit much. You could make this 1 second and it would still use a really low amount of system resources.

lib/kafka/async_producer.rb
```diff
@@ -218,6 +218,12 @@ def run
             raise "Unknown operation #{operation.inspect}"
           end
         end
+      rescue Kafka::Error => e
+        @logger.error "Unexpected Kafka error #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}"
```
film42: 👍 to logging this error. Doesn't have to be now, but we should publish these errors out through the instrumenter.

lib/kafka/async_producer.rb
```diff
@@ -218,6 +218,12 @@ def run
             raise "Unknown operation #{operation.inspect}"
           end
         end
+      rescue Kafka::Error => e
```
film42: Just realized that you might want to make this `rescue => e` because it's still possible for some non-kafka error to crash the host while `thread.abort_on_exception = true` is still used.

lib/kafka/async_producer.rb
dasch: Yeah, it would make sense to track this in a dashboard.

lib/kafka/async_producer.rb
dasch: I think it's fairly typical to have a high number of e.g. Unicorn worker processes running on a host, and this would run per-worker – so if there's a systemic error, you'd get a restart per worker based on the frequency. I think 10s should suffice – this would only cover totally unforeseen errors, as we should add specific handling of errors whenever possible.

lib/kafka/async_producer.rb
dasch: That would be a bug that would warrant a hard crash.

lib/kafka/async_producer.rb
film42: Interesting. We run on jruby with puma backed by a single thread pool, so for us one instance has one async producer. Would you be open to making this configurable? I'm happy to make a follow-up change :)

lib/kafka/async_producer.rb
film42: Maybe for your workload but for us it's unacceptable. Would you be open to making this behavior configurable?

dasch: Have you ever seen a non-Kafka error cause a crash? Do all other libraries you use rescue and hide all runtime errors?

mensfeld: I'm with @dasch  on this one. @film42  can't you wrap around your logic with rescue?

dasch: If you have very specific requirements for an async producer, it's also possible to create your own. The one included in ruby-kafka simply wraps the sync producer, so it's actually not that difficult.

film42: @mensfeld I don’t have access to this code path because it’s in a separate thread that will crash the VM if something isn’t rescued. 

@dasch This is the first time I’ve seen a library that uses abort on exception to find bugs. I don’t think you need to handle any possible error ever here, but if you’re saying “any unhabdled error will crash the VM,” then I think that warrants more protection. But you’re right, it’s not hard to make my own. I was hoping we could find a way to make the default one more safe and generic. Sorry, I’m not trying to be annoying or argumentative. Thanks for working through this :)

So in summary I think not aborting on exception but publishing out all errors, and using a back-off timeout would cover both of our cases and expose any library bugs to the user. And I’m not asking you to write this, I’m just wondering if you think it’s a good pattern or not. I’m more than happy to make a cut and open a PR. I know you prefer the current implementation, but is there some kind of scenario where you’d be happy to remove abort on exception?

dasch: I think not aborting on unexpected exceptions is reasonable, as long as the error is being logged. The worker thread would die though, so all writes would be dropped. 

Let me push an update to the PR.

film42: I have some ideas on preserving writes between worker threads but as for now this branch performs the same as master branch (where the vm would stop and writes would be lost) so 👍  .

film42:
Thanks @dasch !

PR #436, opened 2017-10-18
Report consumer lag to Datadog when processing batches

jasondoc3:
If only processing messages in batches, no consumer lag information is reported to Datadog.


dasch:
:+1:

Issue #424, opened 2017-10-06
Insufficient #stop behaviour for non busy topics with manual offset commits


mensfeld:
This problem is related to all the ruby-kafka versions, since they all have the same ```#stop``` logic.

As described in th docs, the signal gets caught and consumer shuts-down:

```ruby
trap("TERM") { consumer.stop }
```

The problem itself is the trap context which is in a separate thread. It modifies the ```@running``` flag on a consumer and during the next ```consumer_loop``` iteration it stops. It creates a problem with manual offset commitment in a particular case described below. 

This is sufficient enough for:

- the automatic offset commits as they happen upon new messages processing end
- manual offset commit on a busy topics in which we can perform a periodical offset commitment (for example every 10 batches, etc)

However one case is not covered: topics that aren't that busy. Let's assume I get 1 message per 10 minutes and I commit every 10 messages OR during the shutdown process. 

After 5 messages (50 minutes) I get a ```TERM``` upon which I want to commit the offset BUT I won't get any messages for next 10 minutes, which means that my business logic won't get evaluated as it gets evaluated upon new messages arrival.

Theoretically I could try passing last message to the trap context thread and use ```mark_message_as_processed``` from it, however it seems it's not the right way (multithreading same consumer would be crazy).

So the general problem is, that for non busy topics a ```yield``` in a ```consumer_loop``` upon messages is insufficient for handling the manual offset commitment during the shutdown process.

Obviously I cannot commit the offset after the ```#stop``` is sent because consumer is already without any partitions assigned. 

How could we solve it? One of the ideas is to provide an additional callback that could be executed **after** the ```consumer_loop``` (if any)  but **before** the consumer leaves the group. Something like that should be sufficient to commit the offset only during the shutdown:

```ruby
buffer = []

consumer.before_stop do
  consumer.mark_message_as_processed(buffer.last) unless buffer.empty?
  consumer.commit_offsets
end

consumer.each_message(automatically_mark_as_processed: false) do |message|
  buffer << message
end
```

Without being able to provide additional logic (in any way) it's impossible to create long lasting transactions that get committed after a period of time **or** during the shutdown.

dasch:
I really want to avoid having a bunch of callbacks. The issue here seems to be that manual offset management isn't really thought out completely. I must admit that I'm sort of skeptical of adding a bunch of complexity to support what seems to me to be an edge case. Why is manual offset management used here?

I would personally _always_ have a time bound on commits...

[...]

PR #438, opened 2017-10-20
Instantly kill socket reads on consumer shutdown

### LINKED ISSUE #437: Cancel socket reads on shutdown
When e.g. a consumer is shut down while in the midst of a socket read, the time to gracefully shut down can be quite long, especially if a fetch's `max_wait_time` is high. This can cause some process managers to run out of patience and kill the process. This is not a problem by itself, but it can show up as noise in monitoring systems and generally be confusing.

Ideally, we'd be able to cancel any active IO work and shut down gracefully fast enough to please those process managers.

dasch:
Fixes #437.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -94,6 +94,7 @@ module Kafka
     # @return [nil]
     def stop
       @running = false
+      @cluster.disconnect
     end
 
     # Pause processing of a specific topic partition.
--- a/lib/kafka/connection.rb
+++ b/lib/kafka/connection.rb
@@ -105,7 +105,7 @@ module Kafka
 
         response
       end
-    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
+    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, Errno::ESHUTDOWN, EOFError => e
       close
 
       raise ConnectionError, "Connection error: #{e}"
--- a/lib/kafka/socket_with_timeout.rb
+++ b/lib/kafka/socket_with_timeout.rb
@@ -24,6 +24,9 @@ module Kafka
 
       @timeout = timeout
 
+      # This pipe is used to cancel IO.select calls when sockets are closed.
+      @cancel_reader, @cancel_writer = IO.pipe
+
       @socket = Socket.new(Socket.const_get(addr[0][0]), Socket::SOCK_STREAM, 0)
       @socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)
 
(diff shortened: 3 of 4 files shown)


lib/kafka/consumer.rb
```diff
@@ -94,6 +94,7 @@ def subscribe(topic, default_offset: nil, start_from_beginning: true, max_bytes_
     # @return [nil]
     def stop
       @running = false
+      @cluster.disconnect
```
mensfeld: You know that this happens in a different thread when catching a signal right?

dasch: Are you running the consumer outside the main thread? I believe signal handlers run in the main thread, but I may be wrong.

mensfeld: @dasch yes ruby executes the signal handler in the same thread as the parent but I don't execute consumer that way as I have multi threaded environment and this change makes it unsafe to stop the execution (it was ok without that because of a fetch loop)

dasch: The problem is that during the fetch loop, the thread will block even after `stop` is called – that's why a more aggressive shutdown is required.

The consumer isn't designed to be thread safe – in this case, can you push a message into the thread running the consumer and have that thread call `stop` itself rather than calling it from another thread? That would of course mean you couldn't use `each_message` or `each_batch`.

mensfeld: > That would of course mean you couldn't use each_message or each_batch.
So I would be left with the non-merged manual api for consumption, right?

dasch: I might have to push this until after the new consumer design has been implemented – when the fetches are being done in separate threads it will be easier to manage this.

mensfeld: You mean fetches for separate topics out out the same consumer group, right?

dasch: Each consumer would have a separate fetcher thread _per broker connection_, so we can parallelize the fetches. The fetches would still have to be interruptible though, unless I just decide to kill the threads on shutdown...

mensfeld: @dasch yeah I will just drop celluloid in favor of threads and signaling with ```handle_interrupt``` and will just execute shutdown out of each thread natively. But the #424 still remains in that case.

dasch: I still think that #369 is the right level of abstraction for that use case. Can you try to test out the API?

mensfeld: @dasch yes - we agreed that I will test it BUT it still adds a lot of boilerplate to cover just a single side case. Nonetheless you can count on me in that matter (just needs a bit more time to adapt to it)

dasch: 👍 

I think the `each_*` should be the "high-level" consumer APIs in ruby-kafka, with `poll` being used by other libraries/frameworks and people with special needs (a small minority).

dasch: @mensfeld thinking about this, I think it shouldn't be a problem that `close` is called from a different thread – the write to `@cancel_writer` is thread safe, as it's just a pipe.

dasch: Can you verify that it's a problem in an application of yours?

mensfeld: @dasch it's no longer a problem as I redesigned my consumers to be 100% thread safe ;) the trap context changes the internal state and this propages the stop:

```ruby
def consumer_loop
  super do
    if Karafka::App.stopped?
      stop
    else
      yield
    end
  end
end
```

dasch: :+1:

PR #443, opened 2017-10-23
Commit offsets even if we don't read any new messages

jturkel:
This is similar to a83fddfcf84e01d2f242452caa8b8a7507493916 which was applied to Consumer#each_message and ensures that offsets are retained even if we don't receive any new messages for longer than the offset retention period. It would be great if this could be backported to the v0.4-stable branch too.


dasch:
Backparted to v0.4 as well.

jturkel:
Thanks @dasch!

Issue #441, opened 2017-10-22
automatically_mark_as_processed false causes infinite loop


mensfeld:
* Version of Ruby: ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]
* Version of Kafka: 0.10.2.0
* Version of ruby-kafka: '0.4.3'

Fixed on master? **no**

## Required Kafka cluster setup

- Single Kafka node
- Single topic named 'a') with single partition (our favorite setup :D)

## Steps to reproduce

```ruby
require 'ruby-kafka'
kafka = Kafka.new(seed_brokers: ['kafka://127.0.0.1:9092'])
# Rand is on purpose just to prove, that the same message is being fetched
kafka.deliver_message(rand.to_s, topic: 'a')
consumer = kafka.consumer(group_id: 'consumer-name')
consumer.subscribe('a')
consumer.each_message(automatically_mark_as_processed: false) do |message|
  p message
end
```

###### Expected outcome

Ruby-Kafka should consume messages one after another, not a single one in a loop.

```
#<Kafka::FetchedMessage:0x000000000212918 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
```

The problem occurs for both ```each_message``` and ```each_batch```

###### Actual outcome

Neverending stream of the same message fetched multiple times. In case more messages are present, it will fetch them the same way (in a loop and again from the beginning):

```
#<Kafka::FetchedMessage:0x000000000220b6b8 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
#<Kafka::FetchedMessage:0x00000000021fa778 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
#<Kafka::FetchedMessage:0x00000000021ec498 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
#<Kafka::FetchedMessage:0x00000000021dddf8 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
#<Kafka::FetchedMessage:0x00000000021d9c08 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
#<Kafka::FetchedMessage:0x00000000021d5540 @value="0.8694582576553449", @key=nil, @topic="a", @partition=0, @offset=0>
[...]

PR #447, opened 2017-10-24
automatically_mark_as_processed false causes infinite loop #441

mensfeld:
This PR aims to solve the problem of lack of offset management for topics without automatic marking. It introduces an in memory offset cache per each partition and topic. Also it supports the failure and will reconsume the whole match for ```each_batch``` and the last message for ```each_message```.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -57,6 +57,17 @@ module Kafka
 
       # The maximum number of bytes to fetch from a single partition, by topic.
       @max_bytes = {}
+
+      # Hash containing offsets for each topic and partition that has the
+      # automatically_mark_as_processed feature disabled. Offset manager is only active
+      # when everything is suppose to happen automatically. Otherwise we need to keep track of the
+      # offset manually in memory for all the time
+      # The key structure for this equals an array with topic and partition [topic, partition]
+      # The value is equal to the offset of the last message we've received
+      # @note It won't be updated in case user marks message as processed, because for the case
+      #   when user commits message other than last in a batch, this would make ruby-kafka refetch
+      #   some already consumed messages
+      @current_offsets = {}
     end
 
     # Subscribes the consumer to a topic.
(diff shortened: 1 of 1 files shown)


dasch:
Thanks! I'll look at this tomorrow. 👍

lib/kafka/consumer.rb
```diff
@@ -57,6 +57,17 @@ def initialize(cluster:, logger:, instrumenter:, group:, offset_manager:, sessio
...
+      # Hash containing offsets for each topic and partition that has the
+      # automatically_mark_as_processed feature disabled. Offset manager is only active
+      # when everything is suppose to happen automatically. Otherwise we need to keep track of the
+      # offset manually in memory for all the time
+      # The key structure for this equals an array with topic and partition [topic, partition]
+      # The value is equal to the offset of the last message we've received
+      # @note It won't be updated in case user marks message as processed, because for the case
+      #   when user commits message other than last in a batch, this would make ruby-kafka refetch
+      #   some already consumed messages
+      @current_offsets = {}
```
dasch: Can this follow the convention of nested hashes used elsewhere in the code?

mensfeld: Yes. Could you provide me with an example just to be sure.

dasch:
```ruby
@current_offsets = Hash.new {|h, k| h[k] = {} }
```

mensfeld: @dasch but this type of nested hash presents a different behavior than that I'm aiming and disallows for nil check (as the hash is never nil). I can do that but it will require really inelegant changes for comparing the non-existing content with an empty hash (or something like that)

dasch: The nil check is done per partition, right?

```ruby
@current_offsets[topic].key?(partition)
```

mensfeld: Ah you mean that type of structure. Now I get it. Will update soon

dasch:
Thanks!

PR #453, opened 2017-10-26
Support the topic creation API

dasch:
Requires Kafka 0.10.2+

### DIFF
--- a/lib/kafka/broker.rb
+++ b/lib/kafka/broker.rb
@@ -105,5 +105,11 @@ module Kafka
 
       @connection.send_request(request)
     end
+
+    def create_topics(**options)
+      request = Protocol::CreateTopicsRequest.new(**options)
+
+      @connection.send_request(request)
+    end
   end
 end
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -407,12 +407,22 @@ module Kafka
       end
     end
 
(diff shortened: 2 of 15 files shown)


mensfeld:
Don't forget about updating README :dancing_men:  :D

dasch:
I need to update the test infrastructure to Kafka 0.11 first...

tcrayford:
@dasch technically 0.10.2 or so iirc supports this API as well, but there are a lot more admin apis in 0.11 that are very useful.

dasch:
@tcrayford good to know! I'm tracking the 0.11 API support [in this project](https://github.com/zendesk/ruby-kafka/projects/3).

dasch:
@tcrayford it seems that [Kafka 0.10.1.0](https://archive.apache.org/dist/kafka/0.10.1.0/RELEASE_NOTES.html) was the first release with support for this API. I'm thinking about making that the minimum required version for ruby-kafka in order to not have to provide support for multiple versions of the metadata API just yet. Do you have any sense of how many people are still using previous versions of 0.10?

tcrayford:
@dasch only on heroku really, and we've been pushing folk off older 0.10 releases because of other bugs.

dasch:
@tcrayford I'm going to see how difficult it would be to support multiple versions...

PR #454, opened 2017-10-26
Rescue all system call errors

dasch:


lib/kafka/connection.rb
```diff
@@ -109,10 +109,10 @@ def send_request(request)
 
         response
       end
-    rescue Errno::EPIPE, Errno::ECONNRESET, Errno::ETIMEDOUT, EOFError => e
+    rescue SystemCallError, EOFError => e
```
mensfeld: won't it catch also user errors like ```File.open("does/not/exist")```

dasch: Hopefully we won't do that in that specific method call...

PR #465, opened 2017-11-02
SASL SCRAM support

spuun:
Added a SASL SCRAM authenticator.

Instead of adding even more parameters to Client constructor it's now possible to inject an authenticator. This makes it possible to create custom authenticators without modifing ruby-kafka. To make this possible the mechanism validation in SaslHandshakeRequest had to be removed. Possible authentication mechanism is validated by the broker so this shouldn't be needed anyway.

It's also possible to inject a SSL context to the Client constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -935,7 +935,7 @@ Since 0.11 kafka supports [SCRAM](https://kafka.apache.org/documentation.html#se
 kafka = Kafka.new(
   sasl_scram_username: 'username',
   sasl_scram_password: 'password',
-  sasl_scram_mechanism: 'SHA-256',
+  sasl_scram_mechanism: 'sha256',
   # ...
 )
 ```
--- a/README.md
+++ b/README.md
@@ -912,18 +912,6 @@ kafka = Kafka.new(
   sasl_gssapi_keytab: '/etc/keytabs/kafka.keytab',
   # ...
 )
-
-# or
-
-require 'kafka/sasl_gssapi_authenticator'
-kafka = Kafka.new(
-  authenticator: Kafka::SaslGssapiAuthenticator(
-    sasl_gssapi_principal: 'kafka/kafka.example.com@EXAMPLE.COM',
-    sasl_gssapi_keytab: '/etc/keytabs/kafka.keytab',
-    logger: someLogger # optional
-  ),
-  # ...
-)
 ```
 
 ##### PLAIN
@@ -936,50 +924,18 @@ kafka = Kafka.new(
   sasl_plain_password: 'password'
   # ...
 )
-
-# or
-
-require 'kafka/sasl_plain_authenticator'
-kafka = Kafka.new(
-  ssl_ca_cert: File.read('/etc/openssl/cert.pem'), # Optional but highly recommended
-  authenticator: Kafka::SaslPlainAuthenticator(
-    username: 'username',
-    password: 'password',
-    logger: someLogger, # optional
-    authzid: 'auth_identity' # optional. if nil, username will be used by kafka
-  ),
-  # ...
-)
 ```
 
 **NOTE**: It is __highly__ recommended that you use SSL for encryption when using SASL_PLAIN
 
 ##### SCRAM
-
 Since 0.11 kafka supports [SCRAM](https://kafka.apache.org/documentation.html#security_sasl_scram). 
 
 ```ruby
-require 'kafka/sasl_scram_authenticator'
 kafka = Kafka.new(
-  authenticator: Kafka::SaslScramAuthenticator.new(
-    'username',
-    'password',
-    logger: someLogger, # optional
-    mechanism: Kafka::SCRAM_SHA256 # optional Supported values are Kafka::SCRAM_SHA256 (default) and Kafka::SCRAM_SHA512
-  ),
-  # ...
-)
-```
-
-To use SCRAM over SSL and the server cert is signed by a CA in your default cert store, add a SSL context to enable SSL:
-```ruby
-require 'openssl'
-require 'kafka/sasl_scram_authenticator'
-kafka = Kafka.new(
-  ssl_context: OpenSSL::SSL::SSLContext.new,
-  authenticator: Kafka::SaslScramAuthenticator.new(
-    # ...
-  ),
+  sasl_scram_username: 'username',
+  sasl_scram_password: 'password,
+  sasl_scram_mechanism: Kafka::SCRAM_SHA256,
   # ...
 )
 ```
--- a/README.md
+++ b/README.md
@@ -901,8 +901,9 @@ Typically, Kafka certificates come in the JKS format, which isn't supported by r
 
 #### Authentication using SASL
 
-Kafka has support for using SASL to authenticate clients. Currently GSSAPI and PLAIN mechanisms are supported by ruby-kafka.
+Kafka has support for using SASL to authenticate clients. Currently GSSAPI, SCRAM and PLAIN mechanisms are supported by ruby-kafka.
 
+##### GSSAPI
 In order to authenticate using GSSAPI, set your principal and optionally your keytab when initializing the Kafka client:
 
 ```ruby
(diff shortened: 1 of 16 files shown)


dasch:
I'd rather keep adding parameters to the initializer – this was a deliberate design choice that may make the code a bit harder to follow, but would be simpler for most users of the library.

lib/kafka/sasl_scram_authenticator.rb
```diff
@@ -0,0 +1,174 @@
...
+
+        response = parse_response(@decoder.bytes)
+        log_debug "[scram] Received server's final msg: #{response}"
+        log_debug "[scram] Client calculated server signature: #{server_signature}"
+
+        raise FailedScramAuthentication, response['e'] if response['e']
+        raise FailedScramAuthentication, 'Invalid server signature' if response['v'] != server_signature
+      rescue FailedScramAuthentication
+        raise
+      rescue => e
```
dasch: I'd like to avoid catch-all rescues.

spuun:
I've changed to code to use new parameters in the initializer. I also put back the validation in SaslHandshakeRequest.

README.md
```diff
@@ -924,6 +926,18 @@ kafka = Kafka.new(
 
 **NOTE**: It is __highly__ recommended that you use SSL for encryption when using SASL_PLAIN
 
+##### SCRAM
+Since 0.11 kafka supports [SCRAM](https://kafka.apache.org/documentation.html#security_sasl_scram). 
+
+```ruby
+kafka = Kafka.new(
+  sasl_scram_username: 'username',
+  sasl_scram_password: 'password,
```
dasch: Missing an end quite there.

spuun: Oh! Forgot to validate docs..

README.md
```diff
@@ -924,6 +926,18 @@ kafka = Kafka.new(
...
+  sasl_scram_mechanism: Kafka::SCRAM_SHA256,
```
dasch: I'd prefer a String name of the mechanism, e.g. `"sha256"`, as that will make it a lot easier to integrate with config files and the like.

lib/kafka.rb
```diff
@@ -225,6 +225,15 @@ class OffsetCommitError < Error
   class FetchError < Error
   end
 
+  class NoPartitionsAssignedError < Error
+  end
```
dasch: How did this sneak in here?

lib/kafka/sasl_gssapi_authenticator.rb
```diff
@@ -3,7 +3,7 @@ class SaslGssapiAuthenticator
     GSSAPI_IDENT = "GSSAPI"
     GSSAPI_CONFIDENTIALITY = false
 
-    def initialize(connection:, logger:, sasl_gssapi_principal:, sasl_gssapi_keytab:)
+    def initialize(conncetion:, logger:, sasl_gssapi_principal:, sasl_gssapi_keytab:)
```
dasch: Typo here.

lib/kafka/sasl_scram_authenticator.rb
```diff
@@ -0,0 +1,171 @@
...
+      end
+
+      log_debug "authenticating #{@username} with scram, mechanism: #{@mechanism}"
+
+      @encoder = @connection.encoder
+      @decoder = @connection.decoder
+
+      begin
+        msg = first_message
+        log_debug "[scram] Sending client's first message: #{msg}"
```
dasch: There's currently no precedent for "tagging" log lines with e.g. `[scram]`. Could you instead do `Sending first SCRAM client message`? The same for the rest of the log lines.

lib/kafka.rb
spuun: I have no idea... Maybe some merge i missed.

lib/kafka/sasl_scram_authenticator.rb
spuun: Yes

README.md
spuun: Ok. I can change that. Do you prefer `"sha256"` over `"SHA-256"` (the current value of the constant)?

dasch: I prefer `"sha256"`.

dasch:
Looking good! Let me know when you need another round of review.

spuun:
I've changed the mechanism configuration parameters to be 'sha256' and 'sha512'.

lib/kafka/client.rb
```diff
@@ -49,23 +49,36 @@ class Client
     #
     # @param sasl_gssapi_keytab [String, nil] a KRB5 keytab filepath
     #
+    # @param sasl_scram_username [String, nil] SCRAM username
+    #
+    # @param sasl_scram_password [String, nil] SCRAM password
+    #
+    # @param sasl_scram_mechanism [String, nil] Scram mechanism ("sha256", "sha512")
+    #
+    # @param use_ssl [Booleanm false] Use SSL
```
dasch: Why add this parameter? The decision on whether to use SSL is already done here:

https://github.com/zendesk/ruby-kafka/blob/a0e1d14c9a5c3f2c653f73bada5c08d9d957cc5f/lib/kafka/client.rb#L528-L529

lib/kafka/client.rb
```diff
@@ -49,23 +49,36 @@ class Client
...
+                   use_ssl: false, sasl_scram_username: nil, sasl_scram_password: nil, sasl_scram_mechanism: nil)
       @logger = logger || Logger.new(nil)
       @instrumenter = Instrumenter.new(client_id: client_id)
       @seed_brokers = normalize_seed_brokers(seed_brokers)
 
       ssl_context = build_ssl_context(ssl_ca_cert_file_path, ssl_ca_cert, ssl_client_cert, ssl_client_cert_key)
+      ssl_context = OpenSSL::SSL::SSLContext.new if use_ssl and !ssl_context
 
-      sasl_authenticator = SaslAuthenticator.new(
+      sasl_authenticator ||= SaslAuthenticator.new(
```
dasch: The `||` can be removed.

spec/fake_server.rb
```diff
@@ -29,49 +29,135 @@ def start
...
+    decoder = Kafka::Protocol::Decoder.new(client)
 
-      request_decoder = Kafka::Protocol::Decoder.new(StringIO.new(request_bytes))
-      api_key = request_decoder.int16
-      api_version = request_decoder.int16
-      correlation_id = request_decoder.int32
-      client_id = request_decoder.string
+    loop do
+      request_bytes = decoder.bytes
+      request_data = Kafka::Protocol::Decoder.new(StringIO.new(request_bytes));
```
dasch: No reason for a semicolon here.

spec/fake_server.rb
```diff
@@ -29,49 +29,135 @@ def start
...
+      api_key = request_data.int16
+      _api_version = request_data.int16
+      correlation_id = request_data.int32
+      _client_id = request_data.string
```
dasch: No reason for underscores here.

spec/fake_server.rb
```diff
@@ -29,49 +29,135 @@ def start
...
       response_encoder = Kafka::Protocol::Encoder.new(response)
       response_encoder.write_int32(correlation_id)
 
-      if api_key == 17 # SASL Authentication
-        response_encoder.write_int16(0) # no errors
-        response_encoder.write_array(SUPPORTED_MECHANISMS) { |msg| response_encoder.write_string(msg) }
-        @authenticating = true
-        @auth_mechanism = message
+      case api_key
+      when 17 then
```
dasch: No reason to switch to a `case` statement here.

dasch:
Thanks for the hard work! ❤️

dasch:
@spuun I've released v0.5.1.beta1 with this functionality – would you be able to deploy and test that version?

spuun:
I've tested beta2 and it seems to work just fine! 👍

dasch:
@spuun great, thanks!

PR #469, opened 2017-11-03
Improve error message for UnknownTopicOrPartition

dasch:
It's not really useful unless you know which topic is missing.


dasch:
@tcrayford yeah, if there's a use case for it :+1:

PR #472, opened 2017-11-06
Handle messages that exceed the maximum read size

### LINKED ISSUE #429: Consumer gets stuck in loop fetching partial message
* Version of Ruby: 2.4.2
* Version of Kafka: 0.9.0.1
* Version of ruby-kafka: 0.4.2

Recently, a consumer got silently stuck trying to pull a message that was larger than the max_bytes_per_partition we had configured for it. It would have been nice if our consumer had raised an error when this scenario occurred. This looks like something the java consumer does in kafka 0.9 based on https://issues.apache.org/jira/browse/KAFKA-3442.

dasch:
When the client specifies a maximum number of bytes to read from a partition and the first message in the partition exceeds that, the broker will truncate the message set to match the client's maximum. This means that the client is unable to decode any messages.

Rather than infinitely keep asking for messages just to have a truncated message returned, we not raise MessageTooLargeToRead.

Fixes #429.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -21,6 +21,10 @@ module Kafka
   class NoPartitionsToFetchFrom < Error
   end
 
+  # A message in a partition is larger than the maximum we've asked for.
+  class MessageTooLargeToRead < Error
+  end
+
   # Subclasses of this exception class map to an error code described in the
   # Kafka protocol specification.
   #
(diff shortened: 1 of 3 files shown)


tcrayford:
@dasch I’m pretty sure the upstream Java clients ask for a higher max bytes in this case. Should we consider that?

dasch:
Yeah – the UX aspect should be handled farther up the system.

dasch:
I'm not quite sure where to place it, though – currently, this would crash the entire consumer, which maybe isn't the worst default behavior. Having it fail partially would make detecting the issue harder. It would probably also require making the message set decoding lazy, since we currently get the exception while decoding the fetch response, but that's not a great place to enforce a policy.

PR #476, opened 2017-11-08
Refactor Broker and Connection

dasch:


lib/kafka.rb
```diff
@@ -25,6 +25,10 @@ class NoPartitionsToFetchFrom < Error
   class MessageTooLargeToRead < Error
   end
 
+  # A connection has been unused for too long, we assume the server has killed it.
+  class IdleConnection < Error
```
mensfeld: Isn't the IdleConnection = Class.new(Error) syntax nicer?

dasch: Nah.

PR #482, opened 2017-11-10
Ignore invalid message timestamps

dasch:
Apparently Kafka will set the timestamp to zero when migrating pre-0.10 messages.

### DIFF
--- a/lib/kafka/protocol/message.rb
+++ b/lib/kafka/protocol/message.rb
@@ -89,6 +89,11 @@ module Kafka
           timestamp = nil
         when 1
           timestamp = message_decoder.int64
+
+          # If the timestamp is set to zero, it's because the message has been upgraded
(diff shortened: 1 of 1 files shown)


lib/kafka/protocol/message.rb
```diff
@@ -89,6 +89,10 @@ def self.decode(decoder)
           timestamp = nil
         when 1
           timestamp = message_decoder.int64
+
+          # If the timestamp is set to zero (start of UNIX epoch) there's clearly
```
tcrayford: @dasch FYI that this can happen if you do an in place upgrade from 0.9.X to 0.10.X or onwards - old messages get zerod out (iirc)

dasch: Ah! I'll fix the comment.

Issue #444, opened 2017-10-23
Retry upon processing failures processes message twice


mensfeld:
* Version of Ruby: ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]
* Version of Kafka: 0.10.2.0
* Version of ruby-kafka: master

Fixed on master? **no**

## Required Kafka cluster setup

- Single Kafka node
- Single topic named 'a') with single partition (our favorite setup :D)

## Steps to reproduce

```ruby
require 'ruby-kafka'
kafka = Kafka.new(seed_brokers: ['kafka://127.0.0.1:9092'])
# Rand is on purpose just to prove, that the same message is being fetched
kafka.deliver_message(rand.to_s, topic: 'a')
consumer = kafka.consumer(group_id: 'consumer-name')
consumer.subscribe('a')

begin
consumer.each_message(automatically_mark_as_processed: true) do |message|
  p message
  raise StandardError
end
rescue
end

consumer.each_message(automatically_mark_as_processed: true) do |message|
  p message
end

```

###### Expected outcome

Ruby-Kafka should reconsume message once, not twice.

```
#<Kafka::FetchedMessage:0x0000000230cc60 @value="0.8970946704093968", @key=nil, @topic="a", @partition=0, @offset=0, @create_time=2017-10-23 17:27:42 +0200>
#<Kafka::FetchedMessage:0x00000002307558 @value="0.8970946704093968", @key=nil, @topic="a", @partition=0, @offset=0, @create_time=2017-10-23 17:27:42 +0200>
```

The problem occurs for both ```each_message``` and ```each_batch```

###### Actual outcome

Re-consumption happens twice instead of once

```
#<Kafka::FetchedMessage:0x0000000230cc60 @value="0.8970946704093968", @key=nil, @topic="a", @partition=0, @offset=0, @create_time=2017-10-23 17:27:42 +0200>
#<Kafka::FetchedMessage:0x00000002307558 @value="0.8970946704093968", @key=nil, @topic="a", @partition=0, @offset=0, @create_time=2017-10-23 17:27:42 +0200>
#<Kafka::FetchedMessage:0x000000022dea40 @value="0.8970946704093968", @key=nil, @topic="a", @partition=0, @offset=0, @create_time=2017-10-23 17:27:42 +0200>
```

dasch:
@mensfeld is this still the case on `master`?

mensfeld:
@dasch yes:

```bash
[...]

PR #486, opened 2017-11-15
Change highwater mark semantics

dasch:
It seems that Kafka has changed the meaning of the highwater mark. Before, it indicated the *next* offset that would be written to, but now it seems to indicate the *last* offset that has already been written.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -208,7 +208,7 @@ module Kafka
                 topic: message.topic,
                 partition: message.partition,
                 offset: message.offset,
-                offset_lag: batch.highwater_mark_offset - message.offset - 1,
+                offset_lag: batch.highwater_mark_offset - message.offset,
                 create_time: message.create_time,
                 key: message.key,
                 value: message.value,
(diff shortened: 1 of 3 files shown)


dasch:
@tcrayford is this a correct understanding? I started seeing an offset lag of -1 in my dashboards...

mensfeld:
@dasch maybe this is what causes #444

dasch:
@mensfeld could be – can you see if this change fixes that bug?

mensfeld:
@dasch  unfortunately problem still persists for dasch/new-highwater-mark-semantics:

```
[ruby-kafka (dasch/new-highwater-mark-semantics)]$ be ruby test.rb 
#<Kafka::FetchedMessage:0x00000001a63f78 @value="0.8741370783650839", @key=nil, @topic="a", @partition=0, @offset=2, @create_time=2017-11-16 09:18:41 +0100>
#<Kafka::FetchedMessage:0x00000001a5a590 @value="0.8741370783650839", @key=nil, @topic="a", @partition=0, @offset=2, @create_time=2017-11-16 09:18:41 +0100>
#<Kafka::FetchedMessage:0x00000001a305d8 @value="0.8741370783650839", @key=nil, @topic="a", @partition=0, @offset=2, @create_time=2017-11-16 09:18:41 +0100>
```

nguyenquangminh0711:
The code changes alot. And it looks like this PR becomes outdated.

PR #489, opened 2017-11-17
Align buffer overflow erros with sync producer

michaelsauter:
Now both producer emit buffer overflow with a message for consistency.

I investigated some buffer overflow issues on production, and first thought that the issue was to do with the `max_buffer_size` and `max_buffer_bytesize` settings. But the error did not include information which limit we had reached. I planned on adding a message to the exception which limit was reached, then realized that the synchronous producer already did this. Only then did I discover the `max_queue_size` setting and noticed that `BufferOverflow` was raised without a message if that limit was reached. I found the corresponding error message in the logs, so now I know that this was the problem, but I think it would be good to have both error cases handled in the same way.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -100,7 +100,10 @@ module Kafka
     def produce(value, topic:, **options)
       ensure_threads_running!
 
-      buffer_overflow(topic) if @queue.size >= @max_queue_size
+      if @queue.size >= @max_queue_size
+        buffer_overflow topic,
+          "Cannot produce to #{topic}, max queue size (#{@max_queue_size} messages) reached"
+      end
 
       args = [value, **options.merge(topic: topic)]
       @queue << [:produce, args]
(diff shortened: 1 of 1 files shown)


lib/kafka/async_producer.rb
```diff
@@ -148,14 +151,12 @@ def ensure_threads_running!
       @timer_thread ||= Thread.new { @timer.run }
     end
 
-    def buffer_overflow(topic)
+    def buffer_overflow(topic, message)
       @instrumenter.instrument("buffer_overflow.async_producer", {
         topic: topic,
       })
 
-      @logger.error "Cannot produce message to #{topic}, max queue size (#{@max_queue_size}) reached"
```
dasch: I'd like to keep the log statement.

michaelsauter: I removed it because the sync producer doesn't have it either, and the exception already has the message. If you want exception AND logging, and I suggest we add the logging to the sync producer as well for consistency. Should I do that?

dasch:
Good point. It’s okay to leave out the log message.

dasch:
Thanks!

PR #490, opened 2017-11-17
Disconnect cluster in a safer place

### LINKED ISSUE #484: Phobos issues in 0.5.x of Ruby Kafka
We have an issue with ruby kafka >= 0.5.0 in Phobos that we might need your help to investigate, see https://github.com/klarna/phobos/issues/48

We realize that this is our problem, and we have the workaround  to lock < 0.5.0 of Ruby Kafka, but we think you might have seen something similar and can help put us on the right track. I guess the outcome can be either of two things
- There is a bug in ruby kafka that our tests helped expose
- There is a migration path from 0.4.x to 0.5x that we haven't observed

So... Anyway, here is the report and steps to reproduce.

* Version of Ruby: 2.4.1
* Version of Kafka: 0.10.2.0
* Version of ruby-kafka: 0.5.0+

###### Steps to reproduce

1. Clone https://github.com/klarna/phobos
1. In the gemspec, remove the locking of ruby-kafka to allow 0.5.1 to be installed
1. Run `docker-compose up -d --force-recreate kafka zookeeper` to start dependencies
1. Run `docker-compose build test` to build test container
1. Run `docker-compose run --rm test bundle exec rspec`

Many listeners spec will fail with 

```shell
     IOError:
       closed stream
```

###### Expected outcome

Tests should be green, right now we have to lock < 0.5.0 of Ruby Kafka.

###### Actual outcome

Tests are red (listener tests only.

klippx:
Do not disconnect cluster before consumer loop has had a chance to finish its work.

Closes #484

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -105,7 +105,6 @@ module Kafka
     # @return [nil]
     def stop
       @running = false
-      @cluster.disconnect
     end
 
     # Pause processing of a specific topic partition.
@@ -231,7 +230,10 @@ module Kafka
 
             @heartbeat.send_if_necessary
 
-            return if !@running
+            if !@running
+              @cluster.disconnect
+              return
+            end
           end
         end
 
@@ -308,7 +310,10 @@ module Kafka
 
           @heartbeat.send_if_necessary
 
-          return if !@running
+          if !@running
+            @cluster.disconnect
+            return
+          end
         end
 
         # We may not have received any messages, but it's still a good idea to


mensfeld:
What about this: https://github.com/zendesk/ruby-kafka/issues/437 and this https://github.com/zendesk/ruby-kafka/pull/438 ?

klippx:
Ok, I have to look into those... 🤔

mensfeld:
@klippx yeah you need to consult @dasch  on that one as I also wanted to change it but he had some good points on why it should stay that way.

dasch:
We need to be able to shut down immediately, so this won't work.

PR #495, opened 2017-11-27
Fix offsets for messages within MessageSet

ale7714:
* MessageSet holds the offset of the last message which means
  we need to take into account the size of the MessageSet to
  set the Message offset properly

### DIFF
--- a/lib/kafka/protocol/message.rb
+++ b/lib/kafka/protocol/message.rb
@@ -57,11 +57,12 @@ module Kafka
         data = codec.decompress(value)
         message_set_decoder = Decoder.from_string(data)
         message_set = MessageSet.decode(message_set_decoder)
+        base_offset = offset - message_set.size + 1
 
         # The contained messages need to have their offset corrected.
         messages = message_set.messages.each_with_index.map do |message, i|
           Message.new(
-            offset: offset + i,
+            offset: base_offset + i,
             value: message.value,
             key: message.key,
             create_time: message.create_time,
--- a/spec/compressor_spec.rb
+++ b/spec/compressor_spec.rb
@@ -22,7 +22,7 @@ describe Kafka::Compressor do
       # When decoding a compressed message, the offsets are calculated relative to that
       # of the container message. The broker will set the offset in normal operation,
       # but at the client-side we set it to -1.
-      expect(messages.map(&:offset)).to eq [-1, 0]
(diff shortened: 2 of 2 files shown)


lib/kafka/protocol/message.rb
```diff
@@ -57,11 +57,12 @@ def decompress
         data = codec.decompress(value)
         message_set_decoder = Decoder.from_string(data)
         message_set = MessageSet.decode(message_set_decoder)
+        base_offset = offset - message_set.size + 1
```
ale7714: for example offset is 42 but size is 3, we want to start at 40

dasch: This is just weird – what about the message that precedes the container message? Wouldn't that have offset 41?

lib/kafka/protocol/message.rb
```diff
@@ -57,11 +57,12 @@ def decompress
...
 
         # The contained messages need to have their offset corrected.
         messages = message_set.messages.each_with_index.map do |message, i|
```
zmstone: `each_with_index` and `+i` will not work for compacted topics. should use the real relative offsets instead.

Ref: 
https://cwiki.apache.org/confluence/display/KAFK/KIP-31+-+Move+to+relative+offsets+in+compressed+message+sets

> When the log cleaner compacts log segments, if multiple message sets are compacted into one message set, the broker needs to update the inner message's relative offset values. (This will leave "holes" inside the new wrapped message).

lib/kafka/protocol/message.rb
zmstone: offset for the message precedes the container message is `base_offset - 1` (if not compacted)

zmstone: `base_offset` should not be `offset - message_set.size + 1`
should be `offset - last(message_set).offset`
because relative offsets may not be sequential for compacted topics.

dasch:
@zmstone thanks for the review!

dasch:
@ale7714 do you have time to continue work on this?

dasch:
Otherwise I can take a stab at it.

ale7714:
@zmstone thank you for your comments! 
@dasch sorry I’m traveling for the holidays and don’t think I wil be able to work on this soon.

klippx:
https://github.com/zendesk/ruby-kafka/issues/505 @zmstone helped us debug this, this is related :)

PR #496, opened 2017-11-29
Instrument the start of message/batch processing

### LINKED ISSUE #494: Instrument start of message processing
We should instrument the start of `process_message` and `process_batch` so that tracers can know about it.

/cc @palazzem

dasch:
This makes it possible for subscribers to do things before the processing completes.

Closes #494.


jturkel:
Should the [instrumentation section](https://github.com/zendesk/ruby-kafka#consumer-notifications) of the README be updated for this change?

PR #497, opened 2017-12-04
Use JoinGroupResponse metadata for assigning topics as leader

jcalvert:
I observed in a production environment that when you have two processes in the same consumer group but each is subscribed to different topics that the leader will 

- Only assign members to topics that it is subscribed to itself.
- Other members that are not subscribed to the topics they are assigned do not seem to process from those partitions.

Sidestepping this issue is possible by using distinct consumer groups and keeping topic subscriptions consistent within the consumer group, however this is not the behavior observed when using the Java client. From an operational point of view it seems like unexpected behavior. 

Per the [Apache Kafka Protocol Guide](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol) it certainly appears the intention of the metadata is to include the consumer's subscribed topics (which is currently boilerplated with `test-messages`) - so we can take the result of the join group response returned to our leader node and construct a mapping of members to topics and use that to inform our assignment. This pull request should achieve that by:

- Taking the unused `topics` argument and replacing the placeholder in `JoinGroupRequest`
- Adding decoding to the `ConsumerGroupProtocol`
- Deserializing the `ConsumerGroupProtocol` in the `JoinGroupResponse`
- Have the consumer group pass the topics to the initializer of `JoinGroupRequest`
- Updating the assignment strategy to take member objects instead of just member IDs, and interrogating them to get the topic listing in order to produce an accurate consumer topic map.

### DIFF
--- a/spec/functional/consumer_group_spec.rb
+++ b/spec/functional/consumer_group_spec.rb
@@ -147,4 +147,40 @@ describe "Consumer API", functional: true do
 
     expect(processed_messages).to eq(num_messages * 2)
   end
+
+  example "using different topics in a consumer group" do
+    topic = create_random_topic(num_partitions: 2)
+    other_topic = create_random_topic(num_partitions: 2)
+    num_messages = 10
+    processed_messages = 0
+
+    producer = kafka.producer
+    (1..num_messages).each {|i| producer.produce("foo", topic: topic) }
+    (1..num_messages).each {|i| producer.produce("bar", topic: other_topic) }
+    producer.deliver_messages
+
+    group_id = "test#{rand(1000)}"
+    threads = [topic, other_topic].map do |topic_to_use|
+      t = Thread.new do
+        kafka2 = Kafka.new(seed_brokers: kafka_brokers, client_id: "test")
+        consumer = kafka2.consumer(group_id: group_id, offset_retention_time: offset_retention_time)
+        total = []
+        consumer.subscribe(topic_to_use, start_from_beginning: true)
+        begin
+          Timeout::timeout(10) do
+            consumer.each_message do |message|
+              total << message.value
+            end
+          end
+        rescue => e
+          consumer.stop
+        end
+        total
+      end
+      t
+    end
+    threads.each { |t| t.join }
+    expect(threads.first.value).to eq(10.times.map { "foo" })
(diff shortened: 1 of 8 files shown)


jcalvert:
I'm unsure why my functional test has failed in your CI when it passed for me locally.

spec/functional/consumer_group_spec.rb
```diff
@@ -147,4 +147,40 @@
...
+    producer = kafka.producer
+    (1..num_messages).each {|i| producer.produce("foo", topic: topic) }
+    (1..num_messages).each {|i| producer.produce("bar", topic: other_topic) }
+    producer.deliver_messages
+
+    group_id = "test#{rand(1000)}"
+    threads = [topic, other_topic].map do |topic_to_use|
+      t = Thread.new do
+        kafka2 = Kafka.new(seed_brokers: kafka_brokers, client_id: "test")
+        consumer = kafka2.consumer(group_id: group_id, offset_retention_time: offset_retention_time)
```
dasch: I'm not sure where `offset_retention_time` is defined, I would guess that the spec fails because these threads crash with NameError exceptions.

spec/functional/consumer_group_spec.rb
```diff
@@ -147,4 +147,40 @@
...
+        total = []
+        consumer.subscribe(topic_to_use, start_from_beginning: true)
+        begin
+          Timeout::timeout(10) do
```
dasch: I would discourage using Timeout – it wreaks havoc and is highly unpredictable.

spec/functional/consumer_group_spec.rb
```diff
@@ -147,4 +147,40 @@
...
+            consumer.each_message do |message|
+              total << message.value
```
dasch: I would do e.g. `consumer.stop if total.count == num_messages` instead of using Timeout.

dasch:
The code looks good, but I'd like to understand this use case before committing to supporting it in ruby-kafka – I only aim to support a subset of what the official library does, as I don't have matching resource. Why use a single group id when there's going to be two non-overlapping subgroups working on different topics?

nguyenquangminh0711:
I'll close this PR because it is lack of activities. Please feel free to re-open it @jcalvert

PR #499, opened 2017-12-07
Feature/lz4 codec

zerosign:
no test yet

### DIFF
--- a/lib/kafka/version.rb
+++ b/lib/kafka/version.rb
@@ -1,3 +1,3 @@
 module Kafka
-  VERSION = "0.5.1"
+  VERSION = "0.5.2"
 end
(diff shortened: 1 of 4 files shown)


lib/kafka/version.rb
```diff
@@ -1,3 +1,3 @@
 module Kafka
-  VERSION = "0.5.1"
+  VERSION = "0.5.2"
```
dasch: There's no reason to bump the ruby-kafka version.

dasch:
Can you add a unit test of Kafka:: LZ4Codec?

dasch:
Also, thanks for this (:heart:) and sorry for the late response.

zerosign:
sorry for the late response too ..

PR #500, opened 2017-12-07
Allow setting Kafka::Datadog.statsd

jonmagic:
This PR adds a `Kafka::Datadog.statsd` setter so you can re-use an existing dogstatsd client.

I ended up going this route after many attempts to configure `Kafka::Datadog` via the host and port setters. For some reason my already setup dogstatsd client could communicate with the ddagent but no matter what I tried the dogstatsd client `Kafka::Datadog` instantiated wouldn't.

Testing a theory I tried `Kafka::Datadog.instance_variable_set :@statsd, App.statsd` and tested and sure enough metrics started flowing through to Datadog.

### DIFF
--- a/lib/kafka/datadog.rb
+++ b/lib/kafka/datadog.rb
@@ -32,6 +32,10 @@ module Kafka
         @statsd ||= ::Datadog::Statsd.new(host, port, namespace: namespace, tags: tags)
       end
 
+      def statsd=(statsd)
+        @statsd = statsd
+      end
+
       def host
         @host ||= ::Datadog::Statsd::DEFAULT_HOST
       end
--- a/spec/datadog_spec.rb
+++ b/spec/datadog_spec.rb
@@ -2,30 +2,38 @@ require "kafka/datadog"
 require "fake_datadog_agent"
 
 describe Kafka::Datadog do
-  let(:agent) { FakeDatadogAgent.new }
-
-  before do
-    agent.start
-  end
-
-  after do
-    agent.stop
-  end
-
   it "emits metrics to the Datadog agent" do
-    Kafka::Datadog.host = agent.host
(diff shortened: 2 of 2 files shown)


dasch:
I think it would be a bit confusing to allow overriding the client while also allowing the host and port to be set. What version of Dogstatsd are you using?

jonmagic:
> What version of Dogstatsd are you using?

`3.1.0`

I'm really not sure why the instance of `Datadog::Statsd` that `ruby-kafka` instantiated wasn't working. I looked at my instance and the one ruby-kafka created side by side and the only thing that was different was the `UDPSocket:fd` value for the `@socket`.

dasch:
Can you paste your configuration code?

jonmagic:
Initially I configured ruby-kafka via the Racecar gem like this:

```ruby
Racecar.config.datadog_enabled = true
Racecar.config.datadog_host = "127.0.0.1"
Racecar.config.datadog_port = 8125
```

When I couldn't get that to work I made the changes in this PR and https://github.com/zendesk/racecar/pull/45 and everything works:

```ruby
statsd = Datadog::Statsd.new("127.0.0.1", 8125)
Racecar.config.datadog_enabled = true
Racecar.config.statsd = statsd
```

jonmagic:
I also set the namespace on both, forgot that detail.

dasch:
Do you have any logs that demonstrate stuff failing? You can enable logging in dogstatsd by setting `Datadog::Dogstatsd.logger = Logger.new(STDOUT)`.

dasch:
Closing this – we should figure out why the integration fails rather than circumvent the problem.

jonmagic:
I took a second pass and got it working with the host/port settings this time around, not sure what I did differently.

dasch:
Great!

PR #501, opened 2017-12-08
Instrument fetch_batch for each_message

michaelsauter:
Okay, this is a start. Not sure if this is the best place to do this. Also, I did not add the same instrumentation for `each_batch` as it would just duplicate `start_process_batch.consumer`. However, for consistency it might make sense to add it?

Let me know what you think and I'll adapt as necessary.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -202,6 +202,15 @@ module Kafka
         )
 
         batches.each do |batch|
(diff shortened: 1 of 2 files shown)


michaelsauter:
Hmm, tests failed but it looks unrelated. I ran `spec/consumer_spec.rb` locally and it passed.

dasch:
Thanks!

PR #502, opened 2017-12-11
Support custom group protocols

nguyenquangminh0711:
## What is this pull request for?
[Kafka official document](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Client-side+Assignment+Proposal) indicates that the topic partitions assignment should be done at client side. This gives Kafka client a great power and flexibility in controlling the whole consumer group. The current version of ruby-kafka goes with RoundRobinAssignmentStrategy as a hardcoded standard protocol within a consumer group and doesn't provide an interface to let user implement their own assignment strategy. This pull request fulfills the gap by adding an interface to set custom group protocols when declaring a consumer.

## Proposed usage
```ruby
require "kafka"
kafka = Kafka.new(seed_brokers: ["kafka1:9092", "kafka2:9092"])

consumer = kafka.consumer(
  group_id: "my-consumer",
  group_protocols: [
    {
      name: 'cpu_based',
      metadata: '5',
      assignment_strategy: CPUBasedAssignmentStrategy
    }
  ]
)
```

The `assignment_strategy` filed is a user-defined class that has only one method `assign`. The input is a hash of members and their metadata; a hash of topics and their partitions. The output is a hash of members and their `Kafka::Protocol::MemberAssignment` instances. The assignment strategy for above example:

```ruby
class CPUBasedAssignmentStrategy
  def assign(members:, topics:)
    group_assignment = {}
    members.keys.each do |member_id|
      group_assignment[member_id] = Kafka::Protocol::MemberAssignment.new
    end
    total_cpu = members.values.inject(0) do |total, metadata|
      total + metadata.to_i
    end
    topics.each do |topic, partitions|
      index = 0
      members.each_with_index do |member_id, metadata|
        partition_count = ((metadata.to_f * partitions.count) / total_cpu).round
        member_partitions = partitions[index..(index + partition_count)]
        group_assignment[member_id].assign(topic, member_partitions)

        index += partition_count
      end
    end
    group_assignment
  end
end
```
If the user doesn't specify a group protocol, RoundRobinAssignmentStrategy is the fallback protocol. Everything gonna work like before.

## What have I done in this pull request?
- Create `GroupProtocolManager` that manages the group protocols.
- Alter join group request and response to support consumer's group protocols.
- Replace `RoundRobinAssignmentStratgey`in `ConsumerGroup` by new `GroupProtocolManager`. The consumer group is now aware of current agreed group protocol and calls `GroupProtocolManager` to generate group assignment based on corresponding protocol.
- Simplify the  `RoundRobinAssignmentStratgey` so that it is not responsible for fetching topic metadata anymore.
- Add interface to client and consumer.

## Why do I think it is a good fit for ruby-kafka?
In fact, there are some pull requests related to this topic (https://github.com/zendesk/ruby-kafka/pull/105 , https://github.com/zendesk/ruby-kafka/pull/497). If we follow this approach, we could support many standard protocols instead of debating which one should be used exclusively. And since ruby-kafka is a general client library, I think it is important to support different use cases of users with ease, that is also achievable by this approach.

---
Please let me know your thought about this pull request, and sorry for my bad English. I'm not a native speaker btw 😓  
/cc @dasch


dasch:
@nguyenquangminh0711 thanks for taking the time for this!

I'm wary of adding too many customization options to ruby-kafka, as each comes with a support and maintenance cost. Do you have concrete use cases where the current round-robin strategy has failed?

nguyenquangminh0711:
Hi @dasch, this pull request is inspired by one of my own use case. Say I have a set of consumers and each consumer has a different computing capacity; for example, consumer 1 has 2 virtual CPUs, consumer 2 has 5 virtual CPUs and consumer 3 has just 1 virtual CPU. They all consume from the same topic and do exactly the same job. Each message of the topic is costly and takes a significant amount of CPU resource to complete. When I go with default round-robin strategy, each consumer is responsible for the same amount of partitions, as well as the same amount of messages in most case. This assignment makes an imbalance between the consumers. The consumer 1 consumes the messages at a normal rate while consumer 2 is free all the time and consumer 3 is busy all the time and could not catch up with the message input rate. That is a waste of resources.

I know that in perfect condition, all the consumers should have the same computing capacity. However, in many cases, the consuming rate of each consumer is affected by different factors, like CPU, RAM or even user-defined condition. So, I think providing a customization option for work distribution is a fair call. About the support and maintenance cost, IMO, this customization option is simple and straight-forward. In fact, if the users don't set any custom protocol, everything still works out of the box like before. If users choose the different path, we just need to ensure we provide them the right arguments, they will take care of their own assignment strategy since this is an advanced usage after all.

dasch:
The philosophy behind ruby-kafka is to make Kafka usable to non-experts, and as such I'm not sure this is the right approach.

A different way to look at your problem is in terms of _backpressure_ to the scheduler (in this case the consumer instance that happens to do the assignments).

The idea would be this: each consumer instance keeps track of the offset lag of each of its assigned partitions. When there's a rejoin, it pushes that info in the custom data field in the join request. The consumer chosen to assign partitions will receive these data fields and can decode them. We can then do an assignment strategy that tried to assign partitions with big offset lags in a fair way.

The only downside is that this will only be done when there's a rejoin, so we may have to add a mechanism for triggering that. But it's doable and would help everybody.

I'd rather have a single, well-thought-out assignment strategy rather than adding a bunch of customizability.

nguyenquangminh0711:
Hi @dasch, it is good to know the philosophy behind ruby-kafka. If you choose to follow that path, I totally agree that my proposal is not a good fit. Your idea of controlling the assignment based on the consuming lags is a great idea. I'll research and follow this approach. Thanks for your feedback. I'll close this pull request then.

dasch:
Thanks!

Issue #503, opened 2017-12-12
[suggestion] Producing byte arrays and not only Strings (incl. binary strings)


syndbg:
We have an use case where we would want to produce byte arrays.

The following snippet is a small example with bytes array:

```Ruby
require "kafka"

kafka = Kafka.new(
  # At least one of these nodes must be available:
  seed_brokers: ["kafka1:9092", "kafka2:9092"],

  # Set an optional client id in order to identify the client to Kafka:
  client_id: "my-application",
)

# This is where it's important
kafka.deliver_message("ABC".bytes, topic: "example")
```

Obviously the difference is that the current implementation depends on Strings.

dasch:
Closing due to a lack of activity. Please open a new issue if this is still a problem.

PR #504, opened 2017-12-12
[#503] Producing with byte arrays

syndbg:
References issue https://github.com/zendesk/ruby-kafka/issues/503

### DIFF
--- a/lib/kafka/protocol/encoder.rb
+++ b/lib/kafka/protocol/encoder.rb
@@ -93,14 +93,23 @@ module Kafka
 
       # Writes a byte string to the IO object.
       #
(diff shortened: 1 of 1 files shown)


dasch:
Since the Kafka API already expects bytes, would it be wrong to just always convert strings into byte arrays before encoding?

syndbg:
> Since the Kafka API already expects bytes, would it be wrong to just always convert strings into byte arrays before encoding?

No..?

dasch:
I should probably get a better understanding of your problem first. Does your application break if you simply call `produce(my_bytes.to_s, ...)`, and if so, how?

Issue #505, opened 2017-12-19
Compressed messages are yielded with incorrect offsets


klippx:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.4.1
* Version of Kafka: 0.10.2.0
* Version of ruby-kafka: 0.5.1

- [x] Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.

###### Steps to reproduce

See failing tests and investigation below

###### Summary

We are investigating a case which causes our consumer to crash and reconsuming it's whole partition. It is about Ruby Kafka not returning the correct offsets for the fetched batch of compressed messages.

This happens dozens of times on a daily basis.

We are using Phobos. Here you can see our debug level log output:

https://gist.github.com/klippx/65c4ef06eb0e01cbde6d0640e3176f1f

###### Analysis

772033 is not compacted:

```ruby
> kafka.fetch_messages(topic: 'eu-live.kred.account-events', partition: 7, offset: 772033, max_bytes: 1).map { |v| v.offset }
=> [772033]
```

772034 however, IS compacted:

```ruby
> kafka.fetch_messages(topic: 'eu-live.kred.account-events', partition: 7, offset: 772034, max_bytes: 1).map { |v| v.offset }
=> [772043, 772044, 772045, 772046, 772047, 772048, 772049, 772050, 772051, 772052]
```

So, offset 772034 contains a compacted message with 10 messages inside, as you can see above. But the offsets are not handled correctly by ruby-kafka, it should be 10 items ending with 772043:

```ruby
=> [772034, 772035, 772036, 772037, 772038, 772039, 772040, 772041, 772042, 772043]
```

###### Preparing an experiment

We got some hints from our kafka cluster team that our client may not handle compressed messages well. So, in order to see where things go wrong, I put some debug messages to try to find the issue.

In `message.rb`:

```ruby
Kafka::Protocol::Message.class_eval do
      def initialize(value:, key: nil, create_time: Time.now, codec_id: 0, offset: -1)
        @key = key
        @value = value
        @codec_id = codec_id
        @offset = offset
        @create_time = create_time
        puts "Message> Initialized new Message offset=#{offset}"
        @bytesize = @key.to_s.bytesize + @value.to_s.bytesize
      end

      def decompress
        puts 'Message> Decompress messages...'
        codec = Kafka::Compression.find_codec_by_id(@codec_id)

        # For some weird reason we need to cut out the first 20 bytes.
[...]

PR #506, opened 2017-12-19
Fix compressed offset bug

klippx:
Addresses #505 

We are happy to change the test, since we were not sure how to set up the test scenario properly we used the "raw" approach of simply using data from our system console.


dasch:
This should be updated to reflect https://github.com/zendesk/ruby-kafka/issues/505#issuecomment-352826114

lib/kafka/protocol/message.rb
```diff
@@ -58,10 +58,14 @@ def decompress
         message_set_decoder = Decoder.from_string(data)
         message_set = MessageSet.decode(message_set_decoder)
 
+        max_relative_offset = message_set.messages.last.offset
+        return message_set if max_relative_offset == offset
+
         # The contained messages need to have their offset corrected.
+        base_offset = offset - max_relative_offset
         messages = message_set.messages.each_with_index.map do |message, i|
```
zmstone: i no longer need

lib/kafka/compressor.rb
```diff
@@ -28,14 +28,15 @@ def initialize(codec_name: nil, threshold: 1, instrumenter:)
 
     # @param message_set [Protocol::MessageSet]
     # @return [Protocol::MessageSet]
-    def compress(message_set)
+    def compress(message_set, offset: -1)
```
dasch: Can you document that `offset` is just used to simulate the broker behavior?

lib/kafka/protocol/message.rb
```diff
@@ -58,10 +58,14 @@ def decompress
         message_set_decoder = Decoder.from_string(data)
         message_set = MessageSet.decode(message_set_decoder)
 
+        max_relative_offset = message_set.messages.last.offset
+        return message_set if max_relative_offset == offset
```
dasch: This definitely needs a comment explaining why that's the case :)

spec/compressor_spec.rb
```diff
@@ -2,29 +2,6 @@
...
-      decoded_message_set = decoded_message.decompress
-      messages = decoded_message_set.messages
-
-      expect(messages.map(&:value)).to eq ["hello1", "hello2"]
-
-      # When decoding a compressed message, the offsets are calculated relative to that
-      # of the container message. The broker will set the offset in normal operation,
-      # but at the client-side we set it to -1.
-      expect(messages.map(&:offset)).to eq [-1, 0]
-    end
```
dasch: Why remove this?

dasch:
Thanks for the PR! In order to ensure this doesn't break in the future, I'd love to get a few comments in the code describing the behavior.

spec/compressor_spec.rb
klippx: We moved tests to message_set_spec, and we beleived this case was already covered in the other three tests we provided. This test does has -1 as wrapper message offset, and both messages are -1, -1 as well. Hence our code will accept this as "correct from broker side" and return `[-1, -1]`

In the end we just thought the other test cases was more realistic. Should we add this back?

dasch: OK – no, that's fine, just needed to follow along.

dasch:
Looks good! Can you verify that this works with real workloads?

lib/kafka/protocol/message.rb
```diff
@@ -113,6 +102,37 @@ def self.decode(decoder)
...
+      # Offsets may be relative with regards to wrapped message offset, but there are special cases.
+      #
+      # Cases when client will receive corrected offsets:
+      #   - When fetch request is version 0, kafka will correct relative offset on broker side before replying fetch response
+      #   - When messages is stored in 0.9 format on disk (broker configured to do so).
+      #
+      # All other cases, compressed inner messages should have relative offset, with below attributes:
+      #   - The container message should have the 'real' offset
+      #   - The container message's offset should be the 'real' offset of the last message in the compressed batch
+      #   - The first inner message should always have offset = 0
```
zmstone: sorry, I was wrong about this.
It doesn’t always start with 0

klippx: thanks, amended 👍🏼

klippx:
@dasch We are going on christmas vacation right now, so we wont be able to work on this for about 3 weeks. So, this may take a while unfortunately i.e. creating a long feedback loop. Up to you. We are running 0.5.0 while we are on vacation which is stable for us. 

Besides, even if we did run this version our consumers would use the short-circuit of not fixing the offsets since last message offset == wrapped message offset so it wont be a fair test.

dasch:
@klippx a change of this magnitude requires some real-life testing before I dare merge it – let's wait until you're back.

zmstone:
This pr is about the same magnitude as 42821e9f
To be fair, that change was not covered by real-life testing either.
I'm not saying that tests are not important,
just that all consumers running current master code against compressed messages
are committing wrong offsets and likely skipping messages.
This change may have bug, but should not make things worse.

piavka:
totally agree with @zmstone

dasch:
OK, seems like I have a volunteer who can test it. I'll merge and cut a pre-release.

PR #516, opened 2018-01-03
Shut down the consumer more quickly

dasch:
The retry logic would sometimes cause the consumer to be stuck in a fetch loop even when it's supposed to be shutting down.


kirs:
Thank you for the fix!

Is there a test that we can add to make sure there won’t be a regression?

PR #521, opened 2018-01-06
Add support for using the system's default CA cert store

bradurani:
Adds support for using the CA certs installed on your system by default for SSL. This is useful for connecting to Kafka clusters on public domains, for instance clusters hosted on public clouds.

https://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL/X509/Store.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -888,6 +888,21 @@ kafka = Kafka.new(
 
 Without passing the CA certificate to the client it would be impossible to protect against [man-in-the-middle attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack).
 
+##### Using your system's CA cert store
+
+If you want to use the CA certs from your system's default certificate store, you
+can use:
+
+```ruby
+kafka = Kafka.new(
+  use_ssl_ca_default_store: true
+  # ...
+)
+```
+
+This configures the store to look up CA certificates from the system default certificate store on an as needed basis. The location of the store can usually be determined by: 
+`OpenSSL::X509::DEFAULT_CERT_FILE`
+
 ##### Client Authentication
 
 In order to authenticate the client to the cluster, you need to pass in a certificate and key created for the client and trusted by the brokers.
--- a/README.md
+++ b/README.md
@@ -895,7 +895,7 @@ can use:
 
 ```ruby
 kafka = Kafka.new(
-  use_ssl_ca_default_store: true
+  ssl_ca_certs_from_system: true
   # ...
 )
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -888,6 +888,21 @@ kafka = Kafka.new(
...
+
+```ruby
+kafka = Kafka.new(
+  use_ssl_ca_default_store: true
+  # ...
+)
+```
+
+This configures the store to look up CA certificates from the system default certificate store on an as needed basis. The location of the store can usually be determined by: 
+`OpenSSL::X509::DEFAULT_CERT_FILE`
```
bradurani: This resolves to `/usr/local/etc/openssl/cert.pem` on my system

bradurani:
A little background on this, I'm trying to connect to ConfluentCloud, which does not supply SSL certs. It's hosted on AWS via the `confluent.cloud` domain:
```
Certificate chain
 0 s:/CN=kafka-xxxxxxxxxxx.us-west-2.aws.confluent.cloud
   i:/C=US/O=Amazon/OU=Server CA 1B/CN=Amazon
```
My understanding is that since Amazon is a root CA, this cert can be verified by the bundle installed on my system by default.

On my system `OpenSSL::X509::DEFAULT_CERT_FILE` resolves to `/usr/local/etc/openssl/cert.pem`

dasch:
Is there a reason why we wouldn't enable this by default, even removing the option entirely?

@tcrayford you're my go-to SSL guy, what's your take here? Heroku obviously took a different route, was there a reason for that?

trthomps:
@dasch I think the only reason not to have it default set is that you'd have to assume that ssl was enabled by default, which probably isn't the case for a lot of Kafka instances.  As it is now, SSL only gets turned on if you specify a certificate bundle or if you set this new option.

Otherwise, a new option would have to be added to turn off SSL, which maybe it the right way to go since it's good to encourage encryption by default, but would break backwards API compatibility 😄.  

Or code could be added to attempt SSL first and if that fails, try without SSL?  That would keep backwards compatibility and make SSL seamless for people using trusted certificate authorities.

tcrayford:
@dasch most of our stuff was for infrastructural reasons, not anything else. I'm happy with this approach.

dasch:
Okay, this seems like the right approach, then.

README.md
```diff
@@ -888,6 +888,21 @@ kafka = Kafka.new(
...
 Without passing the CA certificate to the client it would be impossible to protect against [man-in-the-middle attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack).
 
+##### Using your system's CA cert store
+
+If you want to use the CA certs from your system's default certificate store, you
+can use:
+
+```ruby
+kafka = Kafka.new(
+  use_ssl_ca_default_store: true
```
dasch: Can you rename this to `ssl_ca_certs_from_system: true`?

dasch:
Thanks! :heart:

PR #524, opened 2018-01-10
change log message to debug

mreinsch:
The "sending heartbeat" message is spamming our logs. I'd like to propose to switching that to debug.


dasch:
Makes sense. Thanks!

PR #526, opened 2018-01-18
Fetch batches asynchronously

dasch:
Experimental branch that adds a background thread that fetches messages.


mensfeld:
Will this break the current API or will it be just a config option?

dasch:
It will be an internal change.

mensfeld:
Question: once it is done (which is super exciting) - will it be optional or not?

dasch:
@mensfeld it won't be optional – it'll be the only mode of operation.

mensfeld:
@dasch great (really)! I am looking forward to this one!

dasch:
@mensfeld one thing that would help get this released sooner would be testing an a staging or production environment – I probably won't have time for that right now, but if you can validate that this branch works as expected it'll give me a lot more confidence! The branch should be completely on parity with master at this point.

mensfeld:
@dasch can do. Will give it a spin in the week after the next one and will let it run for around a week.

dasch:
Awesome!

mensfeld:
@dasch I didn't use it yet for mission critical stuff but it seems to work for a lower-load topics. The only thing that I'm not sure is the error handling and back-off as I tend to write reliable code :D

dasch:
😉 

OK, I'll release another 0.5 version, then merge this and release 0.6.0.beta1 next week. It needs a little time in production before I'm confident in releasing a final version.

mensfeld:
Great. With a beta1 I can give it a spin on a bigger scale as it will be easier for me to lock the dependencies.

mensfeld:
@dasch sure. Will upgrade and deploy it tomorrow morning.

dasch:
:heart:

lib/kafka/fetcher.rb
```diff
@@ -0,0 +1,178 @@
+require "kafka/fetch_operation"
+
+module Kafka
+  class Fetcher
+    MAX_QUEUE_SIZE = 100
```
mensfeld: Could you please document that variable? does it describe the max ahead buffer size or max fetch size? What the queue is? Thanks! :)

PR #528, opened 2018-01-19
Delete Topic API

nguyenquangminh0711:
The management API group already has CreateTopic API. This pull request is to implement DeleteTopicAPI. The implementation is based on official documentation https://kafka.apache.org/protocol#The_Messages_DeleteTopics. I choose API version 0 because, in the newest version, which is 1, the only difference is `throttle_time_ms` field and it seems like we don't have a proper way to handle throttling. Therefore, the version 0 is safer.

### DIFF
--- a/spec/functional/topic_management_spec.rb
+++ b/spec/functional/topic_management_spec.rb
@@ -9,4 +9,17 @@ describe "Topic management API", functional: true do
 
     expect(partitions).to eq 3
   end
+
+  example "deleting topics" do
+    topic = generate_topic_name
+    kafka.create_topic(topic, num_partitions: 3)
+
+    expect(kafka.partitions_for(topic)).to eq 3
+    kafka.delete_topic(topic)
+    kafka.instance_variable_get(:@cluster).mark_as_stale!
+
+    expect do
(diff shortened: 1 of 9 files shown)


spec/functional/topic_management_spec.rb
```diff
@@ -9,4 +9,17 @@
...
     expect(partitions).to eq 3
   end
+
+  example "deleting topics" do
+    topic = generate_topic_name
+    kafka.create_topic(topic, num_partitions: 3)
+
+    expect(kafka.partitions_for(topic)).to eq 3
+    kafka.delete_topic(topic)
+    kafka.instance_variable_get(:@cluster).mark_as_stale!
```
dasch: This cluster should be marked as stale by `Cluster#delete_topic`.

spec/functional/topic_management_spec.rb
```diff
@@ -9,4 +9,17 @@
...
+
+    expect do
+      kafka.partitions_for(topic)
+    end.to raise_error(Kafka::LeaderNotAvailable)
```
dasch: Can you use `Client#has_topic?` here instead?

PR #530, opened 2018-01-22
Describe topic API

nguyenquangminh0711:
### Proposed solution:
Since we are following simple approach for the users, the admin API for topic description is quite simple:

```
client.describe_topic("topic", ["retention.ms", "retention.bytes non_exists"])
# Returns
{
  "retention.ms"=>"604800000", 
  "retention.bytes"=>"-1"
}
```

The underlying API is [DescribeConfigs API](https://kafka.apache.org/protocol#The_Messages_DescribeConfigs). Currently, there is only one interface `describe_topic` but in future, it is easy to pretty easy to add other similar interfaces to support other resource types.


nguyenquangminh0711:
I'll wait for this pull request https://github.com/zendesk/ruby-kafka/pull/533 to be merged to resue `client#support_api?` before running tests.

dasch:
What is the use case for adding this API?

nguyenquangminh0711:
@dasch The main purpose of this pull request is to introduce DescribeConfigs API, and later AlterConfigs API. This API is powerful for Kafka administration or operation since it provides full information like `compression.type`, `max.message.byte`, etc. which are missing from Metadata responses. The `describe_topic` is just a concrete use case for the generic DescribeConfigs API. 

From recent pull requests and your todo in github project, I assume that the 0.6.x provide support some interfaces / APIs for administrations. I think this would fulfill the goal of the next version. Or am I missing something?

dasch:
I'm mostly thinking: what concrete tool would use this API? I'm wary of introducing something like this without knowing how it will be used. Once an API is there it's difficult to change, so I'd like to consider use cases before adding more.

nguyenquangminh0711:
@dasch Ah, about the concrete use case, it's almost for debugging and monitoring. I'm working on multiple Kafka clusters and even in the same cluster, each topic has a different set of configurations. So, the need of fetching, updating and syncing configurations between different environments or clusters usually pop up. It would be pretty nice and helpful if those could be handled programmatically via the client interfaces (like Java / Golang clients also supported this). That's my use case. I would admit that this need is non-trivial and not sure about the usabilities of others too. 🤔

PR #533, opened 2018-01-23
 Increase topic partition count API

nguyenquangminh0711:
### Proposed usage
```ruby
client.create_partitions_for("topic", num_partitions: 10)
```
This API is available after Kafka 1.0.0. The API is implemented based on [official documentation](https://kafka.apache.org/protocol#The_Messages_CreatePartitions). Currently, the interface doesn't support custom partition assignments because I don't think it aligns with the goal of the gem.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -479,12 +479,10 @@ module Kafka
     # @param num_partitions [Integer] the number of desired partitions for
     # the topic
     # @param timeout [Integer] a duration of time to wait for the new
-    # @param validate_only [Boolean] whether this request is to validate only
-    # without actually execute it
     # partitions to be added.
     # @return [nil]
-    def create_partitions_for(name, num_partitions: 1, timeout: 30, validate_only: false)
-      @cluster.create_partitions_for(name, num_partitions: num_partitions, timeout: timeout, validate_only: validate_only)
+    def create_partitions_for(name, num_partitions: 1, timeout: 30)
+      @cluster.create_partitions_for(name, num_partitions: num_partitions, timeout: timeout)
     end
 
     # Lists all topics in the cluster.
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -545,8 +545,8 @@ module Kafka
     # @param api_key [Integer] API key.
     # @param version [Integer] API version.
     # @return [Boolean]
-    def support_api?(api_key, version = nil)
-      @cluster.support_api?(api_key, version)
+    def supports_api?(api_key, version = nil)
+      @cluster.supports_api?(api_key, version)
     end
 
     def apis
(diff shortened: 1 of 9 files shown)


lib/kafka/client.rb
```diff
@@ -473,6 +473,20 @@ def delete_topic(name, timeout: 30)
...
     end
 
+    # Create partitions for a topic.
+    #
+    # @param name [String] the name of the topic.
+    # @param num_partitions [Integer] the number of desired partitions for
+    # the topic
+    # @param timeout [Integer] a duration of time to wait for the new
+    # @param validate_only [Boolean] whether this request is to validate only
+    # without actually execute it
```
dasch: I don't see any use cases for `validate_only` in ruby-kafka – do you have any? Otherwise I think it's better to leave it out (and just pass `false`).

lib/kafka/client.rb
```diff
@@ -526,6 +540,15 @@ def last_offsets_for(*topics)
       }.to_h
     end
 
+    # Check whether current cluster supports a specific version or not
+    #
+    # @param api_key [Integer] API key.
+    # @param version [Integer] API version.
+    # @return [Boolean]
+    def support_api?(api_key, version = nil)
+      @cluster.support_api?(api_key, version)
```
dasch: Can you rename `support_api?` to `supports_api?`?

lib/kafka/protocol/encoder.rb
```diff
@@ -30,7 +30,8 @@ def write(bytes)
       # @param boolean [Boolean]
       # @return [nil]
       def write_boolean(boolean)
-        write(boolean ? 0x1 : 0x0)
+        # Note: 0x1 is represented by 1 byte when being written into the socket
+        write(boolean ? write_int8(1) : write_int8(0))
```
dasch: This seems wrong – you're writing the result of `write_int8`.

spec/functional/topic_management_spec.rb
```diff
@@ -19,4 +19,30 @@
     kafka.delete_topic(topic)
     expect(kafka.has_topic?(topic)).to eql(false)
   end
+
+  example "create partitions" do
+    unless kafka.support_api?(Kafka::Protocol::CREATE_PARTITIONS_API)
+      skip("This Kafka version not support ")
```
dasch: Ah, now I see why it might be handy to include `support_api?`.

nguyenquangminh0711:
@dasch Updated as your comment. About the `encoder#write_boolean`, the comment is wrong, however we need to pack the boolean because 1 (Fixnum) will be converted to string before writing it down to the socket. So, the `@io.write(0x1` actually writes string "1", not 8-bit boolean as we want.

dasch:
@nguyenquangminh0711 in that case, shouldn't we just do `boolean ? write_int8(1) : write_int8(0)` and not the outer `write`?

nguyenquangminh0711:
@dasch Oh you are right. The outer `write` is redundant. I updated it 👍

dasch:
Thanks!

PR #537, opened 2018-01-25
Instrument the consumer loop

dasch:


lib/kafka/consumer.rb
```diff
@@ -371,7 +371,9 @@ def consumer_loop
 
       while @running
         begin
-          yield
+          @instrumenter.instrument("loop.consumer") do
```
mensfeld: won't this have a significant performance cost? AS notifications are global and not the fastest and this code is executed quite often (for each batch and/or each message depending on the settings).

dasch: We already instrument much more granular events.

PR #538, opened 2018-01-26
Allow passing in the seed brokers in a positional argument

dasch:
This is a required argument, and it's pretty central to the concept, so passing it as a positional argument makes sense.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -244,6 +244,8 @@ module Kafka
   # @see Client#initialize
   # @return [Client]
   def self.new(seed_brokers = nil, **options)
+    # We allow `seed_brokers` to be passed in either as a positional _or_ as a
+    # keyword argument.
     if seed_brokers.nil?
       Client.new(**options)
     else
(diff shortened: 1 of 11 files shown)


lib/kafka.rb
```diff
@@ -243,8 +243,14 @@ class FailedScramAuthentication < SaslScramError
...
   # @return [Client]
-  def self.new(**options)
-    Client.new(**options)
+  def self.new(seed_brokers = nil, **options)
+    # We allow `seed_brokers` to be passed in either as a positional _or_ as a
+    # keyword argument.
+    if seed_brokers.nil?
+      Client.new(**options)
+    else
+      Client.new(seed_brokers: seed_brokers, **options)
```
mensfeld: question for the future: will both of them be supported or will the non positional be deprecated?

dasch: In the future, major release, the non-positional one will probably be deprecated and removed. I'll give it at least 6-9 months first, though.

PR #540, opened 2018-02-18
Add support for config entries in the topic creation API

tjwp:
I wanted to add this support so that I can create a topic that uses log compaction.

I did not add any validation of the config entry name and values here because the service will return an error if the name is not recognized or the value is invalid.


dasch:
Thanks! :heart:

Issue #527, opened 2018-01-19
Add property tests


dasch:
For the consumer:

* Messages are processed in-order.
* There are no missing messages.
* There are no duplicate messages under normal operation.

PR #543, opened 2018-02-22
Add more tests to ensure consumer properties

nguyenquangminh0711:
Add one more test to ensure each consumer consumes the messages in-order, non-duplicate and non-missing; from the recommendation https://github.com/zendesk/ruby-kafka/issues/527


dasch:
:+1:

PR #548, opened 2018-03-22
Kafka 0.11 record batch message format

nguyenquangminh0711:
This pull request is to add the support for Kafka 0.11's record batch message format. However, the scope of this pull request just stops at encoding / decoding a record batch. The integration into current consumer / producer operations will be implemented later since this pull request is already big and I still have some concerns about the new record batch format. This implementation is based on the official protocol documentation of [Record batch](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets). I add unit tests and manually test against other implementations in Kafka's java client and golang client. 

Please let me know what you think about this @dasch.


dasch:
Thanks for all this work! The 0.11 format is definitely on the roadmap, but I'm not sure I'll be able to support both it _and_ the 0.10 format, so it might be a little while before this gets released.

That being said, I'd love to see this progress. Would it be possible to do the producer or consumer integration in this PR, so that everything can come together and we can do an integration test?

dasch:
Note that I'll be on vacation next week, but will come back and look more into this after that.

nguyenquangminh0711:
@dasch Glad to hear that. That's also my concern when doing the integration. At the first try, it looks like it is much more complicated to support both 0.10 and 0.11 version at the same time. But if it is okay to get rid of version 0.10, everything becomes clear now. I'll implement the consumer and producer integration in this PR. Hope that would finish after you come back. Have a nice trip 🎉

dasch:
Thanks! Note that it may be a while before I'm ready to merge – this would mean abandoning support for Kafka 0.10, and I'd need to gauge how many people would be affected by that.

nguyenquangminh0711:
Hi @dasch, some updates for this PR:

- I integrated the record batch into the current consumer / producer flow. The Producer API is updated to version 3 and the Fetch API is updated to version 4. 

- All the integration tests passed. Since this PR doesn't support any new features for record batch, I think we don't need new integration tests targeting the Record Batch format.

- I'm afraid that we have to add dependency `digest/crc32` (or another similar gem) because Kafka 0.11.0 starts to use Castagnoli CRC32 algorithm, which is not supported by Ruby standard libs. I believe that if we implement that checksum mechanism by ourselves, it would be error-prone and increase the burden of maintenance.

- All the stuff belongs transaction is simply ignored. To implement that, it would take a huge amount of time. I don't think we should do it for now.

I still have some unfinished parts, such as supporting compressing options, cleanup obsoleted codes, update comments, etc.

dasch:
> I double-checked and added the support to fetch transactional messages. The consumer now could consume non-transaction messages and transactional ones. However, there is a limitation: Consumer fetches committed messages only. The "dirty" messages are not supported. And I could not think of any use cases that needs the dirty, uncommitted messages at the moment.

:+1:

> In each transaction, the record batch includes an empty record named control record. This record is blank and should not be handled by the application layer although its offset must be committed. So I changed Consumer#each_message and Consumer#each_batch a little bit to discard such control records.

:+1:

klippx:
Just to join the discussion - Do we absolutely necessarily need to get rid of 0.10 support?

I mean there are options: We can aim for this library to support 0.10+ or we can decide to drop support for 0.10.

Keeping support for 0.10 might require different code structure, if that is too much effort then abandoning 0.10 might be tempting. Otherwise, I am sure this can be solved and everyone will be able to use the latest version of ruby-kafka. 

Dropping 0.10 support means people will be stuck on e.g. `<= 0.6` version with limited support from the maintainers, probably only critical bugfixes. We might have a slightly simpler code base in >= 0.7.0 and less tests to run on CI, but I am not sure these benefits are massive. We might have to keep releasing some fixes in the parallell track which doesn't seem too "slick"

Just saying that let's merge this but maybe make changes first in a way that 0.10 can still be supported. Unless I missed something and it is _absolutely_ _necessary_ to say goodbye to 0.10 ?

tcrayford:
it's probably worth noting that all the 0.10 version of kafka are no longer supported by the community, so if you're running it in prod, you'll need to upgrade brokers to 0.11 anyway (unless you're ok with running old software with bugs in it!).

dasch:
I don't have enough time to provide support for both 0.10 and 0.11. The added code complexity will be significant and frankly not worth it – I'd rather wait until most people are off 0.10 (which they should be by now) and then focus exclusively on 0.11.

nguyenquangminh0711:
@klippx: From the implementation aspect, it is doable to support both 0.10 and newer version. In fact, in my pull request, the consumer can still consumes from Kafka 0.10. However, like @dasch and @tcrayford already point out, the efforts of maintaining outdated and abandoned Kafka version will be a burden for future development of this project. Therefore, I think the current approach is still good.

@dasch Since there are a lot of works to prepare for Kafka 0.11 support, do you think we should create a branch, like 0.7, then merge this pull request into that branch and let everyone continue to work on that branch parallel with the current 0.6 versions? I'm afraid that if we wait until this PR is merged into the master branch, it would take much longer time to develop new features like transactions.

dasch:
@nguyenquangminh0711 my plan was to release v0.6 as the final minor version with 0.10 support, then move master to support 0.11 only – I'd create a `v0.6-stable` branch in order to allow backports and critical bug fixes for v0.6, but would otherwise focus exclusively on 0.11. I don't have the bandwidth to really support both. I appreciate your patience – v0.6 final will happen once I'm confident that the new consumer fetcher engine works well enough. Any help with testing it will speed this process up.

klippx:
We are looking into 0.11 for Phobos, and we discover that we are lacking a way to set headers in Producer#produce.

It would be nice if this PR includes
- Producer#produce should accept parameter headers: nil
- Producer#assign_partitions! should include headers when doing buffer.write
- PendingMessage#initialize accept headers
- MessageBuffer#write include headers (and change usages of this to include it)
- Protocol::Record#initialize includes headers (and change references of it to include this)
- Protocol::Record#initialize includes headers (and change references of it to include this)
- Client#deliver_message should include headers when doing buffer.write
- Consumer#each_message should include headers into notification variable
- Instrumentation produce_message.producer & ack_message.producer should include headers

Maybe I missed something.

Do we need to beef out the specs? Mainly I was looking at the most changed stuff that doesnt have specs:
- record_spec.rb 
- compressor_spec.rb 
- fetch_response_spec.rb 
- integration spec verifying the produce / consume works with the new fields such as header

Also, maybe you should get rid of unused classes such as:
- Protocol::Message (this is now Record)
- Protocol::MessageSet (this is now RecordBatch)
- something else?

dasch:
I hope to move `master` to only support Kafka 0.11 within the next few weeks, after which we can merge this and discuss APIs for header support. For now it's fine to not allow passing those in. I'd rather this PR focuses on getting the protocol right and that it's exhaustively tested against a real-world Kafka cluster.

nguyenquangminh0711:
@dasch The master branch changed to use Fetcher, which creates an edge case that the `decoder#peek` method didn't handle.  I fixed the failed test and added some more tests. I think that's okay for reviewing now.  💪

dasch:
I've released `v0.7.0.alpha1` – please test this out in production for a while so that I can move to release a stable version!

nguyenquangminh0711:
🎉 🎉 🎉

mensfeld:
:heart:

Elyahou:
@nguyenquangminh0711 really good work !

I tried to use `ruby-kafka` with these changes but I saw a big performance degradation :(, especially I see that the client, get only few messages/batch, and so fetch lot more of little batches. I use the default configuration of `ruby-kafka` and running against kafka cluster with version 1.1

What do you mean by `The integration into current consumer / producer operations will be implemented later since this pull request is already big and I still have some concerns about the new record batch format.`

nguyenquangminh0711:
Hi @Elyahou, I'm really glad that you use the beta version to test out record batch format. Could I know more about your Kafka settings? Like number of consumers, number of partitions, average size of each message, replication factor, etc. That would help a lot with the debugging process. I just have a simple benchmark on my machine and it is true that the speed is unacceptable. I'll work on it to provide a batch asap. Thanks for you report and sorry for this 😞

nguyenquangminh0711:
Found the bug. It turns out a fetch response consists of multiple record batch next to each order. The tails of record batches could has some in-completed record thanks to max_bytes setting in the FetchRequest. This partial trailing record case are not handled properly. I'll push the patch and provide some tests for this bug.

dasch:
Please test again with alpha3.

mensfeld:
@nguyenquangminh0711 any update on the headers integration? Will they we available for both producer and consumer? I may look into wrong places but I don't see support of it in the master branch.

nguyenquangminh0711:
@mensfeld Hi, it is on the halfway, I didn't follow up for a while. I'll reboot and continue on that feature. Perhaps it will be ready in some few days. I'll ping you when it is done. 💪

mensfeld:
@nguyenquangminh0711 thank you for the update :muscle:

lib/kafka/protocol/fetch_request.rb
```diff
@@ -46,6 +48,7 @@ def encode(encoder)
         encoder.write_int32(@max_wait_time)
         encoder.write_int32(@min_bytes)
         encoder.write_int32(@max_bytes)
+        encoder.write_int8(ISOLATION_READ_COMMITTED)
```
jwilger, line 51: @nguyenquangminh0711 : Sorry, I know this change is ancient history, but do you happen to recall why this is hard-coded to `ISOLATION_READ_COMMITTED` when it looks like there was some intent to make this configurable (given that `ISOLATION_READ_UNCOMMITTED` is defined?) I have a situation where I need to be able to read uncommitted messages, but I want to make sure there isn't some good reason for not having this be configurable before I go down that path.

nguyenquangminh0711: @jwilger This hard-coding was for the sake of simplicity. This implementation was complicated enough that it would be hard to ensure all the configurations worked well. Therefore, I wanted to focus on read committed first, then implemented the custom configuration to allow semantic configuration later. However,  I never had a use case for Read uncommitted, and nobody asked about this until now. So, I think it's feasible to add a configuration for this.

jwilger: Perfect, thank you! We will likely be submitting a PR for it soon-ish (for the moment, we just forked ruby-kafka and flip-flopped the hard-coding, because we needed to get it working quickly due to a stuck partition. N.B. it's no fun when your kafka cluster gets restarted in the middle of a transaction.)

Issue #552, opened 2018-03-31
Consumer#pause doesn't pause processing fetched messages


tmattia:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.4.2
* Version of Kafka: 0.11.0.1
* Version of ruby-kafka: 0.5.4

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.

###### Steps to reproduce

```ruby
require 'ruby-kafka'
kafka = Kafka.new(seed_brokers: ['kafka://127.0.0.1:9092'])
# Rand is on purpose just to prove, that the same message is being fetched
kafka.deliver_message(rand.to_s, topic: 'a')
kafka.deliver_message(rand.to_s, topic: 'a')
kafka.deliver_message(rand.to_s, topic: 'a')
consumer = kafka.consumer(group_id: 'consumer-name')
consumer.subscribe('a')
consumer.each_message(automatically_mark_as_processed: false) do |message|
  p message
  consumer.pause(message.topic, message.partition, timeout: 5)
end
```

###### Expected outcome

I would expect it process the first message from the topic and immediately stop, not continue processing the messages that were already fetched. When resuming, it should resume from the same offset where it stopped (if on a brand new topic, it would be offset 0).

```
$ bundle exec ruby test.rb
#<Kafka::FetchedMessage:0x007f8d7f109eb0 @message=#<Kafka::Protocol::Message:0x007f8d7f8ba3f8 @key=nil, @value="0.9944667197555783", @codec_id=0, @offset=0, @create_time=2018-03-31 11:41:28 +0200, @bytesize=18>, @topic="b", @partition=0>
(... waiting for pause timeout to expire...)
#<Kafka::FetchedMessage:0x007f8d80a8a9c0 @message=#<Kafka::Protocol::Message:0x007f8d7f110990 @key=nil, @value="0.9944667197555783", @codec_id=0, @offset=0, @create_time=2018-03-31 11:41:28 +0200, @bytesize=18>, @topic="b", @partition=0>
(... waiting for pause timeout to expire...)
#<Kafka::FetchedMessage:0x007f8d7f1225c8 @message=#<Kafka::Protocol::Message:0x007f8d7f1185c8 @key=nil, @value="0.9944667197555783", @codec_id=0, @offset=0, @create_time=2018-03-31 11:41:28 +0200, @bytesize=18>, @topic="b", @partition=0>
(... waiting for pause timeout to expire...)
```

[...]

PR #553, opened 2018-03-31
Stop processing batch of fetched messages if partition is paused

tmattia:
Attempt to fix https://github.com/zendesk/ruby-kafka/issues/552

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -212,6 +212,8 @@ module Kafka
             })
           end
           batch.messages.each do |message|
+            next if paused?(message.topic, message.partition)
+
             notification = {
               topic: message.topic,
               partition: message.partition,
@@ -229,7 +231,10 @@ module Kafka
             @instrumenter.instrument("process_message.consumer", notification) do
               begin
                 yield message
-                @current_offsets[message.topic][message.partition] = message.offset
+
+                unless paused?(message.topic, message.partition)
+                  @current_offsets[message.topic][message.partition] = message.offset
+                end
               rescue => e
                 location = "#{message.topic}/#{message.partition} at offset #{message.offset}"
                 backtrace = e.backtrace.join("\n")
--- a/spec/functional/consumer_spec.rb
+++ b/spec/functional/consumer_spec.rb
@@ -1,0 +1,49 @@
+describe "Consumer API", functional: true do
+  let!(:topic) { create_random_topic(num_partitions: 2) }
+  let(:consumer) { kafka.consumer(group_id: "consumer-name") }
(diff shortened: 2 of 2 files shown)


lib/kafka/consumer.rb
```diff
@@ -229,7 +231,10 @@ def each_message(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automatica
             @instrumenter.instrument("process_message.consumer", notification) do
               begin
                 yield message
-                @current_offsets[message.topic][message.partition] = message.offset
+
+                unless paused?(message.topic, message.partition)
+                  @current_offsets[message.topic][message.partition] = message.offset
+                end
```
benjaminoakes: I stepped through this issue as well and came to a very similar changeset.

lib/kafka/consumer.rb
```diff
@@ -212,6 +212,8 @@ def each_message(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automatica
             })
           end
           batch.messages.each do |message|
+            next if paused?(message.topic, message.partition)
```
dasch: This will effectively _skip_ all messages while a partition is paused, which is not the desired behavior.

dasch:
I would like to make sure this use case is strictly necessary before adding more logic to the already complex processing code.

tmattia:
Hey @dasch, thanks for taking the time to review this PR.

I'm not sure if I understand what you mean by skip.

From my understanding, the current behavior of the gem is unexpected because it continues to process messages from a partition even after the partition has been paused. The example given in the issue and the functional spec included in this PR reproduce the unexpected behavior.

According to the documentation for `Consumer#pause`, it should immediately stop processing messages from the paused partition. This is not currently happening because messages that have already been fetched (in a batch) are still processed even after the partition is paused.

The second issue with the current implementation is about where it resumes processing messages from a previously paused partition.

When `each_message` is called with `automatically_mark_as_processed: false`, it should resume from the last message that was not successfully marked as processed, allowing it a new chance at being successfully processed.

Does that make sense, or am I missing something? Is this against the behavior you would expect?

dasch:
No, I think you're right to flag the current behavior as wrong.

lib/kafka/consumer.rb
dasch: I think instead of doing this, we should break the loop for the entire batch. Let me take a look within the next few days.

dasch:
I want to merge #303 first, then I'll fix this issue.

tmattia:
@dasch thanks! Let me know if there's anything else you need, happy to help if you need more details or would like a different implementation.

PR #563, opened 2018-04-13
Add a statsd= method in the datadog integration

dmaze:
If an application has a configured statsd client from elsewhere for publishing Datadog stats, allows it to be used for Kafka stats too.


dasch:
Thanks!

PR #565, opened 2018-04-17
Ensure SSL is configured when SASL auth is used

### LINKED ISSUE #564: EOFError when reading response from server
* Version of Ruby: 2.4.3 (tested on 2.5.1 as well, same issue) coming from rbenv
* Version of Kafka: 0.11.0.2 (hosted on CloudKarafka. Port number is 9094)
* Version of ruby-kafka: 0.5.5 (tested on 0.6.0beta1, same issue)
* OS version: MacOS 10.13.4

#### Steps to reproduce

```ruby
kafka = Kafka.new(
  "ark.srvs.cloudkafka.com:9094",
  client_id: "kafka-example",
  sasl_scram_username: ENV['USERNAME'],
  sasl_scram_password: ENV['PASSWORD'],
  sasl_scram_mechanism: 'sha256'
)
kafka.deliver_message("Hello, World!", topic: "myexample-test")
```
#### Expected outcome

The message should be sent to the topic "myexample-test"

#### Actual outcome

An EOFError has been raised:
```
Failed to fetch metadata from kafka://ark.srvs.cloudkafka.com:9094: Connection error EOFError: EOFError
/Users/bdusauso/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/ruby-kafka-0.5.5/lib/kafka/cluster.rb:371:in `fetch_cluster_info': Could not connect to any of the seed brokers: (Kafka::ConnectionError)
- kafka://ark.srvs.cloudkafka.com:9094: Connection error EOFError: EOFError
```

##### Potential source of the problem

It seems the problem comes from an incorrect size being interpreted when the server response is sent back to the client.
The library is trying to read 352518912 bytes from the response, which is seems to be incorrect.
 
kafka-read-size

##### Using rdkafka

I've tried the same tiny example with `rdkafka`and it works perfectly.
Here's the snippet:

```ruby
require 'rdkafka'

config = {
          :"bootstrap.servers" => "ark.srvs.cloudkafka.com:9094" ,
          :"group.id" =>  "kafka-example",
          :"sasl.username" =>  ENV['USERNAME'],
          :"sasl.password"     => ENV['PASSWORD'],
          :"security.protocol" => "SASL_SSL",
          :"sasl.mechanisms"   => "SCRAM-SHA-256"
}

rdkafka = Rdkafka::Config.new(config)
producer = rdkafka.producer
producer.produce(topic: "myexample-test", payload: "Hello World!")
```

dasch:
Fixes #564.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -82,6 +82,10 @@ module Kafka
         logger: @logger
       )
 
+      if sasl_authenticator.enabled? && ssl_context.nil?
+        raise ArgumentError, "SASL authentication requires that SSL is configured"
+      end
+
       @connection_builder = ConnectionBuilder.new(
         client_id: client_id,
         connect_timeout: connect_timeout,
--- a/lib/kafka/sasl_authenticator.rb
+++ b/lib/kafka/sasl_authenticator.rb
@@ -28,21 +28,25 @@ module Kafka
         mechanism: sasl_scram_mechanism,
         logger: @logger,
       )
+
+      @mechanism = [@gssapi, @plain, @scram].find(&:configured?)
+    end
+
+    def enabled?
+      !@mechanism.nil?
     end
 
     def authenticate!(connection)
-      mechanism = [@gssapi, @plain, @scram].find(&:configured?)
+      return unless enabled?
 
-      return if mechanism.nil?
-
-      ident = mechanism.ident
+      ident = @mechanism.ident
       response = connection.send_request(Kafka::Protocol::SaslHandshakeRequest.new(ident))
 
       unless response.error_code == 0 && response.enabled_mechanisms.include?(ident)
         raise Kafka::Error, "#{ident} is not supported."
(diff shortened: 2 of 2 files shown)


novskey:
Is there any chance of this being reverted/some exception being possible? I was using SASL + plaintext  perfectly previously (0.5.5), and now this library is unusable for me.

dasch:
@novskey correct me if I'm wrong, but doesn't that mean you were transmitting credentials as plaintext over a non-encrypted connection?

novskey:
We’re using SASL with SCRAM SHA 256, and some of our use cases include the application being in the same private network as Kafka. In this case we’d rather avoid the performance reduction associated with SSL as credentials aren’t being transmitted publicly.

hallucinations:
@novskey, see whether #613 works for you, please!

novskey:
Thanks, will check tomorrow!

PR #566, opened 2018-04-17
Exponential backoff for pauses

dasch:
When enabled, exponential backoff will cause the pause timeout to double each time a partition is paused, until a message from the partition has been successfully processed, at which time the backoff is reset. It's possible to set a max timeout, above which we'll never pause a partition.

This will allow configuring a low initial timeout but still avoid overwhelming downstream services.

Example usage:

```ruby
begin
    ...
  end
rescue Kafka::ProcessingError => e
  # Pause for 1s initially, then double to 2s, 4s, etc., all the way up to 64s, but not higher.
  consumer.pause(e.topic, e.partition, timeout: 1, max_timeout: 64, exponential_backoff: true)
end
```

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -1,6 +1,7 @@
 require "kafka/consumer_group"
 require "kafka/offset_manager"
 require "kafka/fetcher"
+require "kafka/pause"
 
 module Kafka
 
@@ -50,8 +51,11 @@ module Kafka
       @fetcher = fetcher
       @heartbeat = heartbeat
 
-      # A list of partitions that have been paused, per topic.
-      @paused_partitions = {}
+      @pauses = Hash.new {|h, k|
+        h[k] = Hash.new {|h2, k2|
+          h2[k2] = Pause.new
+        }
+      }
 
       # Whether or not the consumer is currently consuming messages.
       @running = false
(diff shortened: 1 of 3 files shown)


mbibaud:
@dasch : Looks good. 
Logging wise I can only see we log when a partition resume: it'd be nice to have the pause logged along with the pause reason. 

Also wondering if it'd be useful to have the pause/resume register a Datadog event.

dasch:
@mbibaud currently that logging is done in Racecar, but it's a good point that it should be done in ruby-kafka instead/also.

As for monitoring – my thinking was that it would become unwieldy to use events for this. What we really want to monitor is the _duration_ of pauses, i.e. allow short pauses but alert on longer ones. For that I think I'd need to use a _gauge_ that gets emitted periodically with the duration of each current pause, and maybe zero for non-paused partitions. I just worry that it's a lot of overhead to add – but I guess it could only be done if pausing is even enabled.

lib/kafka/consumer.rb
```diff
@@ -449,9 +463,9 @@ def seek_to_next(topic, partition)
     end
 
     def resume_paused_partitions!
-      @paused_partitions.each do |topic, partitions|
-        partitions.keys.each do |partition|
-          unless paused?(topic, partition)
+      @pauses.each do |topic, partitions|
+        partitions.each do |partition, pause|
+          if pause.paused? && pause.expired?
```
mensfeld: this is a bug. Should be if !pause.paused? || pause.expired?

dasch: I don't think it is – basically, it's `if partition has been paused but pause has expired`.

PR #571, opened 2018-04-23
Instrument the fetcher queue size

dasch:


dasch:
I'd be open to a PR that does the `# frozen_string_literal: true` trick in every file.

mensfeld:
@dasch no problem. As always give me 1-2 weeks ;)

tjwp:
Took on adding the comment to all files here: https://github.com/zendesk/ruby-kafka/pull/577

PR #575, opened 2018-04-26
Don't continue processing messages when a partition is paused

dasch:
If a partition was paused from within the processing loop, the consumer would continue to process the rest of the message batch.

Simplified version of #553.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -202,6 +202,10 @@ module Kafka
 
         batches.each do |batch|
           batch.messages.each do |message|
+            # Break out of the loop if the partition has been paused, e.g. when
+            # processing the previous message in the batch.
+            break if paused?(batch.topic, batch.partition)
+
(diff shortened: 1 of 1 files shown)


lib/kafka/consumer.rb
```diff
@@ -237,9 +241,11 @@ def each_message(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automatica
...
           end
 
-          # We've successfully processed a batch from the partition, so we can clear
-          # the pause.
-          pause_for(batch.topic, batch.partition).reset!
+          unless paused?(batch.topic, batch.partition)
+            # We've successfully processed a batch from the partition, so we can clear
+            # the pause state.
+            pause_for(batch.topic, batch.partition).reset!
+          end
```
dasch: We might get to this line even if we _haven't_ processed any messages in the batch since the pause, hence the `unless`.

PR #583, opened 2018-05-07
Add describe groups API

jasondoc3:
Should merge after https://github.com/zendesk/ruby-kafka/pull/582

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -255,6 +255,37 @@ module Kafka
       nil
     end
 
+    def describe_groups(group_ids:)
+      response = random_broker.describe_groups(group_ids: group_ids)
+      response.groups.each { |group| Protocol.handle_error(group.error_code) }
+
+      response.groups.each_with_object({}) do |group, hash|
+        group_id = group.group_id
+        hash[group_id] = {}
+        hash[group_id]['state'] = group.state
+        hash[group_id]['protocol'] = group.protocol
+        hash[group_id]['members'] = []
+
+        group.members.each_with_object(hash[group_id]) do |member, group_hash|
+          member_hash = {}
+
+          member_hash['member_id'] = member.member_id
+          member_hash['client_host'] = member.client_host
+          member_hash['client_id'] = member.client_id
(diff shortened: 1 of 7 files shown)


lib/kafka/cluster.rb
```diff
@@ -255,6 +255,37 @@ def alter_topic(name, configs = {})
...
+          topic_hash = {}
+          topic_hash['name'] = topic
+          topic_hash['partitions'] = partitions
+          member_hash['topics'] << topic_hash
+        end
+
+        group_hash['members'] << member_hash
+      end
+
+      group_hash
```
dasch: I'd rather we return the actual objects, rather than a Hash.

jasondoc3: @dasch Updated to return the actual object

lib/kafka/protocol/describe_groups_response.rb
```diff
@@ -0,0 +1,73 @@
...
+          group_id = decoder.string
+          state = decoder.string
+          protocol_type = decoder.string
+          protocol = decoder.string
+
+          members = decoder.array do
+            member_id = decoder.string
+            client_id = decoder.string
+            client_host = decoder.string
+            metadata = decoder.bytes
```
jasondoc3: This might contain some useful information, but I couldn't find how to decode it in the [protocol docs](https://kafka.apache.org/protocol) (Unless I missed something)

> The metadata corresponding to the current group protocol in use (will only be present if the group is stable).

nguyenquangminh0711: @jasondoc3 The group protocol is the metadata information that the group agree on at the Join Group step. In that step, each consumer group member proposes a list of protocol and its metadata that all members in the group could agree on. The group leader picks a protocol that all members in the group supports and store that information. Therefore, this field is defined by the client. You don't need to parse this. `ruby-kafka` gem doesn't support a custom group protocol. 

P/s: If you are curious, you can look at `lib/kafka/protocol/join_group_request.rb` and `lib/kafka/protocol/consumer_group_protocol.rb`

PR #591, opened 2018-05-29
Fix degrading performance problem

nguyenquangminh0711:
This PR is to fix the problem mentioned in https://github.com/zendesk/ruby-kafka/pull/548. New record batch format meets a serious problem with the performance. The client fires tons of fetch requests to Kafka, instead of working by batch. The main cause of this issue: FetchResponse doesn't handle the case that multiple record batches are tied together. This makes the client not aware of the number of record batches. In that case, we'll need to read to EOF and decode each record batch next to each.

In addition, there is a case which is not handled properly. Each record batch can contain in-completed series of bytes due to new max byte option in Fetch API v4. Those byte sequences doesn't provide any value and we just need to simply ignore them.

I had some benchmarks, to test out *raw fetching speed*:

- Before the fix:

```
Kafka version: 0.7.0.alpha2
 - 1000000 messages, 20 bytes each message
=> 152.63421600003494 seconds
 - 1000000 messages, 100 bytes each message
=> 244.34417100000428 seconds
 - 1000000 messages, 250 bytes each message
=> 300.691599996574 seconds
 - 1000000 messages, 500 bytes each message
=> 321.691599996574
```

- After the fix:

```
Kafka version: 0.7.0.alpha2
 - 1000000 messages, 20 bytes each message
=> 52.63421600003494 seconds
 - 1000000 messages, 100 bytes each message
=> 44.34417100000428 seconds
 - 1000000 messages, 250 bytes each message
=> 43.64991599996574 seconds
 - 1000000 messages, 500 bytes each message
=> 43.933405000017956 seconds
```

Comparing to other versions:

```
Kafka version: 0.6.0.beta2
 - 1000000 messages, 20 bytes each message
=> 43.417863000009675 seconds
 - 1000000 messages, 100 bytes each message
=> 41.23134899995057 seconds
 - 1000000 messages, 250 bytes each message
=> 38.98760399996536 seconds
 - 1000000 messages, 500 bytes each message
=> 40.007947999984026 seconds

Kafka version: 0.5.5
 - 1000000 messages, 20 bytes each message
=> 39.97198299999582 seconds
 - 1000000 messages, 100 bytes each message
=> 37.00100899999961 seconds
 - 1000000 messages, 250 bytes each message
=> 36.51904599997215 seconds
 - 1000000 messages, 500 bytes each message
=> 37.17629799997667 seconds
```

This is just raw fetching speed, which indicates the overhead of the fetching operation. The speed of new record batch is slightly slower than the previous version. In real case scenarios, there are other more important factors, which makes this overhead not significant to the overall consuming speed. But I'll try to optimize more.

### DIFF
--- a/lib/kafka/protocol/fetch_response.rb
+++ b/lib/kafka/protocol/fetch_response.rb
@@ -74,11 +74,21 @@ module Kafka
 
             messages_decoder = Decoder.from_string(decoder.bytes)
             messages = []
-            magic_byte = messages_decoder.peek(MAGIC_BYTE_OFFSET, MAGIC_BYTE_LENGTH)[0].to_i
 
+            magic_byte = messages_decoder.peek(MAGIC_BYTE_OFFSET, MAGIC_BYTE_LENGTH)[0].to_i
             if magic_byte == RecordBatch::MAGIC_BYTE
-              record_batch = RecordBatch.decode(messages_decoder)
-              messages = record_batch.records
+              until messages_decoder.eof?
+                begin
+                  record_batch = RecordBatch.decode(messages_decoder)
+                  messages += record_batch.records
+                rescue InsufficientDataMessage
+                  if messages.length > 0
+                    break
+                  else
+                    raise
+                  end
+                end
+              end
             else
               message_set = MessageSet.decode(messages_decoder)
               messages = message_set.messages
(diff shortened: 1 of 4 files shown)


nguyenquangminh0711:
This fix your problem @Elyahou.

dasch:
@Elyahou @nguyenquangminh0711 any updates on the performance of the latest version?

Elyahou:
@dasch Sorry I had no time to check this since this PR. I will check it later this week

PR #593, opened 2018-06-02
More consistent heartbeat API with non-optional trigger

mensfeld:
This PR provides following functionalities:
- Allows to use both "forceful" and "optional" version of the heartbeat trigger
- Provides that functionality on a consumer level via `trigger_heartbeat` and `trigger_heartbeat!`
- Fixes internal heartbeat usage so all the usages are via the heartbeat method. In case of further refactor, less places to change are needed

The bang/non-bang naming convention is based on the proposed improvement for racecar (https://github.com/zendesk/racecar/issues/61).

The naming was changed not to conflict with Ruby core method`#send`.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -235,7 +235,7 @@ module Kafka
             mark_message_as_processed(message) if automatically_mark_as_processed
             @offset_manager.commit_offsets_if_necessary
 
-            @heartbeat.send_if_necessary
+            @heartbeat.trigger
 
             return if !@running
           end
@@ -330,7 +330,7 @@ module Kafka
 
           @offset_manager.commit_offsets_if_necessary
 
-          @heartbeat.send_if_necessary
+          @heartbeat.trigger
 
           return if !@running
         end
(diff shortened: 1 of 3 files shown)


lib/kafka/consumer.rb
```diff
@@ -366,8 +366,12 @@ def mark_message_as_processed(message)
       @offset_manager.mark_as_processed(message.topic, message.partition, message.offset)
     end
 
-    def send_heartbeat_if_necessary
-      @heartbeat.send_if_necessary
+    def trigger_heartbeat
```
dasch: Since this method has been part of the public API (Consumer is public) I don't want to change its name. We should either keep the old name as an alias or stick to the `send_*` naming at this layer.

mensfeld: You are absolutely right. I will add proper public api references and we can just deprecate that at some point.

mensfeld: @dasch added the compatibility layer.

mensfeld:
Please don't forget to backport to 0.6 if possible since 0.7 is not yet beyond alpha. This will help us fix some problems.

nijikon:
@mensfeld @dasch thanks ❤️

PR #594, opened 2018-06-04
Skip messages that have already been processed

dasch:


marionzualo:
@dasch Does `each_batch` also need to be changed?

lib/kafka/consumer.rb
```diff
@@ -204,7 +204,14 @@ def each_message(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automatica
         batches = fetch_batches
 
         batches.each do |batch|
+          committed_offset = @offset_manager.next_offset_for(batch.topic, batch.partition)
```
marionzualo: I'm not super familiar with the codebase so pardon for any incorrect assumptions.

When I looked at the logs of this behaviour in production I noticed that the worker starts by seeking to the already committed offset. I wonder if there would be a way to seek to the desired commit offset instead of skipping the ones that we already know have been committed within the consumer loop? 

The logs I saw seem to have come from `Fetcher#handle_seek` which was invoked through `Consumer#seek_to_next`. `seek_to_next` seems to already be doing a similar thing as done here (maybe I'm missing something):
https://github.com/zendesk/ruby-kafka/blob/589b3d73386a7d4a818ca6df9844f7019a6387f9/lib/kafka/consumer.rb#L460-L470

dasch:
This PR has been superceded by #595.

Issue #611, opened 2018-07-03
Consumer seeks offsets from previously assigned partitions


stsc3000:
* Version of Ruby: 2.5.0p0
* Version of Kafka: 0.10.1.0
* Version of ruby-kafka: 0.6.7/master

Hi, I believe we have discovered an issue very similar to the one described in issue #347:

Upon rebalancing within our system, a consumer got assigned a partition that it was assigned previously.
When beginning to consume messages, it then used the offset it previously kept in memory for
the partition to seek, resulting in a lot of duplicate messages being processed.

I believe I have found a way to reproduce the behavior:

### 1. Run Kafka in Docker

`$ docker run --rm --env ADVERTISED_HOST=localhost -p 2181:2181 -p 9092:9092 --env ADVERTISED_PORT=9092 --name kafka -h kafka spotify/kafka`

### 2. Create topic with two partitions

`$ docker exec -it CONTAINER_ID /opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh --create --zookeeper localhost:2181 --topic debug-topic --replication-factor 1 --partitions 2`

### 3. Run Consumer 1 (consumer.rb)
* Subscribes to partitions 0 and 1
* Keep running

### 4. Run Producer (producer.rb)
* Consumer 1 receives all messages, sets local offsets

### 5. Run Consumer 2 (consumer.rb, different shell)
* Wait for rebalance
* Each consumer get assigned one partition

### 6. Run Producer (producer.rb)
* Each consumer will receive messages

### 7. Stop Consumer 2
* Wait for rebalance

Assumption: Stopping Consumer 2 does not cause any further processing (as all messages have been already)
Actual Result: Stopping Consumer 2 causes Consumer 1 to process messages that were processed by Consumer 2 already

`consumer.rb`

```ruby
#!/usr/bin/env ruby

require 'ruby-kafka'

TOPIC = 'debug-topic'

logger = Logger.new(STDOUT)
logger.level = Logger::INFO

kafka = Kafka.new(
  seed_brokers: ['localhost:9092'],
  logger: logger
)

consumer = kafka.consumer(group_id: 'consumer-group', offset_commit_threshold: 1)

consumer.subscribe(TOPIC)

consumer.each_message do |msg|
  puts ({ partition: msg.partition, offset: msg.offset }).inspect
end
```

`producer.rb`

```ruby
#!/usr/bin/env ruby

require 'ruby-kafka'

TOPIC = 'debug-topic'

[...]

PR #612, opened 2018-07-03
Clear current offsets on join group

stsc3000:
As is mentioned in a comment in Consumer#join_group, we need to make sure
that a consumer clears its stored offsets once it is no longer subscribed to a partition.

This is managed in the offset manager, however the consumer also keeps the offset
in `@current_offsets`, which is used for seeking when joining a group.

This may lead to a consumer seeking to an offset that is actually already processed once re-subscribing to a partition that it was subscribed to previously.

This attempt at a solution clears `@current_offsets` exactly as is done in the offset manager.

Note: I had a hard time testing the behaviour. My unit test (not included) was quite complex as I did not know how to succinctly tease joining and rejoining a group out of the consumer.
I opted to create a functional spec, however I'd be interested in whether it is possible to not reach into the instances `@group` to check whether it is subscribed to a partition. Also the functional test includes _a lot_ of logging, which is meant to illustrate the behavior and will be removed if this gets accepted.

As we have noticed the issue #611 in our production environment, I'd be interested in whether our analysis of the issue is correct.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -444,6 +444,7 @@ module Kafka
       if old_generation_id && @group.generation_id != old_generation_id + 1
         # We've been out of the group for at least an entire generation, no
         # sense in trying to hold on to offset data
+        clear_current_offsets
         @offset_manager.clear_offsets
       else
         # After rejoining the group we may have been assigned a new set of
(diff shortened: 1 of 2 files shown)


stsc3000:
@dasch I removed the functional spec and added a cleaner version of my previous unit test. The issue of reaching into the instance variable to check for the state of `@current_offsets` is still present, however I feel it is less problematic there as it was also done before. Triggering `join_group` itself is also a bit contrived as you'll notice.

I can remove the spec as well, just felt like I should offer an alternative to the functional spec.

dasch:
Thanks!

PR #614, opened 2018-07-04
Update CHANGELOG with missing unreleased items

hallucinations:


hallucinations:
Not sure why this would break specs. cc: @dasch

PR #616, opened 2018-07-05
Add more instrumentation to allow for more robust lag reporting

dorner:
Right now, RubyKafka exposes the lag metric only after consuming a message. There are cases where we want to see the current lag which is not currently supported:
* The consumer does not start up correctly
* The consumer starts up but is stuck on a message (e.g. it's taking too long to process and session timeout continually gets missed)
* The consumer starts up but is failing somewhere before actually getting the message

This PR adds two more places where we can instrument:
* When the consumer first starts up; this will report the current offset so the initial lag can be computed;
* Whenever a heartbeat happens, so we can report if the lag is still where it is (stuck) or has gone up. I know that in the current implementation heartbeat is tied to consuming, but I want to revisit the implementation discussed in #355 to move this to a thread. We're already using threads in our own implementation, so we can have heartbeats separate from the poll loop.

Another advantage of instrumenting heartbeat is that we can throttle the messages we're sending to e.g. DataDog by sending them only once a heartbeat instead of once every message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -772,7 +772,7 @@ All notifications have `group_id` in the payload, referring to the Kafka consume
 * `leave_group.consumer.kafka` is sent whenever a consumer leaves a consumer group. It includes the following payload:
   * `group_id` is the consumer group id.
   
-* `set_offset.consumer.kafka` is sent when a consumer first seeks to an offset. It includes the following payload:
+* `seek.consumer.kafka` is sent when a consumer first seeks to an offset. It includes the following payload:
   * `group_id` is the consumer group id.
   * `topic` is the topic we are seeking in.
   * `partition` is the partition we are seeking in.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -771,7 +771,16 @@ All notifications have `group_id` in the payload, referring to the Kafka consume
   
 * `leave_group.consumer.kafka` is sent whenever a consumer leaves a consumer group. It includes the following payload:
   * `group_id` is the consumer group id.
- 
+  
+* `set_offset.consumer.kafka` is sent when a consumer first seeks to an offset. It includes the following payload:
```
dasch: How about we instrument all seeks instead? And then use `seek.consumer.kafka`.

dorner:
Congratulations! Best excuse ever! :D and yep, I can make that change.

dorner:
@dasch added another commit, please take a look.

README.md
dasch: You just need to update the README here, i.e. `s/set_offset/seek/`.

dorner:
@dasch done!

dorner:
@dasch any update on this?

dasch:
:heart:

Issue #618, opened 2018-07-19
Use a thread for sending heartbeats


dorner:
This is to track a functionality change whereby instead of sending heartbeats inline with message processing, they are sent via a separate thread.

This introduces the `poll_timeout` configuration variable, which will force the consumer to time out if it is reached and a new message has not been received within that time.

PR is imminent!

theturtle32:
Yeah, the existing inline behavior has bitten us in the ass many, many, many times! I'll be very grateful to see this moved to a thread. I'm glad you're already working on this, as I was thinking of taking a stab at a PR myself.

dorner:
@theturtle32 Well, not sure when I'll be able to get around to rewriting it entirely :) If you're willing to take a stab, please do!

PR #619, opened 2018-07-19
#618: Move heartbeat processing to a separate thread

dorner:
See issue #618. Comments and suggestions are welcome - please tear this apart! 😬

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changes and additions to the library will be listed here.
 - Upgrade and rename GroupCoordinator API to FindCoordinator API (#606).
 - Refresh cluster metadata after topic re-assignment (#609).
 - Disable SASL over SSL with a new config (#613).
+- Move heartbeat to a separate thread and add poll_timeout configuration.
 
 ## 0.6.7
 
--- a/README.md
+++ b/README.md
@@ -872,6 +872,21 @@ When sending many messages, it's likely that the client needs to send some messa
 
 Make sure your application can survive being blocked for so long.
 
+**Consumer timeouts** can be configured when calling `#consumer` on a client instance:
+
+* `session_timeout` - once this value is reached, the broker will remove the consumer from the list of connected
+  consumers. The timeout is reached if there are no successful poll ("get more messages") or heartbeat requests.
+* `heartbeat_interval` - this is the time between when heartbeats are sent.
+   Ideally this should be set to about 1/3 your session timeout so that if
+   a heartbeat fails, you have enough time to retry and re-connect before
+   the consumer is removed.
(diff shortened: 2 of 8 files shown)


dasch:
I'd like to get this functionality, but I want to avoid using a mutex to achieve it. The problem of course is that the heartbeats need to be sent on the same socket as the other group API calls, so we do need to synchronize access to that socket.

One way to go about this would be to run the consumer group client code in a wholly separate thread and use a thread-safe queue to communicate with the processing code. That's one reason I've been hesitant to add support for background heartbeats – doing it properly is hard.

dasch:
Also note that the CI tests are failing.

dorner:
@dasch thanks for the feedback. Can you explain more why you don't think a mutex will work? We've had this in production for about 2 weeks now and have not seen any problems.
I can fix the tests - silly last minute update which I did without checking.

dasch:
@dorner my worry is about long-term maintenance – debugging deadlocks and other mutex-related issues is very hard, so I'm wary about introducing such concepts. If we were to do it, I'd prefer the mutex live in `Connection` if that could be done in a performant way. Sprinkling locks throughout the code base is a recipe for future disaster.

mensfeld:
I agree with @dasch. Ruby Kafka already has concurrency and more than one thread. Introducing new multi thread dependencies will be a nightmare during debugging and error finding. I still recall the separate thread prefetch memory bloats that we had ;)

dorner:
@dasch OK, makes sense. From your previous comment - moving all consumer group functionality to a separate thread - do you have a high level description of how that might be implemented? I can take the time to try and do the work on my end. Would it be similar to how Fetcher is done right now?

Also, OTOH would that include heartbeat, join_group and synchronize? Anything else?

dasch:
@dorner it's a rather large and risky change, and would probably involve multiple rounds of explorative implementation.

A simpler approach might be to start by moving the _processing_ to another thread, which would leave the main thread responsible for the group stuff. A `Queue` could be used to send messages to the processor thread, and another one could be used to send back acknowledgements or errors. It may or may not work well in practice.

dorner:
@dasch well, I'm willing to take whatever approach you feel is best. That one sounds even messier to me though - we'd have to move from a consumer loop to something more of a cross-thread communication (since we can't move on to the next message until the current message is finished processing). Unless I'm misunderstanding your suggestion?

dasch:
Nah, that's exactly right. But finding the right design here is messy – there are a lot of moving pieces.

dasch:
I just got a feeling that the number of interactions in the group code will
make it harder, but feel free to give it a go!

Note though that I’ll only merge the resulting code if it’s sufficiently
maintainable - I personally implemented an async IO system for ruby-kafka
but ended up reverting it because it proved too unstable and difficult to
debug...

Issue #621, opened 2018-07-30
ActiveSupport < 4.0 not supported?


SebastianCoetzee:
I have a Rails app that is running v3.2.22.9 of Rails and v2.3.6 of Ruby and it fails to boot when version 0.6.7 of `ruby-kafka` gem is included in the Gemfile. The following is the stack trace:

```
LoadError: cannot load such file -- active_support/subscriber
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/backports-3.11.1/lib/backports/std_lib.rb:9:in `require'
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/backports-3.11.1/lib/backports/std_lib.rb:9:in `require_with_backports'
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/activesupport-3.2.22.9/lib/active_support/dependencies.rb:251:in `block in require'
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/activesupport-3.2.22.9/lib/active_support/dependencies.rb:236:in `load_dependency'
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/activesupport-3.2.22.9/lib/active_support/dependencies.rb:251:in `require'
/home/travis/build/yyy/yyy/vendor/bundle/ruby/2.3.0/gems/ruby-kafka-0.6.7/lib/kafka/datadog.rb:10:in `<top (required)>'
```

The reason for this is because `active_support/subscriber` is not available in <4.0 of ActiveSupport. Is this dependency restriction by design or may I submit a PR to include the functionality of this class within the `ruby-kafka` gem, thereby allowing the gem to be compatible with ActiveSupport <4.0?

nguyenquangminh0711:
The datalog plugin uses active support subscriber internally to perform its functionalities, so I'm afraid we can't change this dependency, at least in near future. By default, ruby-kafka doesn't load datadog and `active_support/subscriber`, so the normal users won't have any problem with this. 

Since we can't do anything about this, I think I'll close this issue, until we come up with a better solution.

PR #622, opened 2018-07-30
Restrict ActiveSupport to >= 4.0

SebastianCoetzee:
This PR references this issue: https://github.com/zendesk/ruby-kafka/issues/621

Because the the `Datadog` and `Statsd` subscribers both use `active_support/subscriber`, the gem depends on a version of ActiveSupport `>= 4.0`. This PR adds this dependency to the gemspec.

### DIFF
--- a/ruby-kafka.gemspec
+++ b/ruby-kafka.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'digest-crc'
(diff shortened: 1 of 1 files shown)


SebastianCoetzee:
For whatever reason, the test failed. I suspect this is an intermittent failure, but I don't have write permissions to trigger builds.

dasch:
Currently, the dependency on Active Support is a soft one – if it's available we'll load it, but it's not required to run ruby-kafka. As such, I'm not sure there's a good way to achieve this.

Issue #624, opened 2018-08-01
Messages are being skipped


bradurani:
* Version of Ruby: 2.5
* Version of Kafka: 0.11.0.2
* Version of ruby-kafka: 0.6.6

We're seeing behavior that only appears under production load that is causing messages to be skipped. Here's what we're seeing:

![image](https://user-images.githubusercontent.com/4195952/43544917-45a3197c-9589-11e8-8bc9-c5869b366a45.png)

Here you can see `consumer.message` for 2 consumer groups. As you can see the group A, represented by the top line, is consuming more message than group B, represented by the bottom line. This is despite the fact that they're consuming the same topic. Since one is clearly outpacing the other, you would expect to see lag in group B, however:

![image](https://user-images.githubusercontent.com/4195952/43545100-c8d72ab8-9589-11e8-8244-4e3375abc150.png)

Lag is actually ver small, fluctuating by only a few hundred messages per partition. At the same time, group B is consuming hundreds of thousands of message less every hour. For the same hour, here's
`producer.produce`
`consumer.messages` for A
`consumer.messages` for B
Sum of max `consumer.lag` for all partitions for B
![image](https://user-images.githubusercontent.com/4195952/43546235-f56c6f72-958c-11e8-8f61-05ab225e0c67.png)

As you can see there's a gap of 1324688 - 1088320 - 5616 = 230752 messages

Indeed, when we look at the destination db the records are being written to, the gap exists. We record partition and offset in the destination database, and there are gaps of hundreds of offsets consistent with the above graphs. Not only that, group A, which is the healthier group, also has gaps, though not as many. 
Both groups rarely fail to write to the destination, and when they do, we retry until they succeed before proceeding with the next batch.

I don't understand how we could have huge discrepancies in `consumer.messages` for 2 groups consuming the same topic, without seeing commensurate lag.

I haven't been able to reproduce this on my dev env. One thing I have noticed is that in production, both groups are logging `WARN -- : Reached max fetcher queue size (100), sleeping 1s`, and group B is showing this a lot more. 

Configurations I've tried:
- `automatically_mark_as_processed: true` with commit offsets every 10 seconds
- `automatically_mark_as_processed: false` with `mark_as_processed` and `commit_offsets` after every batch

Any help would be appreciated

bradurani:
Update: tried with ruby-kafka: `0.6.7` and issue persists

mensfeld:
@bradurani thanks for reporting.

[...]

Issue #555, opened 2018-04-03
Consumer get stucked in compacted topic


Elyahou:
When replaying a compacted topic, with 20 partitions, the consumer get stucked in 5 partitions at some offsets, and don't succeed to read any more messages. All the other partitions were successully replayed...

I succeeded to fix this by manually increasing the consumer group offset for the problematic partitions, after I saw that it was stucked in an offset that was compacted and there was no messages for it in kafka.

Example:

offset | status | remark
---|---|---
1| read| message present
2| read| message present
3| read| message present
4| stucked here| deleted by kafka compaction
5| |deleted by kafka compaction
6| |deleted by kafka compaction
7|| message present

So it works only after I moved the offset of the consumer group to 7.

Some remarks:
- There is other offsets that was compacted by kafka and there was no issue there
- The kafka console consumer succeeded to consume these messages continuously without any issue
- I tried all kind of configuration changes on `socket_timeout`, `connect_timeout` and `max_wait_time`., but it did not help.

* Version of Ruby: 2.3.5
* Version of Kafka: 0.11.0.2
* Version of ruby-kafka: 0.5.4

###### Code of the consumer

```ruby
require "kafka"

kafka = Kafka.new(["kafka1:9092", "kafka2:9092"])
consumer = kafka.consumer(group_id: "my-consumer")
consumer.subscribe("greetings")
trap("TERM") { consumer.stop }

consumer.each_message do |message|
  puts message.topic, message.partition
  puts message.offset, message.key, message.value
end
```

dasch:
Good job reproducing the bug – it sounds like it's related to how fetches work when asking for messages at a starting offset that no longer exists. It used to be that the Kafka would just start at the next existing offset, but that may have changed. Would you be willing to dive into the Java client source and see how they handle this?

Elyahou:
I am not sure it is the issue, because in the case I reset the consumer to offset 2 in the precedent example, the consumer will receive only message 2 and 3, and will not receive further messages... So it is not only an issue when the fetch operation is on an empty offset

dasch:
It could be a problem with the next offset computation – that's rather complicated.

dasch:
Would you be willing to `bundle open ruby-kafka` and sprinkle debug statements into the code?

Elyahou:
@dasch Yes of course I can

Elyahou:
I continued my analysis of this issue, I think I maybe understood it...

[...]

PR #625, opened 2018-08-07
Fix Fetcher's message skipping 

nguyenquangminh0711:
In the current Fetcher and Consumer, the current implementation has a bug that leads to the situation where a huge amount of messages are skipped, without being processed. The root cause is that the fetcher doesn't handle the case where a fetched batch is empty. 

In normal scenario, it updates the current offset based on the fetched batch last offset. When the batch is empty, it updates the current offset to the highwater mark offset, which is the end of the queue. The empty batch didn't exist before. However, newer versions of Kafka (after 0.11.0, I guess) do return the empty batches in some cases:
- Control batches of transactional message production
- Throttled fetch request
- Incomplete message sets which exceed the max byte limit
- etc.
My fix is just not to update current offset when the batch is empty. In case there is nothing more to fetch, the consumers just don't update the offset, instead of using the highwater mark offset like current.

Fix https://github.com/zendesk/ruby-kafka/issues/624
Merge merge before https://github.com/zendesk/ruby-kafka/pull/608

Reference: two cases that return empty sets. There would be more cases.
screen shot 2018-08-07 at 14 24 59

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -477,7 +477,7 @@ module Kafka
 
         batches.each do |batch|
           batch.messages.each(&block)
-          unless batch.messages.empty?
+          unless batch.empty?
             offsets[batch.partition] = batch.last_offset + 1
           end
         end
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -477,7 +477,9 @@ module Kafka
 
         batches.each do |batch|
           batch.messages.each(&block)
-          offsets[batch.partition] = batch.last_offset + 1
+          unless batch.messages.empty?
+            offsets[batch.partition] = batch.last_offset + 1
+          end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


Elyahou:
@nguyenquangminh0711 Can this issue be related to https://github.com/zendesk/ruby-kafka/issues/555 ?

mensfeld:
@nguyenquangminh0711 @dasch can we bcakport it to `0.6`?

lib/kafka/client.rb
```diff
@@ -477,7 +477,9 @@ def each_message(topic:, start_from_beginning: true, max_wait_time: 5, min_bytes
 
         batches.each do |batch|
           batch.messages.each(&block)
-          offsets[batch.partition] = batch.last_offset + 1
+          unless batch.messages.empty?
```
dasch: You should be able to do `batch.empty?`.

nguyenquangminh0711:
@dasch Updated 💪

dasch:
:heart:

mainameiz:
How about to add specs to test such edge cases, guys?

PR #627, opened 2018-08-11
move sleep values to constant

nijikon:
As a first step of making sleep values configurable move them to constants.

### DIFF
--- a/lib/kafka.rb
+++ b/lib/kafka.rb
@@ -3,6 +3,23 @@
 require "kafka/version"
 
 module Kafka
+  # Set of default backoff values
+  DEFAULT_BACKOFFS = {
+    connection_error: 1,
+    coordinator_not_available: 1,
+    leader_not_available: 1,
+    loop_max_reached: 1,
+    loop_not_running: 0.1,
+    message_delivery_error: 1,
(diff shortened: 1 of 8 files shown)


nijikon:
@dasch we can change the key names to whatever you think is better

dasch:
I'd rather move the constant to `Kafka::DEFAULT_BACKOFFS`. But maybe include the full change in this PR?

nijikon:
@dasch we can move it to `Kafka::DEFAULT_BACKOFFS`, that's even better. I would rather create small changes instead of creating a big PR which is hard to review, and if for whatever reason we would have to revert, we would revert things that did not break anything. So it's better to do it in small chunks, at least that's how I do it. Will that work?

dasch:
I prefer full-featured PRs over a ton of small ones. I'm not interested in half the changes – this only makes sense if we go all the way. But extra points for having nice commits that I can go through one at a time!

nijikon:
I first wanted to clean it up before creating the actual feature, but we can do it in one PR I guess.

nguyenquangminh0711:
Hi @nijikon, it has been a while. The PR is having some conflicts with the master. Do you have any update on this?

PR #630, opened 2018-08-17
Update README.md for max_queue_size key change

jgeiger:
The key is not max_buffer_size but rather max_queue_size. This reflects the change in the README.md file.


dasch:
The queue size only applies to the _async_ producer – there's a separate configuration for a buffer size that applies to both producer types.

dasch:
https://www.rubydoc.info/gems/ruby-kafka/Kafka/Client#producer-instance_method

PR #634, opened 2018-08-24
Add compatibility with dogstatsd-ruby v4.0.0

tjwp:
dogstatsd-ruby v4.0.0 moved the constants for connection defaults into a new class: https://github.com/DataDog/dogstatsd-ruby/commit/055cda22c6f029a411e86292349a73b4b53a05f2

This change keeps this gem compatible with both the old and new versions of dogstatsd-ruby.

Given that the default values represented here probably are truly constants ('127.0.0.1', 8125), it is tempting to just replicate them here. If you'd prefer that approach let me know, and I'll update this PR.

We're still using v0.6.8, so if this change is accepted I'll also backport it to the v0.6.x branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes and additions to the library will be listed here.
 
 ## Unreleased
 
+- Compatibility with dogstatsd-ruby v4.0.0.
+
 ## 0.7.1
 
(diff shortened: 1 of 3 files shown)


dasch:
Thanks! :heart:

And sorry for the late review.

PR #637, opened 2018-08-30
Pass batch to each_message as a second optional argument

mensfeld:
I have a particular case, where I need to track `highwater_mark_offset` and `first_offset` for batches consumed with `each_message`. This change will allow me to reach this data easily by reaching to the second (optional) argument of the `#each_message` method.

I also believe that it will also allow for unification of the API, so the same information is available using both `each_batch` and `each_message`.

This change does not break the current API nor does it require any changes from anyone (second argument when not used will be ignored).

It will also allow me to unify the API for Karafka `Metadata` object that I want to provide for both ways of processing messages.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -223,7 +223,7 @@ module Kafka
 
             @instrumenter.instrument("process_message.consumer", notification) do
               begin
-                yield message unless message.is_control_record
+                yield(message, batch) unless message.is_control_record
                 @current_offsets[message.topic][message.partition] = message.offset
               rescue => e
                 location = "#{message.topic}/#{message.partition} at offset #{message.offset}"


mensfeld:
Spec failing for 1.1 is irrelevant to this PR and looks random.

klippx:
:thinking: Tbh, it feels quite natural that each_message yields each message, and each_batch yields each batch. 

Proposing that each_message should also yield the batch itself seems like it would solve your problem in a convenient way, but I for one do not see how "it allows unification of the API". It just makes the API _weird_ imo. 

I think it would be good if you link to an issue in Karafka describing your use case.

mensfeld:
@klippx I don't agree with you on the api change as it is seamless but I found a way to achieve without introducing any changes to ruby kafka. **However** I still think, that not being able to fetch the batch metadata when using `each_message` should be somehow fixed.

PR #640, opened 2018-09-04
change the log level of the 'Committing offsets'  message to debug

derencius:
change the log level of the 'Committing offsets' message to debug. other low level messages have log_level debug. 

in a app that I'm working, I use ruby-kafka with a log level set to info, and after a message being received it keeps logging this message.


dasch:
Those logs are a key tool to understand incidents, though – without them, it's really difficult to figure out what went wrong when.

derencius:
@dasch yeah, makes more sense to keep it as it is.  I can change the log level in development to warn, so this log would not generate noise.

Shadowssong:
@dasch I understand your point but I actually ran into this separately than @derencius.  I'm processing a lot of messages and my logs are filling up with offset commit messages, a lot of other stuff is useful at info level that isn't as "spammy" as this one is.  Could we alternatively make it configurable to a different log level so that it doesn't affect everyone by default?

dasch:
OK then. @derencius can you rebase on master?

PR #641, opened 2018-09-11
Fresh cluster info on fetch error

dasch:


tjwp:
Would this fix also apply to v0.6.x? Are any patch releases still being made from that branch?

dasch:
It probably will. Can you cherry pick to the stable v0.6 branch and open a PR?

PR #649, opened 2018-09-14
Clear paused partitions on join group

tjwp:
Similar to https://github.com/zendesk/ruby-kafka/pull/612, when a consumer joins a group the paused partitions that are no longer assigned to the consumer need to be cleared.

We've been seeing errors from this when resuming paused partitions and attempting to resolve offsets that are no longer assigned to the consumer:

```
KeyError: key not found: 11
  from kafka/offset_manager.rb:184:in `fetch'
  from kafka/offset_manager.rb:184:in `resolve_offset'
  from kafka/offset_manager.rb:103:in `next_offset_for'
  from kafka/consumer.rb:470:in `seek_to_next'
  from kafka/consumer.rb:155:in `resume'
  from kafka/consumer.rb:487:in `block (2 levels) in resume_paused_partitions!'
  from kafka/consumer.rb:478:in `each'
  from kafka/consumer.rb:478:in `block in resume_paused_partitions!'
  from kafka/consumer.rb:477:in `each'
  from kafka/consumer.rb:477:in `resume_paused_partitions!'
  from kafka/consumer.rb:501:in `fetch_batches'
  from kafka/consumer.rb:288:in `block in each_batch'
  from kafka/consumer.rb:387:in `block in consumer_loop'
```

That trace is from v0.6.8, but it looks like the problem still exists on master.

I'm still working on a unit test (and not sure when I'll get back to that), but wanted to get this out there in case anybody else is experiencing the problem.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -445,6 +445,7 @@ module Kafka
         # We've been out of the group for at least an entire generation, no
         # sense in trying to hold on to offset data
         clear_current_offsets
+        clear_pauses
         @offset_manager.clear_offsets
       else
         # After rejoining the group we may have been assigned a new set of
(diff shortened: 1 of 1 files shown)


tjwp:
Addressed instead by https://github.com/zendesk/ruby-kafka/pull/656

PR #654, opened 2018-09-19
Create thread when starting fetcher

nguyenquangminh0711:
When creating a Fetcher instance in consumer, a thread is created to handle the commands and fetch the data in the background. That threads starts the loop right after the Fetcher is *created*. When the fetcher instance stops, that thread is still running in forever-loop. This creates serious thread leaking in some use cases where many instances of consumers are used. Therefore, I moved the thread initialization to `Fetcher#start` and kill the thread naturally when calling `Fetcher#stop`.  This change seems to be safe and more efficient. In the consumer, before starting the consumer loop, the fetcher is started. Right after the consumer loop breaks (interrupting, error, etc.), the fetcher is cleaned up and stopped totally. 

This script demonstrates the thread leaking in previous versions, and fixed in current version.

```ruby
require 'kafka'
topic = 'hello'

begin
  kafka = Kafka.new(
    seed_brokers: ['localhost:9092'],
    client_id: "simple-consumer",
    socket_timeout: 20
  )

  consumer = kafka.consumer(group_id: "firehose")
  consumer.subscribe(topic)

    #consumer.stop
    raise if rand(0..10) == 5
    puts message.value
  end
rescue => e
  puts e
  consumer.stop
  p Thread.list.count
  retry
end
```

cc @mensfeld @dasch 
P/s: this PR contains some changes to fix flaky tests. They are now stable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,8 @@ jobs:
           KAFKA_DELETE_TOPIC_ENABLE: true
       - image: wurstmeister/kafka:2.11-0.11.0.3
         environment:
+          KAFKA_ADVERTISED_HOST_NAME: localhost
+          KAFKA_ADVERTISED_PORT: 9093
           KAFKA_PORT: 9093
           KAFKA_ZOOKEEPER_CONNECT: localhost:2181
           KAFKA_DELETE_TOPIC_ENABLE: true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,3 +145,4 @@ workflows:
       - kafka-0.11
       - kafka-1.0.0
       - kafka-1.1
+      - kafka-2.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,38 @@ jobs:
       - run: bundle install --path vendor/bundle
       - run: bundle exec rspec --profile --tag functional spec/functional
 
+  kafka-2.0:
+    docker:
+      - image: circleci/ruby:2.5.1-node
+        environment:
+          LOG_LEVEL: DEBUG
+      - image: wurstmeister/zookeeper
+      - image: wurstmeister/kafka:2.11-2.0.0
+        environment:
+          KAFKA_ADVERTISED_HOST_NAME: localhost
+          KAFKA_ADVERTISED_PORT: 9092
+          KAFKA_PORT: 9092
+          KAFKA_ZOOKEEPER_CONNECT: localhost:2181
+          KAFKA_DELETE_TOPIC_ENABLE: true
+      - image: wurstmeister/kafka:2.11-2.0.0
+        environment:
+          KAFKA_ADVERTISED_HOST_NAME: localhost
+          KAFKA_ADVERTISED_PORT: 9093
+          KAFKA_PORT: 9093
+          KAFKA_ZOOKEEPER_CONNECT: localhost:2181
+          KAFKA_DELETE_TOPIC_ENABLE: true
+      - image: wurstmeister/kafka:2.11-2.0.0
+        environment:
+          KAFKA_ADVERTISED_HOST_NAME: localhost
+          KAFKA_ADVERTISED_PORT: 9094
+          KAFKA_PORT: 9094
+          KAFKA_ZOOKEEPER_CONNECT: localhost:2181
+          KAFKA_DELETE_TOPIC_ENABLE: true
+    steps:
+      - checkout
+      - run: bundle install --path vendor/bundle
+      - run: bundle exec rspec --profile --tag functional spec/functional
+
 workflows:
   version: 2
   test:
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -110,6 +110,7 @@ module Kafka
     # @return [nil]
     def stop
       @running = false
+      @fetcher.stop
       @cluster.disconnect
     end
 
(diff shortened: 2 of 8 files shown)


mensfeld:
Here's the code you can use to check if that solves the problem:
```ruby
begin
 kafka = Kafka.new(
   seed_brokers: brokers,
   client_id: "simple-consumer",
   socket_timeout: 20,
   logger: logger,
 )

   consumer = kafka.consumer(group_id: "firehose")
   consumer.subscribe(topic)


 consumer.each_message(topic: topic) do |message|
   #puts message.value
   #consumer.stop
   raise
 end
rescue => e
 consumer.stop
 p Thread.list.count
 retry
end
```

The number of threads shouldn't raise.

mensfeld:
Thank you. This indeed should solve the problem, making the usage of consumers way more flexible. 

Is there a chance to actually backport this fix as well?

nguyenquangminh0711:
@mensfeld This fix is simple and looks safe to me. So, I think there won't be any conflict to the 0.6.x stable versions. @dasch is the only one who can release the version though 😄

dasch:
This needs to be run with production load for a while though – there are some edge cases when restarting a crashed consumer loop that I remember. @mensfeld can you validate this with real workloads?

dasch:
If someone needs a bigger project to work on with regards to ruby-kafka, a much better integration test suite would be great :D

mensfeld:
@dasch I can, however this problem persists for corner-cases that aren't actually related to any of my productions :( The way I've discovered that was by a report of one of the Karafka users, that uses confluence cloud that was dropping the connections upon reaching the tier limits. Anyhow, I can give it a spin just running to make sure it actually works :)

mensfeld:
> If someone needs a bigger project to work on with regards to ruby-kafka, a much better integration test suite would be great :D
I have a bigger integration setup in plans for Karafka. It would pretty much cover many corner cases from ruby-kafka as well, but like any open-source, you paradoxically need some payed resources to do that and Karafka team is already investing their time. Oh... life I guess ;)

lewispb:
We are currently soak testing this on our beta environment and will report back.

mensfeld:
Same here. So far so good. @lewispb could you actually reach the limits of confluence to trigger the issue that caused us to discover this in the first place?

lewispb:
We have been seeing this issue occur even without hitting any limits, so going for 3 days without problems would indicate success in my view.

mensfeld:
@lewispb awesome. It's also running on my stagings. So far so good.

lewispb:
We are now seeing the timeout problem again. This time though, a new error:

```
RuntimeError: already started
  from kafka/fetcher.rb:46:in `start'
  from kafka/consumer.rb:389:in `consumer_loop'
  from karafka/patches/ruby_kafka.rb:15:in `consumer_loop'
  from kafka/consumer.rb:290:in `each_batch'
  from karafka/connection/client.rb:29:in `fetch_loop'
  from karafka/connection/listener.rb:39:in `fetch_loop'
  from karafka/connection/listener.rb:23:in `call'
  from karafka/fetcher.rb:19:in `block (2 levels) in call'
```

nguyenquangminh0711:
@lewispb That error comes from a check that raises exception when the fetcher already started.

```
def start
  raise "already started" if @running
  # ...
end
```

This is not actually an exception. The outsider can call `Fetcher#start` anytime before going into consumer loop to ensure it is running, to be safe. So, the Fetcher should not be panic. Pushed a small change to silently return in this case.

lewispb:
Thanks, will try that out Monday. A new error overnight:

```
TypeError: no implicit conversion of nil into Integer
  from kafka/protocol/encoder.rb:57:in `pack'
  from kafka/protocol/encoder.rb:57:in `write_int32'
  from kafka/protocol/offset_commit_request.rb:31:in `encode'
  from kafka/protocol/request_message.rb:22:in `encode'
  from kafka/protocol/encoder.rb:178:in `encode_with'
  from kafka/connection.rb:162:in `write_request'
  from kafka/connection.rb:103:in `block in send_request'
  from active_support/notifications.rb:170:in `instrument'
  from kafka/instrumenter.rb:21:in `instrument'
  from kafka/connection.rb:96:in `send_request'
  from kafka/broker.rb:188:in `send_request'
  from kafka/broker.rb:86:in `commit_offsets'
  from kafka/consumer_group.rb:82:in `commit_offsets'
  from kafka/offset_manager.rb:125:in `commit_offsets'
  from kafka/consumer.rb:426:in `make_final_offsets_commit!'
  from kafka/consumer.rb:420:in `ensure in consumer_loop'
  from kafka/consumer.rb:422:in `consumer_loop'
  from karafka/patches/ruby_kafka.rb:15:in `consumer_loop'
  from kafka/consumer.rb:290:in `each_batch'
  from karafka/connection/client.rb:29:in `fetch_loop'
  from karafka/connection/listener.rb:39:in `fetch_loop'
  from karafka/connection/listener.rb:23:in `call'
  from karafka/fetcher.rb:19:in `block (2 levels) in call'
```

nguyenquangminh0711:
Hmm. Weird. That error happens because the consumer leaves the consumer loop, tries to make final commit after leaving group *or* before joining any group. Both cases are weird, I could not found any logics lead to this situation. In Karafka, there is a patch on consumer loop (lib/karafka/patches/ruby_kafka.rb) and there is some threaded activities involved. I think we can start investigating from that. cc @mensfeld. This error doesn't change anything, the final commits are discarded anyway in both cases.

Btw, that error does not relate to this PR, that's a different error. So, we can fix that latter.

nguyenquangminh0711:
Since everything seems stable, everyone seems not to have problems with it and this is not a big change, I think I'll merge it now.

PR #655, opened 2018-09-20
adds debugging for CorruptMessage errors

ronan-mch:
Adds error logging when message corruption errors are received.

### DIFF
--- a/lib/kafka/broker.rb
+++ b/lib/kafka/broker.rb
@@ -71,7 +71,31 @@ module Kafka
     def produce(**options)
       request = Protocol::ProduceRequest.new(**options)
 
-      send_request(request)
+      resp = send_request(request)
+
+      return if resp.nil?
+
+      errored_partitions = []
+      resp.topics.each do |topic_info|
+        errored = topic_info.partitions.find {|part_info| part_info.error_code == 2 }
+        errored_partitions << errored unless errored.nil?
+      end
+
+      errored_partitions.each do |partition_info|
+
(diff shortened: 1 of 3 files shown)


dasch:
I've release `v0.7.1.messagecorruptiondebug` with this code.

ronan-mch:
There's a functional spec failing for some reason, is this normal?

dasch:
Yup

lib/kafka/broker.rb
```diff
@@ -71,7 +71,31 @@ def list_offsets(**options)
...
+      resp.topics.each do |topic_info|
+        errored = topic_info.partitions.find {|part_info| part_info.error_code == 2 }
+        errored_partitions << errored unless errored.nil?
+      end
+
+      errored_partitions.each do |partition_info|
+
+        @logger.error "Corrupt message code received for #{request} to partition #{partition_info.partition}"
+
+        #{"topic" => {"partition" => [message, message] }}
```
libo: do we need this comment?

lib/kafka/broker.rb
```diff
@@ -71,7 +71,31 @@ def list_offsets(**options)
     def produce(**options)
       request = Protocol::ProduceRequest.new(**options)
 
-      send_request(request)
+      resp = send_request(request)
+
+      return if resp.nil?
+
+      errored_partitions = []
```
libo: Would it be possible to gate this logging code behind an `ENV` var that we can set in console? So that we can run it in prod?

PR #656, opened 2018-09-20
Only seek if partition still belong to group

jfrancoist:
This PR fixes a `KeyError: key not found:...` in the offset manager https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/offset_manager.rb#L184 during rebalancing.

This happens when a partition is paused in the consumer and a rebalancing occurs in the meanwhile. When the partition pause is over we should verify that it is still assigned to our group before performing a seek. Similar verification is being done here https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L522

#### Risks
Low.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -153,7 +153,8 @@ module Kafka
     def resume(topic, partition)
       pause_for(topic, partition).resume!
 
-      seek_to_next(topic, partition)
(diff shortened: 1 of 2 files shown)


tjwp:
I put up an alternate approach for fixing this here: https://github.com/zendesk/ruby-kafka/pull/649

dasch:
@tjwp I think this approach is a bit simpler.

PR #660, opened 2018-09-29
Enhance pause mechanism

nguyenquangminh0711:
Our recent implementation of pause mechanism supports limited use case. We need to extend it to support more practical use cases, described in https://github.com/zendesk/ruby-kafka/issues/552. A desired usage:

```ruby
kafka.deliver_message('Hello 1', topic: 'a')
kafka.deliver_message('Hello 2', topic: 'a')
kafka.deliver_message('Hello 3', topic: 'a')
consumer = kafka.consumer(group_id: 'consumer-name')
consumer.subscribe('a')
  p message
  consumer.pause(message.topic, message.partition)
end
# Expect only Hello 1 consumed.
```

Or

```ruby

     p message
  end
  consumer.pause(message.topic, message.partition)  if message.partition == 0
end
# Expect partition 0 to be paused after the first batch
```

This issue is address at https://github.com/zendesk/ruby-kafka/pull/575 and https://github.com/zendesk/ruby-kafka/pull/553. However, the two PRs are outdated with the master branch, where from version 0.6.x, we introduce new threaded fetcher. That change makes the situation more complicated.

When a consumer pause a partition, rejecting it in the consumer loop is not enough. That leads to message lost when the consumer resumes. This will demonstrate the situation:
- Fetcher fetches batch 1 of topic/1
- Fetcher fetches batch 2 of topic/1
- Fetcher fetches batch 3 of topic/1
- Consumer pauses topic/1
- Consumer polls batch 1 from the fetcher, then rejected
- Consumer polls batch 2 from the fetcher, then rejected
- Fetcher fetches batch 4 of topic/1
- Consumer resumes topic/1
- Consumer consumes batch 3 and batch 4 and then commit offsets. We lost batch 1 and batch 2

To overcome this issue, I apply some big changes:
- Create PauseManager to manage pause state of multiple parittions and share this object between consumer and fetcher. Fetcher now only fetches the non-paused partitions.
- When a consumer pauses a partition it removes all the fetched batches of a partition in the fetcher.  To enable this, I have to replace fetcher's queue data structure and replace by a similar implementation called LockedQueue.
- Check carefully the pause state of a partition before processing a message. After processing, call seek on the fetcher of recently paused partition to prevent message duplication or lost.

### DIFF
--- a/lib/kafka/fetcher.rb
+++ b/lib/kafka/fetcher.rb
@@ -54,6 +54,10 @@ module Kafka
       @commands << [:start, []]
     end
 
+    def clean(topic, partition)
+      @commands << [:clean, [topic, partition]]
+    end
+
     def handle_start
       raise "already started" if @running
 
@@ -134,6 +138,19 @@ module Kafka
       @next_offsets[topic][partition] = offset
     end
 
+    def handle_clean(topic, partition)
+      @queue.map! do |queue|
+        if queue[0] == :batches
+          new_batches = queue[1].delete_if do |batch|
+            batch.topic == topic && batch.partition == partition
+          end
+          [:batches, new_batches]
+        else
+          queue
+        end
+      end
+    end
+
     def step
       batches = fetch_batches
 
@@ -150,6 +167,7 @@ module Kafka
 
         @next_offsets[batch.topic][batch.partition] = batch.last_offset + 1 unless batch.unknown_last_offset?
       end
+
 
       @queue << [:batches, batches]
     rescue Kafka::NoPartitionsToFetchFrom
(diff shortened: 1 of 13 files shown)


lib/kafka/fetcher.rb
```diff
@@ -173,6 +192,7 @@ def fetch_batches
         max_bytes = @max_bytes_per_partition[topic]
 
         partitions.each do |partition, offset|
+          next if @pause_manager.paused?(topic, partition)
```
dasch: PauseManager isn't thread safe, so this could result in very hard-to-trace bugs.

dasch:
I worry that this will be a maintenance nightmare. Do the use cases warrant this added complexity?

nguyenquangminh0711:
From the discussion on Slack, it's true that this PR adds a lot of complexity. I'm working on a simpler solution. I'll close this one.

PR #661, opened 2018-10-04
Synchronize @worker_thread and @timer_thread in AsyncProducer

### LINKED ISSUE #560: async producer crash

* Version of Ruby: jruby 9.1.8.0 (2.3.1)
* Version of Kafka: 0.11.0.1 (landoop/fast-data-dev)
* Version of ruby-kafka: 0.5.4

I'm sorry, I haven't done that.

###### Steps to reproduce

I'm not sure, it just happens in the middle of the server's up time. Will be happy to co-operate if the issue isn't obvious with the stack trace.

###### Expected outcome

All good.

###### Actual outcome
```
NoMethodError: undefined method `partition' for nil:NilClassWarning
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:337:in `block in assign_partitions!'
org/jruby/RubyArray.java:1734:in `each'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/pending_message_queue.rb:36:in `each'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:336:in `assign_partitions!'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:297:in `block in deliver_messages_with_retries'
org/jruby/RubyKernel.java:1298:in `loop'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:286:in `deliver_messages_with_retries'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:236:in `block in deliver_messages'
/usr/local/bundle/gems/activesupport-4.2.10/lib/active_support/notifications.rb:166:in `instrument'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/instrumenter.rb:19:in `instrument'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/producer.rb:229:in `deliver_messages'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/async_producer.rb:241:in `deliver_messages'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/async_producer.rb:199:in `block in run'
org/jruby/RubyKernel.java:1298:in `loop'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/async_producer.rb:191:in `run'
/usr/local/bundle/gems/ruby-kafka-0.5.4/lib/kafka/async_producer.rb:148:in `block in ensure_threads_running!'
09 Apr 2018 11:57:55.3962018-04-09 08:57:55 UTC severity=ERROR, Async producer crashed!
```

mattbooks:
Should help to fix #560

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes and additions to the library will be listed here.
 
 ## Unreleased
 
+- Synchronize access to @worker_thread and @timer_thread in AsyncProducer to prevent creating multiple threads (#661).
+
 ## 0.7.2
 
 - Handle case when paused partition does not belong to group on resume (#656).
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -59,6 +59,7 @@ module Kafka
   #     producer.shutdown
   #
   class AsyncProducer
+    THREAD_MUTEX = Mutex.new
 
     # Initializes a new AsyncProducer.
     #
@@ -146,11 +147,15 @@ module Kafka
     private
 
     def ensure_threads_running!
-      @worker_thread = nil unless @worker_thread && @worker_thread.alive?
-      @worker_thread ||= Thread.new { @worker.run }
+      THREAD_MUTEX.synchronize do
+        @worker_thread = nil unless @worker_thread && @worker_thread.alive?
+        @worker_thread ||= Thread.new { @worker.run }
+      end
 
-      @timer_thread = nil unless @timer_thread && @timer_thread.alive?
-      @timer_thread ||= Thread.new { @timer.run }
+      THREAD_MUTEX.synchronize do
+        @timer_thread = nil unless @timer_thread && @timer_thread.alive?
+        @timer_thread ||= Thread.new { @timer.run }
+      end
     end
 
     def buffer_overflow(topic, message)


mattbooks:
In terms of other options, it seems that it either needs to hook into something that is called regularly as is done here, or something needs to watch it. If something is to watch it, then you need to make sure that if _that_ dies, then it is restarted as well. You'll be more familiar with the design of the gem then me, so maybe you can provide some input on the alternative approaches, but the current approach seems simpler to me.

As written though, it is very much not thread-safe and trading that off for potential latency improvements seems misguided. Furthermore, `Queue#push` is already present in `produce` and internally does the same thing as this.

[`Queue#push`](https://ruby-doc.org/stdlib-2.0.0/libdoc/thread/rdoc/Queue.html#method-i-push)
```ruby
def push(obj)
  Thread.handle_interrupt(StandardError => :on_blocking) do
    @mutex.synchronize do
      @que.push obj
      @cond.signal
    end
  end
end
```

dasch:
I guess we can try this out and see if there's a non-neglible effect on performance, in which case we'd need to revisit.

dasch:
Thanks!

mattbooks:
Great, thanks! Do you have an idea of when this will be released?

dasch:
Probably within the next couple of weeks.

Issue #664, opened 2018-10-11
Fetched messages can be reprocessed after consumer group rebalance


BenEddy:
* Version of Ruby: 2.5.1
* Version of Kafka: 1.0.0
* Version of ruby-kafka: 0.7.2

When a consumer joins a consumer group, it [tells the fetcher to reset](https://github.com/zendesk/ruby-kafka/blob/a42f9045900764c0ea272d868ab81c17534716b4/lib/kafka/consumer.rb#L461) and then [seeks to committed offsets](https://github.com/zendesk/ruby-kafka/blob/a42f9045900764c0ea272d868ab81c17534716b4/lib/kafka/consumer.rb#L468). The reset and seek commands are handled asynchronously. If the fetcher has previously fetched messages, the consumer will continue processing fetched message before the reset and seek commands are executed. When the seek command executes, it will rewind to offsets behind the most recently processed messages, resulting in reprocessing. 

For example:
```
consumer processes messages 1, 2, 3
consumer commits offset 3
consumer re-joins the group
consumer calls `@fetcher.reset` (async)
consumer effectively calls `@fetcher.seek(offset: 4)` (async)
consumer processes messages 4, 5, 6 from the un-reset fetcher queue
`@fetcher` actually resets
`@fetcher` actually seeks to 4
consumer reprocess messages 4, 5, 6
```

###### Steps to reproduce

- [Here's a script](https://gist.github.com/BenEddy/13a194ad57cf409b6b7e02e9ec12b045) to reproduce the issue using redis as an external offset store 
- [Here's a spec](https://github.com/BenEddy/ruby-kafka/blob/cc13f1f054362db5a63a49de32b89c84f40dd51c/spec/functional/consumer_group_spec.rb#L277-L354) that reproduces the issue. 

###### Expected outcome

Expected no reprocessing after re-joining a consumer group. 

###### Actual outcome

Messages are reprocessed.

PR #665, opened 2018-10-11
Discard messages that were fetched under a previous consumer group generation

BenEddy:
As described in https://github.com/zendesk/ruby-kafka/issues/664, when a consumer re-joins a consumer group, it instructs the fetcher to asynchronously reset and seek to committed offset positions. The consumer will then continue processing fetched messages until the reset and seek take place. When the seek eventually occurs, it will seek to offsets < most recently processed messages. 

This PR tags fetched message batches with the consumer group `generation_id` that was active when fetch took place. When a consumer re-joins a group and receives an incremented `generation_id`, it passes the `generation_id` along to the fetcher reset command. Then, the consumer discards messages until it sees the current `generation_id` in fetched batches and can be assured that the reset occurred. 

An alternative to this implementation would be to make `Fetcher#reset` synchronous. 

Closes https://github.com/zendesk/ruby-kafka/issues/664

### DIFF
--- a/lib/kafka/fetcher.rb
+++ b/lib/kafka/fetcher.rb
@@ -28,6 +28,10 @@ module Kafka
 
       # The maximum number of bytes to fetch per partition, by topic.
       @max_bytes_per_partition = {}
+
+      # An incrementing counter used to synchronize resets between the
+      # foreground and background thread.
+      @current_reset_counter = 0
     end
 
     def subscribe(topic, max_bytes_per_partition:)
@@ -85,6 +89,8 @@ module Kafka
     end
 
     private
+
+    attr_reader :current_reset_counter
 
     def loop
       @instrumenter.instrument("loop.fetcher", {
@@ -206,12 +212,6 @@ module Kafka
       sleep backoff
 
       []
-    end
-
-    # An incrementing counter used to synchronize resets between the
-    # foreground and background thread.
-    def current_reset_counter
-      @current_reset_counter ||= 0
     end
   end
 end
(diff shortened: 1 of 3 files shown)


nguyenquangminh0711:
This idea of adding generation to fetcher is a smart idea. We used to step on this problem before and end up replacing the default Queue by a handmade locking array (https://github.com/zendesk/ruby-kafka/pull/660/files). By following your idea, the fetcher now can easily clean up the queue, or even clean up just only a particular partition (used in pause mechanism). 

But I think we can make some changes to make it simpler and easier to extend in future. The generation id in the fetcher seems not to relate to and should not depend on the consumer group's generation id. Therefore, I suggest to maintain a generation in the fetcher, starting with 0 when it starts, and increases when the fetcher is resetted. When giving out the messages for the outsider, the fetcher rejects stale generations. 

What do you think about this?

BenEddy:
I should have mentioned this in my previous comment - the code has been updated per @nguyenquangminh0711's suggestion.

lib/kafka/fetcher.rb
```diff
@@ -196,5 +207,11 @@ def fetch_batches
 
       []
     end
+
+    # An incrementing counter used to synchronize resets between the
+    # foreground and background thread.
+    def current_reset_counter
+      @current_reset_counter ||= 0
```
nguyenquangminh0711: The ||= operator is not thread-safe. So when calling this lazy initalizer, it is possible to face a threading issue. How about moving this to the initializer of the fetcher?

BenEddy: Sounds good 👍. Updated!

Issue #675, opened 2018-10-31
Message reprocessing due to failure to commit offsets before joining consumer group


BenEddy:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.5.1
* Version of Kafka: 1.0.0
* Version of ruby-kafka: `master` @ https://github.com/zendesk/ruby-kafka/commit/5b322043dc7d9c2aadee25a564ad8dbf3ff23d12

If a consumer has processed messages since the last offset commit, messages may be reprocessed after a consumer group rebalance. When a new consumer joins a group, existing consumers must rejoin the group to receive partition assignments. Currently, consumers don't synchronously commit offsets before rejoining a group. If a partition is reassigned from consumer 1 to consumer 2, consumer 2 will resume at the last committed offset, which could be < last processed offset.

For example:

```
Consumer 1 is assigned topic "test", partition 0
Consumer 1 processes messages 10,11,12
Consumer 1 commits offset 12
Consumer 1 process messages 13,14,15
Consumer 2 joins the group
Consumer 1 gets a heartbeat error and rejoins the group
Consumer 2 is assigned partition 0
Consumer 1 is assigned nothing
Consumer 2 starts processing at 13, reprocessing 13,14,15
```

###### Steps to reproduce

Here's a script that runs a producer and a consumer, and then a second consumer joins the group. 

```Ruby
require "bundler/setup"
require "kafka"
require "tempfile"

Thread.abort_on_exception

BROKERS = ["localhost:9092"]
TOPIC = "test" # Make sure this topic has > 1 partitions
GROUP_ID = "test-group"

producer_pid = fork do
  producer = Kafka.new(seed_brokers: BROKERS).producer
  loop do
    5.times { producer.produce("message", topic: TOPIC) }
    producer.deliver_messages
    sleep 0.1
  end
end

def run_consumer(id:, output:)
  puts "Started consumer #{id}..."

  consumer = Kafka.new(seed_brokers: BROKERS).consumer(group_id: GROUP_ID)
  consumer.subscribe(TOPIC, start_from_beginning: false)

  trap("TERM") { consumer.stop }

  consumer.each_batch(min_bytes: 1024 * 5, max_wait_time: 2) do |batch|
    batch.messages.each do |message|
      output.puts([message.partition, message.offset].join("-"))
    end
  end

  puts "Stopped consumer #{id}..."
end

consumer_1_file = Tempfile.new('consumer_1')
[...]

PR #676, opened 2018-10-31
Synchronously commit offsets after HeartbeatError

BenEddy:
As detailed in https://github.com/zendesk/ruby-kafka/issues/675, if a consumer has processed messages since the last offset commit, messages may be reprocessed after a consumer group rebalance. This PR attempts to synchronously commit offsets after the consumer learns of the rebalance (from `HeartbeatError`) and before rejoining the group. For reference, the java client also tries to [commit offsets before joining a group](https://github.com/apache/kafka/blob/044350f28e4d0edcc0503b2bc2593c0d25948ed0/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L449-L452):

```Java
    @Override
    protected void onJoinPrepare(int generation, String memberId) {
        // commit offsets prior to rebalance if auto-commit enabled
        maybeAutoCommitOffsetsSync(time.timer(rebalanceTimeoutMs));
```

Closes https://github.com/zendesk/ruby-kafka/issues/675

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -203,7 +203,7 @@ module Kafka
         max_wait_time: max_wait_time,
       )
 
-      consumer_loop do
+      consumer_loop(automatically_mark_as_processed: automatically_mark_as_processed) do
         batches = fetch_batches
 
         batches.each do |batch|
@@ -288,7 +288,7 @@ module Kafka
         max_wait_time: max_wait_time,
       )
 
-      consumer_loop do
+      consumer_loop(automatically_mark_as_processed: automatically_mark_as_processed) do
         batches = fetch_batches
 
         batches.each do |batch|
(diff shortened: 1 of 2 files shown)


lib/kafka/consumer.rb
```diff
@@ -203,7 +203,7 @@ def each_message(min_bytes: 1, max_bytes: 10485760, max_wait_time: 1, automatica
         max_wait_time: max_wait_time,
       )
 
-      consumer_loop do
+      consumer_loop(automatically_mark_as_processed: automatically_mark_as_processed) do
```
BenEddy: It seems like we could make `automatically_mark_as_processed` an instance variable and avoid passing it to `consumer_loop`, but I'm not sure.

BenEddy:
@nguyenquangminh0711 Great suggestions! Updated 👍

Issue #647, opened 2018-09-13
Unbalanced Consumer Assignment with Unbalanced Topics


psycotica0-shopify:
This is mostly a question to make sure I'm understanding the code.

I was looking at the [RoundRobinAssignmentStrategy](https://github.com/zendesk/ruby-kafka/blob/cc52504bc2b8dc59dbccbebc1e0e64a7c1aaeda8/lib/kafka/round_robin_assignment_strategy.rb#L20-L50) and it _looks_ like it's slicing at the partition level.

So if I have 3 topics, the first has 4 partitions, and the other two each have 1, and I have 3 consumers, I think it'll assign as follows:

| Consumer | Assignment |
| - | - |
| consumer0 | `t0p0, t1p0, t2p0, t0p3` |
| consumer1 | `t0p1` |
| consumer2 | `t0p2` |

Which doesn't _feel_ right.
It's Round Robin-y from the perspective of a single topic, but not overall.

It also limits such a group's scalability to highest number of partitions.
In theory there are 6 partitions here that _could_ be assigned, but I can't scale higher than 4 consumers because it always gives 3 of them to the first consumer.

Am I way off?

nguyenquangminh0711:
@psycotica0-shopify In real life environment, like production, this barely happens. If the number of partitions increases, like 10 partitions per topic, the imbalances between consumers fade out. And IMO, that's way better in term of partition consuming distribution. The rule of thumb is that number of partitions should be greater than number of consumers, for each topic.

Elyahou:
@nguyenquangminh0711 

In the case of
- 3 topics with 20 partitions each
- 1 consumer group with 60 members

If I understand it correctly, the current algorithm will assign 3 partitions (1 for each topic) to 20 members, and the 40 remaining will be without any assignment

Maybe we can enhance this by evenly distribute to members, and in that case assign 1 partition for each member

nguyenquangminh0711:
@Elyahou Yes, that's true that the current round-robin strategy doesn't work well in that case. It's not hard to fix this issue. I'm just curious about your concrete use case. In your described configuration, do all 3 topics have the same schema and meanings? In most of my use cases, I usually split into 3 different consumer groups, each consumer group consumes from a topic:
- Group 1 (20 members) consumes from topic 1 (20 partitions)
- Group 2 (20 members) consumes from topic 2 (20 partitions)
- Group 3 (20 members) consumes from topic 3 (20 partitions)

[...]

PR #681, opened 2018-11-06
Even out partition assignment

isaacseymour:
In a situation where there are more consumers than partitions on all
topics, some consumers will be assigned no work at all with the current
code.

For example, with 2 topics, each with 2 partitions, and 3
consumers running, each topic's partition 0 will be assigned to the
first member ID, and partition 1 will be assigned to the second member
ID. This means 2 members with 2 partitions to consume, and 1 with
nothing!

With this change, all topic:partitions are considered together. In the
above example, 2 members would get 1 partition each, and one would be
assigned 2 partitions.

### DIFF
--- a/lib/kafka/round_robin_assignment_strategy.rb
+++ b/lib/kafka/round_robin_assignment_strategy.rb
@@ -33,7 +33,8 @@ module Kafka
         Array.new(partitions.count) { topic }.zip(partitions)
       end
 
-      partitions_for_each_member = topic_partitions.sort.each_slice(topic_partitions.count / members.count)
+      partitions_per_member = [1, topic_partitions.count / members.count].max
+      partitions_for_each_member = topic_partitions.sort.each_slice(partitions_per_member)
 
       members.zip(partitions_for_each_member).each do |member_id, member_partitions|
         unless member_partitions.nil?
--- a/lib/kafka/round_robin_assignment_strategy.rb
+++ b/lib/kafka/round_robin_assignment_strategy.rb
@@ -33,19 +33,14 @@ module Kafka
         Array.new(partitions.count) { topic }.zip(partitions)
       end
 
-      partitions_per_member = [1, (topic_partitions.count / members.count)].max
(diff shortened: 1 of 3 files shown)


nguyenquangminh0711:
This PR would solve https://github.com/zendesk/ruby-kafka/issues/647.

lib/kafka/round_robin_assignment_strategy.rb
```diff
@@ -24,20 +24,28 @@ def assign(members:, topics:)
...
+        Array.new(partitions.count) { topic }.zip(partitions)
+      end
+
+      partitions_per_member = [1, (topic_partitions.count / members.count)].max
+      partitions_for_each_member = topic_partitions.sort.each_slice(partitions_per_member).to_a
 
-        partitions_per_member = partitions.group_by {|partition_id|
-          partition_id % members.count
-        }.values
+      if partitions_for_each_member.count > members.count
```
nguyenquangminh0711: I don't get the idea behind this code. Isn't it a little bit complicated? The problem with old code is that it is topic-scoped. So, to make the partitions spread equally, we just need to split by all topics, not single one. Something like this:

```ruby
topic_partitions = topics.flat_map do |topic|
  begin
    @cluster.partitions_for(topic).map { |p| [topic, p.partition_id }
  rescue UnknownTopicOrPartition
    raise UnknownTopicOrPartition, "unknown topic #{topic}"
  end
end

partitions_per_member = partitions.group_by.each_index do |_, index| 
  index % members.count 
end.values

members.zip(partitions_per_member).each do |member_id, member_partitions|
  unless member_partitions.nil?
    member_partitions.each do |topic, partition|
      group_assignment[member_id].assign(topic, [partition])
    end
  end
end
```

isaacseymour: You're 100% right. I thought I had a simpler implementation but then it missed a whole load of edge cases and fixing them made it all vastly less simple. Updated.

Issue #682, opened 2018-11-06
CorruptMessage raised if message size is in the range of  32760 - 32831 bytes(inclusive)


doritochips:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.4
* Version of Kafka:
* Version of ruby-kafka: 0.7.2 +

The problem persists in the latest master.

###### Steps to reproduce

```ruby
# Please write an example that reproduces the problem you're describing.
require "bundler/setup"
require "kafka"

def foo(msg)
  puts "=="*50
  kafka = Kafka.new(
    seed_brokers: ["shopify.broker"],
    client_id: "whatever",
    logger: Logger.new(STDOUT)
  )

  producer = kafka.producer

  producer.produce( msg, topic: "flow_runs")
  producer.deliver_messages
rescue => _
end

# corrupted
puts ">> corrupted"
msg = "1" * 32760
foo(msg)
puts "<<"*20

# not corrupted
puts ">> not corrupted"
msg = "1" * 32759
foo(msg)
puts "<<"*20
```

###### Expected outcome

The messages should be delivered successfully.

###### Actual outcome

CorruptMessage is raised.

nguyenquangminh0711:
This issue could be same as described in https://github.com/zendesk/ruby-kafka/issues/680. At the first glance, it could be from the checksum method, but not 100% sure. I'm working on this issue.

doritochips:
Also did a `git bisect` and here is the commit that might cause the problem: d19432c13f34019f9e2b532a2fbaa4a5d4761cb9
Hope this helps.

Issue #680, opened 2018-11-06
Can't produce messages with over 98,295 bytes in ruby-kafka 0.7.1+


jturkel:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.5.1
* Version of Kafka: 1.1.1
* Version of ruby-kafka: 0.7.1, 0.7.2, 0.7.3 and master

###### Steps to reproduce

There's a failing test on https://github.com/zendesk/ruby-kafka/compare/master...jturkel:feature/big-message-failure but pasting the relevant bits here:

```ruby
kafka = Kafka.new(...)
producer = kafka.producer(max_retries: 3, retry_backoff: 1)

value = 'a' * 98296
producer.produce(value, topic: topic, partition: 0)
producer.deliver_messages

message = kafka.fetch_messages(topic: topic, partition: 0, offset: :earliest).last

expect(message.value).to eq value
```

###### Expected outcome

The test should pass like it does in ruby-kafka <= 0.7.0.

###### Actual outcome

The test fails with the following error:

```
     Kafka::DeliveryFailed:
       Failed to send messages to f2e93c50d3490054-topic-295d4aee-8cdc-41aa-ad0d-8a0686a565b9/0
     # ./lib/kafka/producer.rb:412:in `deliver_messages_with_retries'
     # ./lib/kafka/producer.rb:247:in `block in deliver_messages'
     # /Users/jturkel/.rvm/gems/ruby-2.5.1@ruby-kafka/gems/activesupport-5.2.1/lib/active_support/notifications.rb:168:in `block in instrument'
     # /Users/jturkel/.rvm/gems/ruby-2.5.1@ruby-kafka/gems/activesupport-5.2.1/lib/active_support/notifications/instrumenter.rb:23:in `instrument'
     # /Users/jturkel/.rvm/gems/ruby-2.5.1@ruby-kafka/gems/activesupport-5.2.1/lib/active_support/notifications.rb:168:in `instrument'
     # ./lib/kafka/instrumenter.rb:21:in `instrument'
     # ./lib/kafka/producer.rb:240:in `deliver_messages'
     # ./spec/functional/producer_spec.rb:135:in `block (2 levels) in <top (required)>'
```

Here is the corresponding log output (with most of the message value elided for readability):
```
I, [2018-11-05T21:55:23.811246 #19885]  INFO -- : Fetching cluster metadata from kafka://localhost:9092
[...]

PR #684, opened 2018-11-07
Fix wrong varint calculation that leads to message corruption

nguyenquangminh0711:
The issue is reported in https://github.com/zendesk/ruby-kafka/issues/682 and https://github.com/zendesk/ruby-kafka/issues/680. In summary, ruby-kafka fails to deliver messages when the message size is equal to some magic number, like 32760 or 98296.

The issues come from varint encoding method of the encoder. The bit mask used in the method to determine the ending of encoding process is 16-bit long (2 bytes), which doesn't cover enough cases. Consider this example:
- Message size is 98296 byte. Record batch format adds some overhead and it becomes 98304 bytes.
- 98304 is `11000000000000000` in binary.
- When calculaing `int & 0xff80`, the 0xff80 is actually `000001111111110000000`
- So, the result is 0. This makes the process stop prematurely.

This error happens with all integers which are over 16-bit long, and have bit 7 -> 16 filled by zero. The condition is to stop when the integer has 7 bit left. So, we can replace error-prone bit mask by shifting operation, which produces the same result. 

I tested with reported examples. Could you please verify this fix on your system?
cc @doritochips, @jturkel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Changes and additions to the library will be listed here.
 
 ## Unreleased
+- Fix wrong encoding calculation that leads to message corruption (#682, #680)
 
 ## 0.7.3
 
--- a/lib/kafka/protocol/encoder.rb
+++ b/lib/kafka/protocol/encoder.rb
@@ -133,7 +133,7 @@ module Kafka
         int = ~int | 1 if int < 0
 
         chunks = []
-        while int & 0xff80 != 0
+        while int >> 7 != 0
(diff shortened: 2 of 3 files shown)


dasch:
:heart:

We ran into this in production and have been pulling our hair out!

klippx:
@nguyenquangminh0711 the type of contributor that all open source maintainers dream about 💖

jwilger:
We also ran into this in production shortly before the associated bugs were reported. Thank you so much to all involved in getting it fixed. First round's on me if we're ever in the same place! 👍

Issue #686, opened 2018-11-15
Processing error causes Kafka to exit the fetcher thread


mensfeld:
* Version of Ruby: `ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]`
* Version of Kafka: `1.0.1`
* Version of ruby-kafka: `0.7.4`

Does **NOT** happen for `0.7.2`. It does happen for `0.7.3` as well.

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka:

it is not fixed.

###### Steps to reproduce

```ruby
require 'kafka'

logger = Logger.new(STDOUT)

kafka = Kafka.new(
  ['localhost:9092'],
  logger: logger
)

10.times do |i|
  kafka.deliver_message("Hello, World! #{i}", topic: 'test')
end

consumer = kafka.consumer(group_id: 'test')

consumer.subscribe('test')

begin
  consumer.each_message do |message|
    puts message.offset, message.key, message.value
    raise StandardError
  end
rescue Kafka::ProcessingError => e
  p e
  retry
end
```

###### Expected outcome

It is expected to retry the processing.

###### Actual outcome

`Fetcher thread exited` - it stops fetching messages and no batches to process.

```
I, [2018-11-15T11:45:44.858521 #31649]  INFO -- : New topics added to target list: test
I, [2018-11-15T11:45:44.858569 #31649]  INFO -- : Fetching cluster metadata from kafka://172.20.24.239:9092
D, [2018-11-15T11:45:44.858604 #31649] DEBUG -- : Opening connection to 172.20.24.239:9092 with client id ruby-kafka...
D, [2018-11-15T11:45:44.858805 #31649] DEBUG -- : Sending topic_metadata API request 1 to 172.20.24.239:9092
D, [2018-11-15T11:45:44.858898 #31649] DEBUG -- : Waiting for response 1 from 172.20.24.239:9092
D, [2018-11-15T11:45:44.859832 #31649] DEBUG -- : Received response 1 from 172.20.24.239:9092
I, [2018-11-15T11:45:44.859867 #31649]  INFO -- : Discovered cluster metadata; nodes: 172.20.24.239:9092 (node_id=1001)
D, [2018-11-15T11:45:44.859885 #31649] DEBUG -- : Closing socket to 172.20.24.239:9092
[...]

PR #687, opened 2018-11-16
Stop fetcher thread synchronously

anujdas:
Stopping the fetcher asynchronously (by emitting a message onto a queue
for pickup) results in a potential race condition on consumer processing
errors, wherein the consumer can be restarted while the fetcher is
sleeping, resulting in the fetcher being stopped after the consumer has
already started. This results in processing errors taking down the
consumer in a hard-to-detect way: the fetcher thread, being stopped, no
longer fetches any new messages, but the consumer continues to loop as
though everything is fine.

Fix this by waiting until the fetcher thread to stops when it's asked
to. I don't think there are performance implications here: given that we
only stop the fetcher before committing offsets and leaving the consumer
group, there doesn't appear to be any benefit to leaving it running.


anujdas:
This PR fixes https://github.com/zendesk/ruby-kafka/issues/686.

PR #689, opened 2018-11-16
Mark fetcher as running before starting thread

### LINKED ISSUE #688: Fetcher stopped at startup (racing condition?)

* Version of Ruby: 2.5.3
* Version of Kafka: 1.1.1
* Version of ruby-kafka: 0.7.3

###### Steps to reproduce

This is a racing condition, so hard to reproduce. `Fetcher` will set `@running = true` after creating the thread which means if the thread runs before `@running` is set, then the thread exits.

Line: https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/fetcher.rb#L60

###### Expected outcome

Fetcher thread should run until we stop the consumer.

###### Actual outcome

Fetcher thread exited at startup which means the consumer was up but not actually reading any messages.

isaacseymour:
This should fix #688 I think.

There seems to be a case where the background fetching thread starts running before `@running = true` executes. That causes the fetcher thread to shut itself down before it does any work.

### DIFF
--- a/lib/kafka/fetcher.rb
+++ b/lib/kafka/fetcher.rb
@@ -49,6 +49,8 @@ module Kafka
     def start
       return if @running
 
+      @running = true
+
       @thread = Thread.new do
         while @running
           loop
@@ -56,8 +58,6 @@ module Kafka
         @logger.info "Fetcher thread exited."
       end
       @thread.abort_on_exception = true
-
-      @running = true
     end
 
     def stop


nguyenquangminh0711:
It's true that the misplacement of `@running = true` will cause the fetcher to stop even when it starts. Thanks for patching this ❤️

ronan-mch:
@nguyenquangminh0711 - would you be able to make a new release for this please? We are having a lot of problem with this issue in production...

nguyenquangminh0711:
@ronan-mch Can you temporarily use the master branch to fix this issue. I'm planning a more complete solution that would fix some other bugs. They will be more things in next release.

PR #691, opened 2018-11-26
Add the option of including metadata into messages

dorner:
We have a pattern where we are using Kafka to be the data backbone for our systems - we write data into a database and use Kafka to publish that data. Sometimes there's downtime for our brokers or schema registry which results in a message being unable to be published even after retries. We have a method to re-publish this message after a wait time. However, at this point the database may have been updated, meaning the message would be outdated. We want to inspect the original payload to check the latest state and re-send the message.

The problem is that we've already Avro-encoded the message and we'd need to decode it again to get access to the original. We also need to store the class name we're using to do the publishing so it can use its internal logic to get the right model from the database.

This PR adds the ability to add metadata to all messages. This data is ignored most of the time, unless there is an error, at which point it is added into the `DeliveryFailed` exception.

Questions and comments welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -734,11 +734,14 @@ All notification events have the `client_id` key in the payload, referring to th
   * `topic` is the topic that the message was produced to.
   * `buffer_size` is the size of the producer buffer after adding the message.
   * `max_buffer_size` is the maximum size of the producer buffer.
+  * `metadata` is any additional metadata that was passed into the `produce` method.
 
 * `deliver_messages.producer.kafka` is sent whenever a producer attempts to deliver its buffered messages to the Kafka brokers. It includes the following payload:
   * `attempts` is the number of times delivery was attempted.
   * `message_count` is the number of messages for which delivery was attempted.
   * `delivered_message_count` is the number of messages that were acknowledged by the brokers - if this number is smaller than `message_count` not all messages were successfully delivered.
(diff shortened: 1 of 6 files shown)


dasch:
So this metadata is never transmitted to the broker, it's an ephemeral in-memory hash? It's sort of weird. Could you try to make it work with message headers first?

dorner:
@dasch Hmm -my issue is that I don't actually need to pass these to the message being sent out - most of the time they won't even be needed. It's only if there's a communication error sending the message where I want metadata about what the message was that failed to send.

I guess another way to implement this might be an internal hash of [topic, partition, offset] => metadata and then store the metadata in there as soon as a message tries to be sent, and delete it once it's successfully sent, and add the metadata to just the failed messages in the `deliver_message` instrumentation. That might get messy when sending batches though. :(

Another way might be for me to try to reverse-engineer the metadata from the message. In my case  I'm trying to store the producer class name and the un-encoded payload. I can do something where I decode the message on error and then figure out which producer could possibly produce to that topic. That might be error-prone though.

dasch:
It doesn't sound like it's something that generalizes, so I'm reluctant to add support for it in ruby-kafka.

dorner:
We came up with a better way to handle this. Closing it off.

PR #692, opened 2018-11-29
Add instrumentation for heartbeats

dorner:
This is useful for lag reporting as heartbeats e.g. during batch processing are more frequent than the consumer loop.


dorner:
Awesome! :D

nguyenquangminh0711:
@dorner Thanks for your PR, we actually need this instrumentation. Sorry for the late response though, I didn't online much last week.

PR #694, opened 2018-12-03
Instrument batch fetching

dasch:
1. Track the offset lag _before_ processing a batch – if processing fails, we still want to know the lag.
2. Start tracking the size of batches. This can be important when tuning e.g. max bytes per partition in order to not overwhelm a downstream system.


dasch:
@nguyenquangminh0711 do you have time to review?

dasch:
@nguyenquangminh0711 the `histogram` method was missing...

PR #697, opened 2018-12-27
Add tagged logging for more verbose / useful logs.

dorner:
Currently, RubyKafka's logs are peppered with single log messages that don't have context - especially in debug mode. Things like `Received response 6 from broker.hostname` or `No batches to process.`. In a case where we have multiple threads listening to different topics (e.g. Phobos), it can make it almost impossible to trace what's happening to a particular consumer or producer.

This PR adds support for `ActiveSupport::TaggedLogging`, which lets us add relevant tags to all log messages. These tags are thread-specific, so they are added to a particular block of code and everything inside that code automatically inherits the tags. My output now looks something like this:

```
[[fadmin_item_groups_group] {fadmin_item_groups: 0}:] [fetch] Waiting for response 12 from macbook-pro-3:9092
[[fadmin_flyer_types_group] {Flyers.FlyerType: 0}:] [fetch] Waiting for response 12 from macbook-pro-3:9092
[[storefronts_status_group] {storefronts.task.status: 0}:] [fetch] Received response 11 from macbook-pro-3:9092
[[storefronts_status_group] {storefronts.task.status: 0}:] Fetching batches
[[storefronts_status_group] {storefronts.task.status: 0}:] [fetch] Sending fetch API request 12 to macbook-pro-3:9092
[[storefronts_status_group] {storefronts.task.status: 0}:] [fetch] Waiting for response 12 from macbook-pro-3:9092
```

It logs group IDs, topics, partitions and the current API request name.

I figured that since we already have an optional dependency on ActiveSupport for instrumentation, adding another ActiveSupport dependency to give us much more useful logs would be acceptable.

Haven't used this in production yet but it seems benign enough.

Comments and suggestions welcome!


lib/kafka/consumer_group.rb
```diff
@@ -186,6 +186,12 @@ def synchronize
       end
     end
 
+    def to_s
+      "[#{@group_id}] {" + assigned_partitions.map { |topic, partitions|
+        "#{topic}: #{partitions.join(', ')}"
+      }.join('; ') + '}:'
```
dasch: This can quickly grow out of hand – we have topics with 100 partitions (don't ask...)

dasch:
I actually think it would be better to use a custom Logger rather than depend on AS::TaggedLogger. This is what I'm using in another project:

```ruby
class PrefixedLogger
  def initialize(prefix, logger)
    @prefix = prefix
    @logger = logger
  end

  [:unknown, :fatal, :error, :warn, :info, :debug].each do |level|
    class_eval <<-RUBY, __FILE__, __LINE__ + 1
      def #{level}(message)
        @logger.#{level}("[\#{@prefix}] \#{message}")
      end
    RUBY
  end
end
```

Then we would wrap any logger passed to `Kafka.new` with this class.

lib/kafka/consumer_group.rb
dorner: Oof... maybe I can just put in the first 5 and then ... - that should suffice for most use cases.

dorner:
Can you elaborate on why the custom logger would do better than the tagged logger? We can also wrap any logger that comes in with a TaggedLogger if that's the way you want to do it - I just feel like the TaggedLogger is more of a standard and would be better understood. It also allows to push/pop tags which matches more closely with the flow I've put in.

dasch:
I'd rather have something that doesn't depend on AS in order for the logging to be consistent.

dorner:
@dasch Took a look at that - thanks for keeping the conversation going!

The issue with the approach in #701 is that you are stuck with whatever information you had when the logger was initialized. E.g. for connections, you can't tag the request with the current request, just the fact that it's *in* a connection. That's really the information I find useful - if there's a request failing, I want to know which request it is, and which topic/partition it's being used for, so I can reconcile my metrics with my logs when I have a dozen different producers and consumers in the same service (which I do).

I also still don't really understand your reticence about using AS when you already use it for instrumentation.

Re partitions, I'm fine with limiting it to the first N (that can be 1 or 3 or 5 or whatever), just so I have some idea of which one is being affected. All of our topics have 5 or fewer partitions, so I get that this wouldn't help your use case, but it would certainly make a big difference for mine. :)

dasch:
The problem is that AS is an opt-in dependency, and I'd like the logging to work the same whether or not you have it. I can update the PrefixedLogger to add support for runtime tags.

dorner:
Ah - I hadn't realized that. Fair enough. In that case we can honestly just copy the single TaggedLogging file from AS and use it as a drop-in replacement. :)

dasch:
If it's not too complicated.

dorner:
@dasch I've updated the PR. I set the logger everywhere to TaggedLogger just to be safe, so we can always ensure that the logger responds to the correct methods, even though right now it's only used in a handful of classes. Calling `TaggedLogger.new` on something that's already a TaggedLogger will just return the original, so it should be safe. Comments welcome!

dorner:
Sweet! Let me know when the version bump happens so I can try it out! ❤️

Elyahou:
I think it is kind of annoying for application that use a global logger, because this PR alter it, ans so all other usage of this logger outside of ruby-kafka will receive theses tags

dorner:
Hm, that's true. I actually see this as a feature instead of a bug though - if you have your own application code running inside RubyKafka code (e.g. notifications), I'd actually like to have those tags in those log lines so I can see what's happening where.

If your code is running outside of RubyKafka code, it shouldn't make a difference, because tags are pushed and popped in a thread-safe manner and you shouldn't see any changes.

PR #700, opened 2019-01-17
Introduce regex matching in Consumer#subscribe

### LINKED ISSUE #299: Feature Request: subscribe to multiple topics by regex
This feature is [supported](https://kafka.apache.org/0101/javadoc/org/apache/kafka/clients/consumer/Consumer.html#subscribe(java.util.regex.Pattern,%20org.apache.kafka.clients.consumer.ConsumerRebalanceListener)) by the canonical Java client and would be useful for my company’s architecture.

Thanks!

rkrage:
Resolves #299 

This PR certainly isn't ready to merge yet. A few notes / questions:

- This doesn't auto-discover new topics after the consumer starts - not sure what that implementation would look like
- Do you have any ideas on how to avoid the code duplication introduced in the `cluster_topics` method
- What kind of tests would be worthwhile for this?

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -103,6 +103,21 @@ module Kafka
       nil
     end
 
+    def subscribe_by_regex(topic_regex, default_offset: nil, start_from_beginning: true, max_bytes_per_partition: 1048576)
+      cluster_topics.each do |topic|
+        next unless topic =~ topic_regex
+
+        subscribe(
+          topic,
+          default_offset: default_offset,
+          start_from_beginning: start_from_beginning,
+          max_bytes_per_partition: max_bytes_per_partition,
+        )
+      end
+
+      nil
+    end
+
     # Stop the consumer.
     #
     # The consumer will finish any in-progress work and shut down.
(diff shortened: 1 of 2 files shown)


lib/kafka/consumer.rb
```diff
@@ -103,6 +103,21 @@ def subscribe(topic, default_offset: nil, start_from_beginning: true, max_bytes_
       nil
     end
 
+    def subscribe_by_regex(topic_regex, default_offset: nil, start_from_beginning: true, max_bytes_per_partition: 1048576)
```
dasch: Could we call this e.g. `subscribe_matching` instead?

dasch: Or maybe just merge this into `subscribe`? I actually think it's better to just check if the user passes a regex there. Then existing integrations can just start passing regexes.

lib/kafka/consumer.rb
```diff
@@ -103,6 +103,21 @@ def subscribe(topic, default_offset: nil, start_from_beginning: true, max_bytes_
...
+      cluster_topics.each do |topic|
+        next unless topic =~ topic_regex
```
dasch: `cluster_topics.select {|topic| topic =~ topic_regex }.each do |topic|

rkrage:
If this looks alright, I'll add some functional tests and update the inline documentation

mensfeld:
I did ;) LGTM!

dasch:
Thanks!

PR #701, opened 2019-01-25
Prefix log lines with relevant info

dasch:


lib/kafka/consumer.rb
```diff
@@ -46,7 +46,7 @@ class Consumer
 
     def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:)
       @cluster = cluster
-      @logger = logger
+      @logger = PrefixedLogger.new(group.group_id, logger)
```
nguyenquangminh0711: Since we already have `consumer` and `worker`, I think this prefix should be `consumer/#{group.group_id}`, and the below one is `consumer/#{group.group_id}/fetcher`

PR #706, opened 2019-02-11
use maxTimestamp for logAppendTime timestamps

Elyahou:
When consuming records that were produced to kafka with the new Record format and when the `log.message.timestamp.type` configuration for the broker/topic is `LogAppendTime`, we need to fetch the record timestamp from `maxTimestamp` field and not from `FirstTimestamp + delta`. Because Kafka send `-1` as the `FirstTimestamp` in this case.

This Pr fix this.

Similar implementation can be found for other clients:
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/record/DefaultRecord.java#L325-L327
https://github.com/Shopify/sarama/blob/a6c1f7eaf9e4ed696183f5be645361ac97ab38bf/consumer.go#L526-L528

### DIFF
--- a/lib/kafka/protocol/record_batch.rb
+++ b/lib/kafka/protocol/record_batch.rb
@@ -11,6 +11,7 @@ module Kafka
       CODEC_ID_MASK = 0b00000111
       IN_TRANSACTION_MASK = 0b00010000
       IS_CONTROL_BATCH_MASK = 0b00100000
+      TIMESTAMP_TYPE_MASK = 0b001000
 
       attr_reader :records, :first_offset, :first_timestamp, :partition_leader_epoch, :in_transaction, :is_control_batch, :last_offset_delta, :max_timestamp, :producer_id, :producer_epoch, :first_sequence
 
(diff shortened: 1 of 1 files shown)


Elyahou:
After some tests I see that there is many cases where even if `firstTimestamp` is not `-1`, the calculation of the message timestamp is wrong. I got this from my tests:

RecordBatch with two records:

```
max_timestamp: 2019-02-10 00:30:59 +0000

Record 1: record.create_time =  first_timestamp + record.timestamp_delta = 2019-02-09 23:27:01 +0000 + 0 = 2019-02-09 23:27:01 +0000
Record2: record.create_time =  first_timestamp + record.timestamp_delta = 2019-02-09 23:27:01 +0000 + 3837565 = 2019-03-26 09:26:26 +0000
```

When reading the same messages with ruby-kafka `0.5.5` or with java, the timestamp of these messages is `2019-02-10 00:30:59`

Elyahou:
@nguyenquangminh0711 Can you please take a look

Elyahou:
Thanks ! When a new version will be released ?

dasch:
I'll let @nguyenquangminh0711 decide that :)

Issue #707, opened 2019-02-19
AsyncProducer will enter to an infinite loop when the internal `sync_producer` will keep raising `Kafka::BufferOverflow`


gilibelz2:
* Version of Ruby: 2.4.2
* Version of Kafka: 2.0
* Version of ruby-kafka: 0.7.5

###### Steps to reproduce

In the `async_producer`, private method `produce` will enter to an infinite loop when the internal `sync_producer` will keep raising `Kafka::BufferOverflow`.

```
      def produce(*args)
        @producer.produce(*args)
      rescue BufferOverflow
        deliver_messages
        retry
      end
```

###### Please write an example that reproduces the problem you're describing.

Send at least one message above the `producer.max_buffer_size` (default 1000) and turn off all the brokers. 

###### Expected outcome

There should be a limit number of retries for `producer.produce` call.

###### Actual outcome

Endless loop.

PR #708, opened 2019-02-19
Async producer limit number of retries

gilibelz2:
Related issue: https://github.com/zendesk/ruby-kafka/issues/707

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -258,7 +258,7 @@ module Kafka
             retry
           else
             @logger.error("Failed to asynchronously produce messages due to BufferOverflow")
-            @instrumenter.instrument("error.async_producer", {error: e})
+            @instrumenter.instrument("error.async_producer", { error: e })
           end
         end
       end
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -83,13 +83,13 @@ module Kafka
       @logger = TaggedLogger.new(logger)
 
       @worker = Worker.new(
-        queue: @queue,
-        producer: sync_producer,
-        delivery_threshold: delivery_threshold,
-        max_retries: max_retries,
-        retry_backoff: retry_backoff,
-        instrumenter: instrumenter,
-        logger: logger,
+          queue: @queue,
+          producer: sync_producer,
+          delivery_threshold: delivery_threshold,
+          max_retries: max_retries,
+          retry_backoff: retry_backoff,
+          instrumenter: instrumenter,
+          logger: logger,
       )
 
       # The timer will no-op if the delivery interval is zero.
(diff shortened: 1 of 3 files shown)


gilibelz2:
There are randomly failed tests...

lib/kafka/client.rb
```diff
@@ -296,14 +296,16 @@ def producer(
...
-    def async_producer(delivery_interval: 0, delivery_threshold: 0, max_queue_size: 1000, **options)
+    def async_producer(delivery_interval: 0, delivery_threshold: 0, max_queue_size: 1000, max_retries: -1, retry_backoff: 0, **options)
       sync_producer = producer(**options)
 
       AsyncProducer.new(
         sync_producer: sync_producer,
         delivery_interval: delivery_interval,
         delivery_threshold: delivery_threshold,
         max_queue_size: max_queue_size,
+        max_retries: max_retries,
```
nguyenquangminh0711: Should we add these options into the sync producer too?

spec/async_producer_spec.rb
```diff
@@ -67,4 +69,26 @@ def instrument(name, payload = {})
       expect(metric.payload[:message_count]).to eq 42
     end
   end
+
+  describe "#produce" do
+    it "delivers buffered messages" do
```
nguyenquangminh0711: Should we have a test case to cover the retrying ability?

lib/kafka/client.rb
gilibelz2: We have `@max_retries` in the `sync_producer` (it will stop after all retries) but the `async_producer` (in the current version) still send `deliver_messages` in an endless loop to the `sync_producer`

spec/async_producer_spec.rb
gilibelz2: the second spec covers the retrying ability

lib/kafka/async_producer.rb
```diff
@@ -240,10 +244,23 @@ def run
...
-        retry
+        retries = 0
+        begin
+          @producer.produce(*args)
+        rescue BufferOverflow => e
+          deliver_messages
+          if @max_retries == -1
+            retry
+          end
+          if retries < @max_retries
```
dasch: Can you do `elsif` here instead, just to make the code a bit more legible?

gilibelz2: Sure. Thanks!

dasch:
Can you skip the "rubocop changes"? It makes the diff harder to follow. You can create a separate PR for that if you'd like.

gilibelz2:
Some of the CircleCI tests failed in `change to elsif` commit because of my local code formatter. The last commit was to move back to the formmat you have.

gilibelz2:
Is that ok?:) @dasch

gilibelz2:
I don't have merge permissions...

Issue #705, opened 2019-02-06
Support OAuthBearer Authentication


psycotica0-shopify:
With [KIP-255 OAuth Authentication via SASL/OAUTHBEARER](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876), Kafka now supports OAuth tokens as a method of authentication.

For us to use it on the brokers requires the library to support it, so I'm creating this issue here in case someone beats us to it (or it's already done somewhere I haven't found)

pt2pham:
I am currently working on adding OAuthBearer support!

PR #709, opened 2019-02-19
Support SASL OAuthBearer Authentication

pt2pham:
**Note:** _Please leave any comments/reviews on https://github.com/zendesk/ruby-kafka/pull/710 instead since that's forked from Shopify as opposed to my personal account!_

This PR adds support for OAuthBearer as a method of SASL Authentication as described in [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876) and https://github.com/zendesk/ruby-kafka/issues/705. 

When initializing the client, the user must pass in an instance of a class that implements a `token` method (the interface is described in `lib/kafka/sasl/oauth.rb`) which returns an ID Token. 

**Example:**

```ruby
require 'ruby-kafka'

class TokenProvider
    def token
        "some_id_token"
    end
end
# ...
client = Kafka.new(
    ["127.0.0.1:9094"],
    client_id: "phong-ruby-kafka",
    # ...
    sasl_over_ssl: false,
    sasl_oauth_token_provider: TokenProvider.new
)
```

Tested with a local Kafka cluster and ID Token, and successfully authenticated with the example client initialization above. I am open to suggestions for more cases to write unit tests for.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -62,13 +62,16 @@ module Kafka
     #
     # @param sasl_over_ssl [Boolean] whether to enforce SSL with SASL
     #
+    # @param sasl_oauth_token_provider [Object, nil] OAuthBearer Token Provider class that
+    #   implements method token. See {Sasl::OAuth#initialize}
+    #
     # @return [Client]
     def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_timeout: nil, socket_timeout: nil,
                    ssl_ca_cert_file_path: nil, ssl_ca_cert: nil, ssl_client_cert: nil, ssl_client_cert_key: nil,
                    ssl_client_cert_key_password: nil, ssl_client_cert_chain: nil, sasl_gssapi_principal: nil,
                    sasl_gssapi_keytab: nil, sasl_plain_authzid: '', sasl_plain_username: nil, sasl_plain_password: nil,
                    sasl_scram_username: nil, sasl_scram_password: nil, sasl_scram_mechanism: nil,
-                   sasl_over_ssl: true, ssl_ca_certs_from_system: false)
+                   sasl_over_ssl: true, ssl_ca_certs_from_system: false, sasl_oauth_token_provider: nil)
       @logger = TaggedLogger.new(logger)
       @instrumenter = Instrumenter.new(client_id: client_id)
       @seed_brokers = normalize_seed_brokers(seed_brokers)
@@ -92,6 +95,7 @@ module Kafka
         sasl_scram_username: sasl_scram_username,
         sasl_scram_password: sasl_scram_password,
         sasl_scram_mechanism: sasl_scram_mechanism,
+        sasl_oauth_token_provider: sasl_oauth_token_provider,
         logger: @logger
       )
 
--- a/lib/kafka/sasl/oauth.rb
+++ b/lib/kafka/sasl/oauth.rb
@@ -51,12 +51,12 @@ module Kafka
       private
 
       def initial_client_response
-        "n,,\x01auth=Bearer #{@token_provider.token}#{token_extensions}\x01\x01"#.force_encoding("utf-8")
+        "n,,\x01auth=Bearer #{@token_provider.token}#{token_extensions}\x01\x01"
       end
 
       def token_extensions
         return nil unless @token_provider.respond_to? :extensions
-        "\x01#{@token_provider.extensions.map{|e| e.join("=")}.join("\x01")}"
+        "\x01#{@token_provider.extensions.map {|e| e.join("=")}.join("\x01")}"
       end
     end
   end
--- a/lib/kafka/sasl/oauth.rb
+++ b/lib/kafka/sasl/oauth.rb
@@ -51,6 +51,7 @@ module Kafka
       private
 
       def initial_client_response
+        raise Kafka::NoTokenMethodError, "Token provider doesn't define 'token'" unless @token_provider.respond_to? :token
         "n,,\x01auth=Bearer #{@token_provider.token}#{token_extensions}\x01\x01"
       end
 
--- a/lib/kafka/sasl/oauth.rb
+++ b/lib/kafka/sasl/oauth.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+module Kafka
+  module Sasl
+    class OAuth
+      OAUTH_IDENT = "OAUTHBEARER"
+
+      # token_provider: THE FOLLOWING INTERFACE MUST BE FULFILLED:
+      #
+      # [REQUIRED] TokenProvider#token      - Returns an ID/Access Token to be sent to the Kafka client.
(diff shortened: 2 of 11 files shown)


pt2pham:
Currently just working on unit tests

lib/kafka/client.rb
```diff
@@ -62,13 +62,16 @@ class Client
...
+    #   implements method token. See {Sasl::OAuth#initialize}
+    #
     # @return [Client]
     def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_timeout: nil, socket_timeout: nil,
                    ssl_ca_cert_file_path: nil, ssl_ca_cert: nil, ssl_client_cert: nil, ssl_client_cert_key: nil,
                    ssl_client_cert_key_password: nil, ssl_client_cert_chain: nil, sasl_gssapi_principal: nil,
                    sasl_gssapi_keytab: nil, sasl_plain_authzid: '', sasl_plain_username: nil, sasl_plain_password: nil,
                    sasl_scram_username: nil, sasl_scram_password: nil, sasl_scram_mechanism: nil,
-                   sasl_over_ssl: true, ssl_ca_certs_from_system: false)
+                   sasl_over_ssl: true, ssl_ca_certs_from_system: false, sasl_oauth_token_provider: nil)
```
dorner: Man this constructor makes me sad now :( I feel like all the SSL stuff should be extracted into some kind of configuration object. Not necessary for this PR but god this is hard to read.

lib/kafka/sasl/oauth.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module Kafka
+  module Sasl
+    class OAuth
+      OAUTH_IDENT = "OAUTHBEARER"
+
+      # token_provider: THE FOLLOWING INTERFACE MUST BE FULFILLED:
+      #
+      # [REQUIRED] TokenProvider#token      - Returns an ID/Access Token to be sent to the Kafka client.
```
mensfeld: you seem to use yard in the previous section but here it seems like a custom format `[REQUIRED]`. Is that intended?

lib/kafka/sasl/oauth.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      def ident
+        OAUTH_IDENT
+      end
+
+      def configured?
+        @token_provider
+      end
+
+      def authenticate!(host, encoder, decoder)
```
mensfeld: maybe it would be good to document this method since it's public

lib/kafka/sasl/oauth.rb
```diff
@@ -0,0 +1,64 @@
...
+        @logger.debug "Authenticating to #{host} with SASL #{OAUTH_IDENT}"
+
+        encoder.write_bytes(initial_client_response)
+
+        begin
+          # receive SASL OAuthBearer Server Response
+          msg = decoder.bytes
+          raise Kafka::Error, "SASL #{OAUTH_IDENT} authentication failed: unknown error" unless msg
+        rescue Errno::ETIMEDOUT, EOFError => e
+          raise Kafka::Error, "SASL #{OAUTH_IDENT} authentication failed: #{e.message}"
```
mensfeld: shouldn't we have a custom error for that?

lib/kafka/sasl/oauth.rb
```diff
@@ -0,0 +1,64 @@
...
+      def authenticate!(host, encoder, decoder)
+        # Send SASLOauthBearerClientResponse with token
+        @logger.debug "Authenticating to #{host} with SASL #{OAUTH_IDENT}"
+
+        encoder.write_bytes(initial_client_response)
+
+        begin
+          # receive SASL OAuthBearer Server Response
+          msg = decoder.bytes
+          raise Kafka::Error, "SASL #{OAUTH_IDENT} authentication failed: unknown error" unless msg
```
mensfeld: shouldn't we have a custom error for that?

lib/kafka/sasl_authenticator.rb
```diff
@@ -31,7 +33,12 @@ def initialize(logger:, sasl_gssapi_principal:, sasl_gssapi_keytab:,
         logger: @logger,
       )
 
-      @mechanism = [@gssapi, @plain, @scram].find(&:configured?)
+      @oauth = Sasl::OAuth.new(
+        token_provider: sasl_oauth_token_provider,
+        logger: @logger,
```
mensfeld: trailing coma not needed

lib/kafka/client.rb
sam-obeid: Agree, an SSL config object/builder would be ideal!

spec/fake_token_provider.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class FakeTokenProvider
+  def initialize
```
mensfeld: this could be skipped

spec/fake_token_provider.rb
```diff
@@ -0,0 +1,19 @@
...
+    "SASLOAUTHBEARER.TEST_ID_TOKEN"
+  end
+
+  def extensions
+    { test_key: "test_value", test_key_2: "test_value_2" }
+  end
+end
+
+class FakeBrokenTokenProvider
+  def initialize
```
mensfeld: not needed

lib/kafka/client.rb
mensfeld: All of the settings should be a value object :(

lib/kafka/sasl_authenticator.rb
pt2pham: I tried to keep the same style as the other constructors with the trailing commas! In general I believe these are just to minimize git commit lines if another parameter is added here.

lib/kafka/sasl/oauth.rb
pt2pham: Currently, all of the classes in `Sasl` just raise a `Kafka::Error` with a message, rather than a custom error. Maybe this refactor could be done in a separate PR?

lib/kafka/sasl/oauth.rb
pt2pham: I tried to keep the same style as the other `Sasl` classes. All of them do the same thing which is get called by `Kafka::SaslAuthenticator#authenticate!` depending on which mechanism is configured.

lib/kafka/sasl/oauth.rb
pt2pham: I didn't use yard since in this case, I'm just adding documentation for the kind of class that the client/user will need to write in order to use OAuthBearer. It doesn't necessarily apply to this file but it's the most appropriate spot to have it. 

Documentation for how to use this should probably also be in the README.

lib/kafka/client.rb
pt2pham: Yeah I was thinking the same thing when I was writing this PR.

pt2pham:
https://github.com/zendesk/ruby-kafka/pull/710

Please add any other comments to this PR instead since it's forked from Shopify as opposed to my personal fork!

PR #710, opened 2019-02-19
Support SASL OAuthBearer Authentication

pt2pham:
**Note:** _This is the same as this PR: (https://github.com/zendesk/ruby-kafka/pull/709). This one is just forked from Shopify as opposed to my personal fork. All the changes here are the last PR + any PR feedback that was left before @sam-obeid said "LGTM"!_ 

This PR adds support for OAuthBearer as a method of SASL Authentication as described in [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876) and https://github.com/zendesk/ruby-kafka/issues/705. 

When initializing the client, the user must pass in an instance of a class that implements a `token` method (the interface is described in `lib/kafka/sasl/oauth.rb`) which returns an ID Token. 

**Example:**

```ruby
require 'ruby-kafka'

class TokenProvider
    def token
        "some_id_token"
    end
end
# ...
client = Kafka.new(
    ["127.0.0.1:9094"],
    client_id: "phong-ruby-kafka",
    # ...
    sasl_over_ssl: false,
    sasl_oauth_token_provider: TokenProvider.new
)
```

Tested with a local Kafka cluster and ID Token, and successfully authenticated with the example client initialization above. I am open to suggestions for more cases to write unit tests for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -988,6 +988,26 @@ kafka = Kafka.new(
 )
 ```
 
+##### OAUTHBEARER
+This mechanism is supported in kafka >= 2.0.0 as of [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876)
+
+In order to authenticate using OAUTHBEARER, you must set the client with an instance of a class that implements a `token` method (the interface is described in `lib/kafka/sasl/oauth.rb`) which returns an ID/Access token.
+
+Optionally, the client may implement an `extensions` method that returns a map of key-value pairs. These can be sent with the SASL/OAUTHBEARER initial client response. This is only supported in kafka >= 2.1.0.
+
+```ruby
+class TokenProvider
+    def token
+        "some_id_token"
+    end
+end
+# ...
+client = Kafka.new(
+    ["kafka1:9092"],
+    sasl_oauth_token_provider: TokenProvider.new
+)
+```
+
 ### Topic management
 
 In addition to producing and consuming messages, ruby-kafka supports managing Kafka topics and their configurations. See [the Kafka documentation](https://kafka.apache.org/documentation/#topicconfigs) for a full list of topic configuration keys.
--- a/README.md
+++ b/README.md
@@ -991,7 +991,7 @@ kafka = Kafka.new(
 ##### OAUTHBEARER
 This mechanism is supported in kafka >= 2.0.0 as of [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876)
 
-In order to authenticate using OAUTHBEARER, you must set the client with an instance of a class that implements a `token` method (the interface is described in `lib/kafka/sasl/oauth.rb`) which returns an ID/Access token.
+In order to authenticate using OAUTHBEARER, you must set the client with an instance of a class that implements a `token` method (the interface is described in [Kafka::Sasl::OAuth](lib/kafka/sasl/oauth.rb)) which returns an ID/Access token.
 
 Optionally, the client may implement an `extensions` method that returns a map of key-value pairs. These can be sent with the SASL/OAUTHBEARER initial client response. This is only supported in kafka >= 2.1.0.
 
--- a/README.md
+++ b/README.md
@@ -997,14 +997,14 @@ Optionally, the client may implement an `extensions` method that returns a map o
 
 ```ruby
 class TokenProvider
-    def token
-        "some_id_token"
-    end
+  def token
+    "some_id_token"
+  end
 end
 # ...
 client = Kafka.new(
-    ["kafka1:9092"],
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -988,6 +988,26 @@ kafka = Kafka.new(
 )
 ```
 
+##### OAUTHBEARER
+This mechanism is supported in kafka >= 2.0.0 as of [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876)
+
+In order to authenticate using OAUTHBEARER, you must set the client with an instance of a class that implements a `token` method (the interface is described in `lib/kafka/sasl/oauth.rb`) which returns an ID/Access token.
```
psycotica0-shopify: I'm not an expert on GitHub Readme Markdown, but I assume there's some way to make this a relative link to the file, so the reader can just click it, rather than finding it themselves.

lib/kafka/sasl/oauth.rb
```diff
@@ -0,0 +1,64 @@
...
+      def authenticate!(host, encoder, decoder)
+        # Send SASLOauthBearerClientResponse with token
+        @logger.debug "Authenticating to #{host} with SASL #{OAUTH_IDENT}"
+
+        encoder.write_bytes(initial_client_response)
+
+        begin
+          # receive SASL OAuthBearer Server Response
+          msg = decoder.bytes
+          raise Kafka::Error, "SASL #{OAUTH_IDENT} authentication failed: unknown error" unless msg
```
psycotica0-shopify: Reading through the broker code, it looks like the `msg` they return is meant to be a JSON object describing the error.
I wonder if it's worth putting the object either into the exception object, or logging it, or something?

I don't think we can interpret the failure, but perhaps the client author could?

pt2pham: `decoder.bytes` will raise an error and crash if we get a bad response from Kafka and it will just log a general error in the `rescue` block. 

I'm guessing it assumes that if we receive a msg at all then Kafka succeeded, otherwise it'll internally decide if the response is bad. 

It probably does make sense to get the actual response from Kafka and log a more meaningful error, since Kafka will return a proper code, but that's probably for another PR.

psycotica0-shopify: That's alright by me. It looks like all of the SASL implementations are making a similar assumption, so 🤷‍♂️

pt2pham: https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes 

This might be useful for a future PR?

README.md
```diff
@@ -988,6 +988,26 @@ kafka = Kafka.new(
...
 
+##### OAUTHBEARER
+This mechanism is supported in kafka >= 2.0.0 as of [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876)
+
+In order to authenticate using OAUTHBEARER, you must set the client with an instance of a class that implements a `token` method (the interface is described in [Kafka::Sasl::OAuth](lib/kafka/sasl/oauth.rb)) which returns an ID/Access token.
+
+Optionally, the client may implement an `extensions` method that returns a map of key-value pairs. These can be sent with the SASL/OAUTHBEARER initial client response. This is only supported in kafka >= 2.1.0.
+
+```ruby
+class TokenProvider
```
bai: Style mnitpick: 2 spaces indentation in ruby here and in other examples.

spec/sasl_authenticator_spec.rb
```diff
@@ -91,4 +93,24 @@
       }.to raise_error(Kafka::FailedScramAuthentication)
     end
   end
+
+  context "when SASL OAuthBearer has been configured" do
```
nguyenquangminh0711: I think there should be a test case for the missing of extensions here.

pt2pham: Strange. I added a test case so that I pass in a TokenProvider instance that doesn't have `extensions` defined. I ran `rspec` and that test failed with a TIMEOUT error during Kafka authentication. 

However, when I use my ruby-kafka fork with a test-client I wrote that authenticates to a local Kafka cluster, I pass in a token provider that doesn't have `extensions` defined, and it works fine.

pt2pham:
What other steps are there before getting this PR merged? 

I ran tests with a test client that I wrote using my own TokenProvider against brokers in our staging environment and it was able to successfully authenticate and hold a connection!

nguyenquangminh0711:
Sorry, usually, after we approve, you can merge by yourself, but perhaps something changes. I already merge the PR for you 🙆‍♂️

PR #712, opened 2019-02-22
Add missing documentation to each_batch method

sam-obeid:
Add missing documentation to `Consumer::each_batch` method.

Both methods `each_message` and `each_batch` raise `Kafka::ProcessingError` exception. But only the former has that declared in documentation.
There is already a unit test for method `each_batch` raising `Kafka::ProcessingError` exception.
https://github.com/zendesk/ruby-kafka/blob/master/spec/consumer_spec.rb#L439-L453


dasch:
Thanks!

PR #714, opened 2019-03-18
Change asyn_producer method arguments name

gilibelz2:
In PR (already merged) https://github.com/zendesk/ruby-kafka/pull/708, I added 2 new arguments: `max_retries` and `retry_backoff` to `async_producer` method in `client.rb`. It seems that the internal `sync_producer` doesn't get those params because they have the same names (like in the `AsyncProducer`) and it takes the default (`max_retries: 2, retry_backoff: 1`). I changed the params name in the `async_producer` to fix that.
### _This will break the current API_

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -300,15 +300,15 @@ module Kafka
     #
     # @see AsyncProducer
     # @return [AsyncProducer]
-    def async_producer(delivery_interval: 0, delivery_threshold: 0, max_queue_size: 1000, max_retries: -1, retry_backoff: 0, **options)
+    def async_producer(delivery_interval: 0, delivery_threshold: 0, ap_max_queue_size: 1000, ap_max_retries: -1, retry_backoff: 0, **options)
       sync_producer = producer(**options)
 
       AsyncProducer.new(
         sync_producer: sync_producer,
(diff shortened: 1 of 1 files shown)


mensfeld:
Worth pointing out, that this breaks the current api.

gilibelz2:
Agree. I'll point it out. Thanks.

mensfeld:
Before accepting and merging that, can we coordinate? It would be good to sync with delivery boy and waterdrop releases.

gilibelz2:
Sure. What do you want to coordinate?

mensfeld:
@gilibelz2 I could bump and fixs deps in delivery boy, waterdrop and karafka for that so @dasch if this passed by, please ping me before the release and we can sync.

gilibelz2:
O.k. great, so what should I do to promote this?:)

dasch:
Breaking the current API is not the way to go.

Is there a need for these configs to be defined separately for the async and inner sync producer? Can we just reuse the same values for both? We're really just talking about two "layers" of a single logical operation.

gilibelz2:
? @dasch @mensfeld

mensfeld:
@gilibelz2 I share the same opinion as @dasch

PR #716, opened 2019-03-21
Refactor tagged logging

drcapulet:
PR #697 added support for tagged logging, but because it modifies the existing formatter any subsequent changes to the formatter cause exceptions at runtime. This instead wraps an underlying logger and duplicates the logic from the core `Logger` class to prepend the tags text when needed.

Also, could the changelog for 0.7.6 be updated to include PR #697?

### DIFF
--- a/lib/kafka/tagged_logger.rb
+++ b/lib/kafka/tagged_logger.rb
@@ -4,10 +4,64 @@ require 'forwardable'
 # ActiveSupport::TaggedLogging.
 
 module Kafka
-  module TaggedFormatter
+  class TaggedLogger
+    def self.new(logger)
+      return logger if logger.is_a?(Kafka::TaggedLogger)
+      return super(logger)
+    end
 
-    def call(severity, timestamp, progname, msg)
-      super(severity, timestamp, progname, "#{tags_text}#{msg}")
+    def initialize(logger)
+      @logger = logger || Logger.new(nil)
+    end
+
+    def add(severity, message = nil, progname = nil)
+      return true if (severity || ::Logger::UNKNOWN) < @logger.level
+
+      if message.nil?
+        if block_given?
+          message = yield
+        else
+          message = progname
+          progname = nil
+        end
+      end
+
+      @logger.add(severity, "#{tags_text}#{message}", progname)
+    end
+
+    def clear_tags!
+      current_tags.clear
+    end
+
+    def debug(progname = nil, &block)
+      add(::Logger::DEBUG, nil, progname, &block)
+    end
+
+    def info(progname = nil, &block)
+      add(::Logger::INFO, nil, progname, &block)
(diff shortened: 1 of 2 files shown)


dasch:
@dorner all this logging business is leading to a lot of maintenance problems. Perhaps we should just revert the original PR?

dorner:
@dasch hm... obviously if it's causing crashes then it needs to be backed out for now. I would like to try to reintroduce it in another fashion. If this refactor fixes the other issues I'm fine with it. Another method might be to go back to my original attempt at doing this, which was to try to detect if ActiveSupport::TaggedLogging is in use and if so, to just use it straight. I actually think that method is the least disruptive. 

But yeah, in any case, we can definitely revert for now to fix the immediate bugs. We can use this PR to reopen the conversation if you're OK with that. @drcapulet too ⬆️

dasch:
I think that’s probably the best course of action right now.

dorner:
Opened #729 to revert the changes. 

@drcapulet @dasch The sheer openness of Ruby logging frameworks means that trying to wrap a logger isn't quite as straightforward as I thought it was. My original idea with the tagged logging was just to detect if ActiveSupport::TaggedLogging was in use, and if so, to use its `push_tags` and `pop_tags` methods. I feel like it could be treated as an optional dependency which would help provide more information.

Do you think it's still worth it to try to find a better way to wrap the provided logger? Or are we OK to say that we'll just use duck typing for `push_tags` and `pop_tags` and use those features if they exist?

dasch:
> Opened #729 to revert the changes.
> 
> @drcapulet @dasch The sheer openness of Ruby logging frameworks means that trying to wrap a logger isn't quite as straightforward as I thought it was. My original idea with the tagged logging was just to detect if ActiveSupport::TaggedLogging was in use, and if so, to use its `push_tags` and `pop_tags` methods. I feel like it could be treated as an optional dependency which would help provide more information.
> 
> Do you think it's still worth it to try to find a better way to wrap the provided logger? Or are we OK to say that we'll just use duck typing for `push_tags` and `pop_tags` and use those features if they exist?

Honestly, I worry that we'll continually run into issues if we try to be clever in a library that's supposed to be portable. I think a better approach would be a wrapper that just delegates to the underlying Logger, but does its own formatting.

dorner:
TBH the problem with wrapping loggers (without asking at least) is that there are so many different patterns that I'd have a hard time being confident that what we do isn't breaking something somewhere. :( The advantage of the duck-typing solution is that it's very specific and doesn't try to do anything fancy.

But let's definitely give the class implementation a try and see if we run into issues. If we do and we can't figure them out we can always fall back to my original design :)

dasch:
As long as we only call standard Logger methods (eg `info`) then I think we’ll be safe.

PR #722, opened 2019-04-04
[Improvement] Allow sleep time when no batch to process to be configured

quangduc:
Currently, when we fetch_batch from consumer, if batch has no data, we will sleep for 2 seconds before returning. It's better if we can let this fixed number(2) to be configured, because some consumer needs to receive and handle events asap.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -340,7 +340,8 @@ module Kafka
         offset_commit_threshold: 0,
         heartbeat_interval: 10,
         offset_retention_time: nil,
-        fetcher_max_queue_size: 100
+        fetcher_max_queue_size: 100,
+        sleep_time_when_no_message: 2
     )
       cluster = initialize_cluster
 
@@ -393,6 +394,7 @@ module Kafka
         fetcher: fetcher,
         session_timeout: session_timeout,
         heartbeat: heartbeat,
+        sleep_time_when_no_message: sleep_time_when_no_message
       )
     end
 
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -44,7 +44,7 @@ module Kafka
   #
   class Consumer
 
-    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:, sleep_time_when_no_message:)
+    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:, sleep_time_when_no_message: 2)
       @cluster = cluster
       @logger = TaggedLogger.new(logger)
       @instrumenter = instrumenter
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -44,7 +44,7 @@ module Kafka
   #
   class Consumer
 
-    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:)
+    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:, sleep_time_when_no_message:)
       @cluster = cluster
       @logger = TaggedLogger.new(logger)
       @instrumenter = instrumenter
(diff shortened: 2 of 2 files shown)


dasch:
What is your `max_wait_time`? Would it be possible to reuse that value?

quangduc:
My ```max_wait_time```is 0.5. But I think we still need to separate those 2 values because of their difference in meaning. What do you think?

dasch:
I'd rather start by not introducing more config, then we can see if we need to keep them separate. I think `max_wait_time` serves roughly the same purpose as a bound on the sleep time – it sets an upper bound on read latency.

quangduc:
Then, in your code, instead of alway sleeping for 2 seconds, you should sleep ``max_wait_time``.  2 seconds sleep time is too much for some low-latency application

jaredonline:
👋 Heya @dasch and @quangduc. We're running into this problem in production as well. We have a low throughput, latency sensitive use case, which means when there's no message, sleeping for 2 seconds absolutely kills the viability of this client for us.

Re-using `max_wait_time` would give us a way to configure it, but feels like it overloads the meaning of that configuration, and may have unexpected consequences for some clients.

Introducing a new config would suit us very well.

Either way, we'd love to be able to help push this forward if there's anything we can do. Thanks!

quangduc:
> 👋 Heya @dasch and @quangduc. We're running into this problem in production as well. We have a low throughput, latency sensitive use case, which means when there's no message, sleeping for 2 seconds absolutely kills the viability of this client for us.
> 
> Re-using `max_wait_time` would give us a way to configure it, but feels like it overloads the meaning of that configuration, and may have unexpected consequences for some clients.
> 
> Introducing a new config would suit us very well.
> 
> Either way, we'd love to be able to help push this forward if there's anything we can do. Thanks!

I can't agree with you more. Anyway, I can't wait for them to accept my PR, so I'm using my fork in production. But it's quite disappointed when they didn't listen to our feedback and recognize the issue

mensfeld:
> Introducing a new config would suit us very well.

Same here.

BenEddy:
Agreed with @dasch that introducing another configuration isn't ideal.

- The `sleep_time_when_no_message` config has no analog in other kafka clients (AFAIK) and controls fairly specific ruby-kafka internals.
- `sleep_time_when_no_message` feels like an advanced config that users will find after debugging consumer latency. It would be great if the consumer was fast by default.
- Setting `sleep_time_when_no_message` to `0` puts the consumer into a tight `consumer_loop`. When I tried `0` locally, the `consumer_loop` thread seemed to hold onto the GIL and starve IO threads, significantly increasing consumer latency.

But to echo @quangduc and @jaredonline, sleeping is certainly a problem for low-latency consumers of low-volume topics. Overloading `max_wait_time` isn't a perfect solution because it doubles max wait time in the worst case and because users will presumably want to configure sleep time much, much lower than `max_wait_time` -- as close to 0 as possible.

@dasch (or anyone else) do you happen to recall the intent of the `!@fetcher.data?` check? I'm curious what would happen if that clause was dropped in favor of simply `@fetcher.poll` (and indirectly a blocking `Queue#deq`).

```Diff
- if !@fetcher.data?
-   @logger.debug "No batches to process"
-   sleep 2
-   []
- else
    tag, message = @fetcher.poll
- end
```

dasch:
What value would you configure the sleep to? Can we hardcode something reasonable?

mensfeld:
>  thread seemed to hold onto the GIL and starve IO threads, significantly increasing consumer latency.

There is one more thing we could use: Ruby thread priority. Since almost no one runs ruby-kafka together with other things in a single VM, setting the priority could get us much closer to what we're aiming with this PR.

dasch:
@mensfeld can you share an example of that?

mensfeld:
> thread seemed to hold onto the GIL and starve IO threads, significantly increasing consumer latency.

I did some work related to starving threads with IO operations for a different project. 

From Ruby docs (MRI):

> Higher-priority threads will run more frequently than lower-priority threads.

This can be used to "force" the other threads to get more power than the one that is starving the rest. 

```ruby
count1 = count2 = 0
a = Thread.new do
      loop { count1 += 1 }
    end
a.priority = -1

b = Thread.new do
      loop { count2 += 1 }
    end
b.priority = -2
sleep 1   #=> 1
count1    #=> 622504
count2    #=> 5832
```

If we would provide an API to prioritize/deprioritize the ruby-kafka threads, than it could be balanced.

PR #724, opened 2019-04-09
Support zstd compression

paulczajka:
Kafka [now supports](https://cwiki.apache.org/confluence/display/KAFKA/KIP-110%3A+Add+Codec+for+ZStandard+Compression) zstd compression with its `2.1` release.  This does a little cleanup of codec-related code, then adds zstd compression support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,7 @@ Compression is enabled by passing the `compression_codec` parameter to `#produce
 * `:snappy` for [Snappy](http://google.github.io/snappy/) compression.
 * `:gzip` for [gzip](https://en.wikipedia.org/wiki/Gzip) compression.
 * `:lz4` for [LZ4](https://en.wikipedia.org/wiki/LZ4_(compression_algorithm)) compression.
+* `:zstd` for [zstd](https://facebook.github.io/zstd/) compression.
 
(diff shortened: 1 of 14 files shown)


dasch:
I've tried restarting the build, but it still fails on your new test. Can you take a look?

paulczajka:
@dasch I fixed the 2.1 issue; appears that [zstd requires an api version bump](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L117-L120).  2.0 runs fine locally - can you kick the build?

dasch:
in that case, can you use this method to skip the test when needed? https://github.com/zendesk/ruby-kafka/blob/02f7e2816e1130c5202764c275e36837f57ca4af/lib/kafka/client.rb#L702

Example: https://github.com/zendesk/ruby-kafka/blob/02f7e2816e1130c5202764c275e36837f57ca4af/spec/functional/topic_management_spec.rb#L16

paulczajka:
Fixed! Confirmed all other codecs were [present by 0.8.2.0](https://issues.apache.org/jira/browse/KAFKA-1456) on [PRODUCE_API V0](https://github.com/apache/kafka/blob/0.8.2.0/clients/src/main/java/org/apache/kafka/common/protocol/Protocol.java#L82)

dasch:
Thanks!

PR #729, opened 2019-05-01
Revert #697 (93c45a8c531e8b447f1cd9127c8bc6ce8070c0e6) - Add tagged l…

dorner:
…ogging
See discussion in #716. We'll use that PR for discussion.


dorner:
Took a different approach in #732.

PR #732, opened 2019-05-10
Change TaggedLogger to use a class implementation

dorner:
See discussion in #716. This is an attempt at turning this into a separate class rather than modifying the underlying logger. Hopefully this will fix some of the issues we've been seeing.
@dasch


dorner:
Nice, thanks!

PR #733, opened 2019-05-14
Optionally verify hostname on SSL certs

rafaelpetry:
The change introduced in #730 caused our application to not be able to connect to Kafka in Heroku. It looks like Heroku client certs do not have a matching hostname. This change allows the `ssl_verify_hostname` option to be passed to the `Kafka::Client`, so it can be overwritten if necessary.

/cc @kyleholzinger


rafaelpetry:
Quick update... Along with this PR, we opened a support ticket with Heroku to see if they could address it on their side. Doesn't look like that will happen anytime soon. This was their response:

> Hello,
> 
> While we have noted this as a feature request - we need to balance it against other feature request and at this time we are not able to make this change. We would suggest using a version of the library that does not require this hostname verification until the change can be merged.

kyleholzinger:
@dasch I saw you merged the PR that introduced this behavior, have any thoughts on allowing users to configure this option?

PR #737, opened 2019-05-28
Block on Fetcher#poll in Consumer#fetch_batches

BenEddy:
As detailed in https://github.com/zendesk/ruby-kafka/pull/722, [`sleep 2` in `Consumer#fetch_batches`](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L528) presents a problem for latency-sensitive consumers. This PR removes the `Fetcher#data?` check and `sleep` in favor of just calling `Fetcher#poll`, which calls the blocking `Queue#deq`. `Fetcher` pushes all batches, even empty batches, onto the `Queue`, so it seems like `Queue#deq` shouldn't block for much longer than the duration of a fetch request, but maybe I'm missing something.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -523,21 +523,16 @@ module Kafka
 
       resume_paused_partitions!
 
-      if !@fetcher.data?
-        @logger.debug "No batches to process"
-        sleep 2
-        []
-      else
-        tag, message = @fetcher.poll
+      tag, message = @fetcher.poll
 
-        case tag
-        when :batches
-          # make sure any old batches, fetched prior to the completion of a consumer group sync,
-          # are only processed if the batches are from brokers for which this broker is still responsible.
-          message.select { |batch| @group.assigned_to?(batch.topic, batch.partition) }
-        when :exception
-          raise message
-        end
+      case tag
+      when :batches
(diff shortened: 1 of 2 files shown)


dasch:
The problem is that the consumer needs to be able to respond to events that don't come from the fetcher, and be able to gracefully handle situations in which the fetcher dies. We'd need a timeout on the `poll` for that reason.

dasch:
#125 actually includes an example of a blocking queue with a timeout on `deq`.

BenEddy:
@dasch Gotcha, makes sense 👍. Would you be opposed to adopting the blocking queue implementation from #125 in the `Fetcher`? A less involved but less elegant alternative might be to periodically push no-op messages through the `Queue` if `poll` has been blocking for more than n seconds. Roughly:

```Ruby
class Fetcher
  def start
    Thread.new do
      loop do
        sleep 1
        queue.push([:noop]) if queue.empty? && Time.now - @last_poll_at > 2
      end
    end

    # ...
  end

  def poll
    @last_poll_at = Time.now
    tag, message, reset_counter = @queue.deq
    return :batches, [] if tag == :noop

    # ...
  end
end
```

dasch:
I think using the blocking queue would be preferable, but we should evaluate the performance impact.

Issue #736, opened 2019-05-22
Invalid server signature for 1'st try


fist28:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.6.3
* Version of Kafka:  2.0.0
* Version of ruby-kafka: 0.7.7
* Version of avro_turf: 0.8.1

###### Steps to reproduce

```ruby
require "avro_turf"
require 'avro_turf/messaging'
require "kafka"

avro = AvroTurf::Messaging.new(registry_url: "https://schemaregistry-app.example.com")
kafka = Kafka.new(
    %w[kafka://kafka-app.example.com:9092],
    sasl_scram_mechanism: 'sha512',
    sasl_scram_username: 'MyAppUserName',
    sasl_scram_password: 'MyAppUserPassword',
    sasl_over_ssl: false
   )
consumer = kafka.consumer(group_id: "MP_APP_CONSUMER_GROUP")
consumer.subscribe("my_topic")
consumer.each_message do |message|
  greeting = avro.decode(message.value)
  puts greeting['documentId']
end
```
I got exception:
```ruby
Traceback (most recent call last):
       16: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/consumer_group.rb:47:in `join'
       15: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/consumer_group.rb:139:in `join_group'
       14: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/instrumenter.rb:35:in `instrument'
       13: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/instrumenter.rb:23:in `instrument'
       12: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/consumer_group.rb:140:in `block in join_group'
       11: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/consumer_group.rb:199:in `coordinator'
       10: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/cluster.rb:121:in `get_group_coordinator'
        9: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/cluster.rb:416:in `get_coordinator'
        8: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/cluster.rb:416:in `each'
        7: from /bundle/gems/ruby-kafka-0.7.7/lib/kafka/cluster.rb:419:in `block in get_coordinator'
[...]

PR #743, opened 2019-06-10
Scram authentication thread save

fist28:
Fix for issue #736 

Invalid server signature was raise because client mixed server_signature between threads.


dasch:
Thanks!

Issue #745, opened 2019-06-13
MessageSet with messages prior to requested offset raises "undefined method `topic' for nil:NilClass (NoMethodError)"


brettburley:
* Version of Ruby: 2.4.5
* Version of Kafka: 0.11.0.1
* Version of ruby-kafka: 0.7.8

###### Steps to reproduce
Fetch a batch that returns with a `MessageSet` with records that have offsets that are prior to the requested offset.

```ruby
fetched_partition = Kafka::Protocol::FetchResponse::FetchedPartition.new(
  partition: 0,
  error_code: 0,
  highwater_mark_offset: 1,
  last_stable_offset: 1,
  aborted_transactions: [],
  messages: [
    Kafka::Protocol::MessageSet.new(
      messages: [
        Kafka::Protocol::Record.new(
          offset: 7,
          value: '1234'
        ),
        Kafka::Protocol::Record.new(
          offset: 8,
          value: '1234'
        )
      ]
    ),
    Kafka::Protocol::MessageSet.new(
      messages: [
        Kafka::Protocol::Record.new(
          offset: 9,
          value: '1234'
        ),
        Kafka::Protocol::Record.new(
          offset: 10,
          value: '1234'
        )
      ]
    ),
    Kafka::Protocol::MessageSet.new(
      messages: [
        Kafka::Protocol::Record.new(
          offset: 11,
          value: '1234'
        )
      ]
    )
  ]
)

generator = Kafka::FetchedBatchGenerator.new('Hello', fetched_partition, 10, logger: nil)
result = generator.generate
```

###### Expected outcome
The response should return only the two messages with offsets equal to or greater than the requested offset of `10`. The array length should be two.
```
[
  #<Kafka::FetchedMessage @offset=10>,
  #<Kafka::FetchedMessage @offset=11>
]
```

###### Actual outcome
The array length is 5, with nil entries for the unexpected messages:
```
[
  nil,
  nil,
  nil,
  #<Kafka::FetchedMessage @offset=10>,
  #<Kafka::FetchedMessage @offset=11>
]
```

This results in this exception being thrown when a consumer is running on such a batch:
```
ruby-kafka-0.7.8/lib/kafka/consumer.rb:217:in `block (3 levels) in each_message': undefined method `topic' for nil:NilClass (NoMethodError)
from ruby-kafka-0.7.8/lib/kafka/consumer.rb:215:in `each'
[...]

PR #746, opened 2019-06-13
Do not add nil message entries when ignoring old messages.

brettburley:
When using MessageSets, we want to remove messages when their offset is
smaller than the requested offset.

It appears that this was handled properly for the `RecordBatch` uses cases, but not for the `MessageSet` path.

This is intended to address #745


dasch:
Thanks!

Issue #756, opened 2019-07-15
Custom logger can no longer be used in 0.7.9.


brettburley:
* Version of Ruby: 2.4.6
* Version of ruby-kafka: 0.7.9

###### Steps to reproduce

```ruby
kafka = Kafka.new(['localhost:9092'], logger: MyLogger.new)

TypeError: no implicit conversion of MyLogger into String
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/open-uri.rb:37:in `open'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/open-uri.rb:37:in `open'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/logger.rb:744:in `open_logfile'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/logger.rb:736:in `set_dev'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/logger.rb:671:in `initialize'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/logger.rb:387:in `new'
/usr/local/rvm/rubies/ruby-2.4.6/lib/ruby/2.4.0/logger.rb:387:in `initialize'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka/tagged_logger.rb:63:in `new'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka/tagged_logger.rb:63:in `initialize'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka/tagged_logger.rb:56:in `new'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka/tagged_logger.rb:56:in `new'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka/client.rb:75:in `initialize'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka.rb:366:in `new'
~/.bundle_local/ruby/2.4.0/gems/ruby-kafka-0.7.9/lib/kafka.rb:366:in `new'
```

###### Expected outcome

A custom logger class can be used.

###### Actual outcome

[...]

PR #762, opened 2019-07-26
#756: Fix logger again

dorner:
This should enable custom loggers again. Tested with Logging framework.

### DIFF
--- a/lib/kafka/tagged_logger.rb
+++ b/lib/kafka/tagged_logger.rb
@@ -6,7 +6,7 @@ require 'logger'
 module Kafka
   class TaggedLogger < SimpleDelegator
 
-    %i(debug info warn error fatal).each do |method|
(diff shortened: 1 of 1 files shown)


dorner:
@dasch

dasch:
I’m fine with this as a short term fix, but I’m starting to wonder if the improved logging is worth the maintenance burden and user frustration at this point.

dorner:
You can place the blame firmly on my plate. The initial commit missed a huge edge case and the fix wasn't tested correctly since I was mainly using my own use cases for testing. I'm still getting used to the difference between open source and standard company workflow (with manual QA, for example). :( We can release this as a beta version if you like and we can ask people to test them out - I'm learning all about how various frameworks just decide to act *mostly* like a logger but not *fully* like one -_-

Issue #721, opened 2019-04-02
timestamp_delta & create_time


RD-Mobile-Admin:
* Version of Ruby: 2.5.3
* Version of Kafka: 
* Version of ruby-kafka: 0.7.6

Some messages in a topic have, for some reason, create time that is well into the future. While digging into this issue, I've noticed there seems to be a correlation between that and the timestamp_delta.
In short, 'good' messages have a timestamp_delta==0, while the messages from the future have a timestamp_delta that appears to be equal to the number of seconds in the future.
Fetching the timestamps with kafka-console-consumer returns timestamps as expected.

Any thoughts on why this difference between timestamps happen and how to deal with it (other than customizing the gem)?

>mm= k.fetch_messages(topic:topic,partition:partition,offset: :earliest)
> 
> #<Kafka::FetchedMessage:0x0000559f96995eb8
>  @message=
>   #<Kafka::Protocol::Record:0x0000559f969a4d28
>    @attributes=0,
>    @bytesize=153,
>    @create_time=2021-01-02 21:02:27 +0000,
>    @headers={},
>    @is_control_record=false,
>    @key="81888377-c0bf-48b9-9e45-f9bcac1e1e25",
>    @offset=1,
>    @offset_delta=1,
>    @timestamp_delta=73927219,
> 
> 
> kafka-console-consumer --topic topic --bootstrap-server server:9092 --offset 0 --partition 81 --property print.timestamp=true
> 
> CreateTime:1535768055634  81888377-c0bf-48b9-9e45-f9bcac1e1e25

 DateTime.strptime("1535768055",'%s')
 => Sat, 01 Sep 2018 02:14:15 +0000

(mm[1].create_time - 73927219)
 => 2018-08-31 05:42:08 +0000

Elyahou:
What is the version of kafka ?

RD-Mobile-Admin:
> What is the version of kafka ?

2.0.0 (using confluent 5.0.0)

Elyahou:
I had exactly the same issue with precedent ruby-kafka version and fixed it with this: 

https://github.com/zendesk/ruby-kafka/pull/706

Can you maybe debug it  further and check what is the `first_timestamp` of the record batch in this case ?

RD-Mobile-Admin:
[...]

PR #766, opened 2019-08-05
Correctly handle timestamp deltas as milliseconds

stevenjm:
Kafka timestamps are always given in milliseconds, which is why first_timestamp was already divided by 1000 here. However, the timestamp_delta was then added as is, which is incorrect -- adding milliseconds to a timestamp in seconds will always produce the wrong result (except in the degenerate case where the delta is zero).

We could simply divide the delta by 1000 in the same way that was already being done for the first timestamp, but that would discard the sub-second part of the timestamp. Instead, first calculate the first timestamp using the full millisecond resolution (using integer arithmetic for this avoids floating point rounding errors). Then, add the delta divided by 1000 as a float to come up with each record's correct timestamp (we can't use integer arithmetic for this because a Time can only have seconds added to it).

With this patch, the record timestamps reported by ruby-kafka match those from <https://github.com/fgeller/kt>, so I'm reasonably confident this is correct.

I believe this is a fix for <https://github.com/zendesk/ruby-kafka/issues/721>.

### DIFF
--- a/lib/kafka/protocol/record_batch.rb
+++ b/lib/kafka/protocol/record_batch.rb
@@ -1,3 +1,4 @@
+require 'bigdecimal'
 require 'digest/crc32'
 require 'kafka/protocol/record'
 
@@ -131,7 +132,7 @@ module Kafka
 
         records.each_with_index do |record, index|
           record.offset_delta = index
-          record.timestamp_delta = (record.create_time - first_timestamp).to_i
+          record.timestamp_delta = ((record.create_time - first_timestamp) * 1000).to_i
         end
         @last_offset_delta = records.length - 1
       end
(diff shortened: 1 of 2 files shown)


Elyahou:
This syntax is not supported before ruby 2.5. Maybe you should BigDecimal to avoid floating point rounding errors ?

Using something like 'seconds/BigDecimal(1000)' in place of just `/1000` in all places

lib/kafka/protocol/record_batch.rb
```diff
@@ -167,8 +167,12 @@ def self.decode(decoder)
         log_append_time = (attributes & TIMESTAMP_TYPE_MASK) != 0
 
         last_offset_delta = record_batch_decoder.int32
-        first_timestamp = Time.at(record_batch_decoder.int64 / 1000)
-        max_timestamp = Time.at(record_batch_decoder.int64 / 1000)
+
+        first_timestamp = record_batch_decoder.int64
+        first_timestamp = Time.at(first_timestamp / 1000, first_timestamp % 1000, :millisecond)
```
Elyahou: This syntax is not supported before ruby 2.5. Maybe you should BigDecimal to avoid floating point rounding errors ?

Using something like `seconds/BigDecimal(1000)` in place of just /1000 in all places

stevenjm: Thanks, fixed.

stevenjm:
I just noticed that the reverse bug exists when producing messages. I've force-pushed a revision that fixes that as well.

PR #769, opened 2019-08-20
add kafka 2.3.1

vvuibert:
add kafka 2.3.1
update readme

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,6 +203,38 @@ jobs:
       - run: bundle install --path vendor/bundle
       - run: bundle exec rspec --profile --tag functional spec/functional
 
+  kafka-2.3:
+    docker:
+      - image: circleci/ruby:2.5.1-node
+        environment:
+          LOG_LEVEL: DEBUG
+      - image: wurstmeister/zookeeper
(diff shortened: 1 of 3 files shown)


dasch:
Can you rebase on master? There was a regression in the test suite.

dasch:
So the Kafka 2.3 build is failing the consumer tests, and I _think_ that's actually because of a regression in Kafka 2.3. Can you check out if that's the case, and if so whether a fix has been released?

andrewjamesbrown:
@dasch I'm seeing some weirdness with consumer-group partition assignments - to me it looks like with Kafka 2.3, there is a re-assignment happening after consumers start processing the events.  We're working on some debugging - will let folks know what we come up with 👍

krasnopolsky:
Running into the consumer-group partition issue with our 2.3.0 cluster, just wanted to reference the kafka issue since I didn't see it here https://issues.apache.org/jira/browse/KAFKA-8653

vvuibert:
I'll update the docker tag once `2.3.1` is released with https://github.com/apache/kafka/pull/7235

vvuibert:
The `2.3.1` docker image is still not out yet

PR #774, opened 2019-08-24
Support Rails 6.0, require active_support for specs

tjwp:
This addresses the build issue seen in #770, to restore support for Rails 6.0.

Based on https://guides.rubyonrails.org/active_support_core_extensions.html#stand-alone-active-support, `active_support` should be required. That doesn't actually load much, but then when `active_support/notifications` is required the other relevant dependencies will be found.

Still needs to require  `active_support/core_ext/object/try` to work around a missing require within active_support/notifications. This has been fixed on master here: https://github.com/rails/rails/commit/530f7805ed5790af1d472a041bc74089dc183f47#diff-7d21210eb262dd39635c5b4952d29dee

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
+require "active_support"
 require "active_support/notifications"
 require "kafka"
 require "kafka/tagged_logger"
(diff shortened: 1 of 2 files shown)


tjwp:
Test failure looks unrelated.

spec/spec_helper.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 $LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
+require "active_support"
+require "active_support/core_ext/object/try"
```
dasch: Can you add a comment about this require so it can be removed at some point?

dasch:
Thanks!

PR #779, opened 2019-09-24
fix require on README.md

cesartalves:
update require 'kafka' to 'ruby-kafka', which is the correct gem name.

I was trying the project on my local machine and took about 30 minutes to figure out that I had required the wrong lib :(

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Please see the [documentation site](http://www.rubydoc.info/gems/ruby-kafka) for
 A client must be initialized with at least one Kafka broker, from which the entire Kafka cluster will be discovered. Each client keeps a separate pool of broker connections. Don't use the same client from more than one thread.
 
 ```ruby
-require "kafka"
+require "ruby-kafka"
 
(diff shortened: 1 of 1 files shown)


dasch:
It should be possible to require `kafka`. Did you use Bundler?

cesartalves:
Yes. Apparently there's another gem with the name kafka on rubygems. This is what I get when I try to install it via bundler: 
Screenshot from 2019-11-11 10-46-05

here's a screenshot from rubygems:
Screenshot from 2019-11-11 10-47-24

So whenever I install using 'kafka' on my gemfile or via gem install, I get this other library.

dasch:
`gem install` and `require` are separate things – you need to `gem install ruby-kafka`, but you can `require "kafka"` just fine.

cesartalves:
I see. I doubled checked and in fact I can require 'kafka' if I've got the correct gem installed and no other named kafka.

Maybe there could be a warning in the README indicating that there's another gem with the name kafka, in case somebody makes this same mistake in the future. Closing this request

PR #780, opened 2019-09-30
Prometheus v0.10.0.pre.alpha2 api

dgmcguire:
We needed prometheus metrics and were already using the 0.10.0.pre.alpha2 branch for prometheus-client.  This PR has updates that conform to the breaking API changes in that prometheus-client version.

I realize this is targeting an alpha version of prometheus-client, so I didn't know if you wanted to support that or not, but I figured I would make this PR for anyone else who needed the same support.

### DIFF
--- a/ruby-kafka.gemspec
+++ b/ruby-kafka.gemspec
@@ -44,7 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rspec_junit_formatter", "0.2.2"
   spec.add_development_dependency "dogstatsd-ruby", ">= 3.0.0", "< 5.0.0"
   spec.add_development_dependency "statsd-ruby"
-  spec.add_development_dependency "prometheus-client"
+  spec.add_development_dependency "prometheus-client", "0.10.0.pre.alpha2"
   spec.add_development_dependency "ruby-prof"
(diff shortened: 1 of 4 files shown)


dasch:
It’s probably best to wait for a stable API before merging.

dgmcguire:
It looks like there's a non-alpha 0.10.0 out [here](https://github.com/prometheus/client_ruby/tree/v0.10.0)

the tests pass for me locally (although both this branch and master have thread exceptions on testing for me locally)

ruby-kafka.gemspec
```diff
@@ -44,7 +44,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rspec_junit_formatter", "0.2.2"
   spec.add_development_dependency "dogstatsd-ruby", ">= 3.0.0", "< 5.0.0"
   spec.add_development_dependency "statsd-ruby"
-  spec.add_development_dependency "prometheus-client"
+  spec.add_development_dependency "prometheus-client", "0.10.0.pre.alpha2"
```
dasch:
```suggestion
  spec.add_development_dependency "prometheus-client", "~> 0.10.0"
```

dasch:
Thanks!

PR #786, opened 2019-10-15
Send JoinGroupRequest v1 messages with rebalance timeouts

stevenjm:
If we send a JoinGroupRequest v0 message to join an existing consumer group, the broker does not wait for other members to send their own JoinGroupRequests, resulting in all other members being kicked out of the group. When they rejoin later, the same thing happens, resulting in the group continually being rebalanced with a single member each time.

Sending a JoinGroupRequest v1 message instead lets us specify a timeout to wait for other members. A new member is now added to the consumer group as expected, rather than becoming the only member.

The default value of 1 minute for rebalance_timeout is taken from what Sarama (Go library) does. librdkafka defaults to 5 minutes here, which seems a little excessive.

This will break if talking to a Kafka broker older than 0.10.1. This is unlikely to be a problem because these messages are only used in ruby-kafka to support consumer groups, which were already broken due to the issue described above.

While here, also send the correct group protocol information -- we only support round-robin assignment, and we have an actual list of topics rather than "test-messages". This is needed for interoperability with other clients participating in the same consumer group (found when debugging this issue with Sarama).

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -337,6 +337,7 @@ module Kafka
     def consumer(
         group_id:,
         session_timeout: 30,
+        rebalance_timeout: 60,
         offset_commit_interval: 10,
         offset_commit_threshold: 0,
         heartbeat_interval: 10,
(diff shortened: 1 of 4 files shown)


stevenjm:
The unit test failures appear to be unrelated to my change.

stevenjm:
Rebased on master since the test failures are now fixed there.

dasch:
Thank you! Apologies for the wait time.

PR #791, opened 2019-11-07
AsyncProducer#produce should fail fast if `topic` is a "bad" type

yarmiganosca:
One of my users (I maintain a library that wraps ruby-kafka) was producing asynchronously with the topic argument being a Symbol, but they weren't seeing problems until messages couldn't be added to the async message queue because the thread producer couldn't deliver messages with Symbol topics.

The proposed code ensures that `topic` is either a String or a Symbol upfront (every Ruby object has a String representation, and while a Symbol makes sense, a Hash doesn't, so the code needs to explicitly typecheck `topic`).

I updated all 3 producing methods because it seems nicer to have a uniform, and explanatory error message (the previous one was a `NoMethodError` about a method named `bytesize`).

If y'all don't want to allow Symbols, let me know, and I'll change it to only allow Strings.

If the "preamble" that's now in all 3 producing methods offends your sensibilities (it kind of offends mine), I thought creating an uppercase `Topic` method (like `Kernel#Array`) to encapsulate all the type-checking logic might be a nice way to do it. That way, the single line `topic = Topic(topic)` is all that would have to be added to any producing method. Let me know if you'd prefer that (or something else) to the typecheck preamble I've written, and I'll change it.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -103,6 +103,12 @@ module Kafka
     # @raise [BufferOverflow] if the message queue is full.
     # @return [nil]
     def produce(value, topic:, **options)
+      unless topic.is_a?(String) || topic.is_a?(Symbol)
+        raise ArgumentError, "topic must be either a String or a Symbol, not a #{topic.class.name}"
+      end
+
+      topic = topic.to_s
+
       ensure_threads_running!
 
       if @queue.size >= @max_queue_size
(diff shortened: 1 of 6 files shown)


dasch:
I think the right approach in ruby is to call the `#to_str` method on the input.

yarmiganosca:
Just to make sure I understand, you want to remove the typechecking and just keep `topic = topic.to_str`?

dasch:
Yup. Then any string-like object will be accepted. ‘to_s’ is too broad and wouldn’t catch bugs, but ‘to_str’ is fine.

yarmiganosca:
`#to_str` means Symbols wouldn't be accepted. Is that your intent? (I don't mind if it is one way or another for the reasons I already mentioned, I just want confirmation)

dasch:
Yeah, I see no reason to support passing in a Symbol directly.

dasch:
Thanks!

PR #792, opened 2019-11-21
Cover transactional manager with retries on certain errors

germanosin:
Kafka Java Client handle certain types of errors with retries
For example:
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L1185
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L1189
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L1141

This PR sync behavior for these scenarios.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -270,6 +270,8 @@ module Kafka
         transactional: transactional,
         transactional_id: transactional_id,
         transactional_timeout: transactional_timeout,
+        max_retries: max_retries,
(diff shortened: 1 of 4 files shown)


dasch:
This seems to break the tests.

PR #793, opened 2019-12-06
Expose method for fetching all group offsets

### LINKED ISSUE #785: Expose a consumer group offsets method in the client
For tooling / automation scripts, it would be super helpful to expose a method in the client object for querying consumer group offsets. It looks like this information is available via an `OffsetFetchRequest`.

If this is something that makes sense to add, I can take a stab at writing it.

Another interesting question, should this be a new method or an enhancement to the existing `describe_group` method?

rkrage:
Resolves #785 

@dasch 

Not sure if there's a better approach to the api version logic. Lemme know what you think!

I'll take a stab at writing some tests if this looks good!

### DIFF
--- a/lib/kafka/protocol/offset_fetch_request.rb
+++ b/lib/kafka/protocol/offset_fetch_request.rb
@@ -13,7 +13,7 @@ module Kafka
       end
 
       def api_version
-        1
+        @topics.nil? ? 2 : 1
       end
 
       def response_class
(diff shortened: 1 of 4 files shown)


lib/kafka/protocol/offset_fetch_request.rb
```diff
@@ -13,7 +13,7 @@ def api_key
       end
 
       def api_version
-        1
+        @topics.nil? ? 2 : 1
```
dasch: Can you add a comment here documenting why this is necessary?

rkrage: Done

dasch:
Thanks!

PR #795, opened 2019-12-09
Define instance variables to avoid warnings

andrewjamesbrown:
We've noticed warnings when calling 
- `open?` like on https://github.com/zendesk/ruby-kafka/blob/v0.7.10/lib/kafka/connection.rb#L98
- `idle?` like on https://github.com/zendesk/ruby-kafka/blob/v0.7.10/lib/kafka/connection.rb#L94 

To fix this, we're defining them in the initializer.  
Those interrogator methods already check if those instance variables are non-nil so defining them will be safe.


andrewjamesbrown:
@dasch in case you missed this 😸

dasch:
Thanks!

Issue #796, opened 2019-12-09
Possible flaky create partitions test


andrewjamesbrown:
If this is a bug report, please fill out the following:

* Version of Ruby: `circleci/ruby:2.5.1-node`
* Version of Kafka: `2.1.1`
* Version of ruby-kafka: `master`

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.

###### Steps to reproduce

I've been getting failing tests for the `create partitions` test in CI but only on a single version of Kafka, which _feels_ to me like a flaky test due to async operations or something like that

@dasch how would you like to see this fixed?
Ideas:
- Add a static `sleep(1)` between the `create_partitions_for` and `partitions_for` check
- Add a retry so we check several times (e.g. by adding https://github.com/NoRedInk/rspec-retry or https://github.com/rstacruz/rspec-repeat)
- Other?

andrewjamesbrown:
Failing test: https://circleci.com/gh/Shopify/ruby-kafka/295?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

dasch:
I think the retry approach could work well. I'd love to see a PR.

PR #797, opened 2019-12-10
Add retries to create partitions test

andrewjamesbrown:
Closes https://github.com/zendesk/ruby-kafka/issues/796

I noticed when running CI on https://github.com/zendesk/ruby-kafka/pull/795 that `create partitions` was failing on a single Kafka version, which seemed weird.  That particular test also depends on the topic metadata getting updated, which may be subject to a race condition.

To fix this, I'm adding a retry to that particular case (and including a small `with_retry` function to accomplish that).  I investigated using https://github.com/NoRedInk/rspec-retry or https://github.com/rstacruz/rspec-repeat, however those wrap the entire `example` block when in this particular case we only want the `expect` to be retried.

### DIFF
--- a/spec/functional/topic_management_spec.rb
+++ b/spec/functional/topic_management_spec.rb
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 
+def with_retry(opts = {}, &block)
+  retries = opts[:retry] || 3
+  retry_wait = opts[:retry_wait] || 1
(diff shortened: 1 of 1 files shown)


andrewjamesbrown:
@dasch in case you missed this 😸

dasch:
Thanks!

PR #810, opened 2020-02-20
Fix missing constant errors

jonmagic:
## Why?

After cloning https://github.com/zendesk/ruby-kafka and following the development instructions to run `bin/setup` and then `rake spec` I encountered a number of failed specs with errors that looked like this:

```
      NameError:
        uninitialized constant Datadog::Statsd::Connection::DEFAULT_HOST
        Did you mean?  Datadog::Statsd::DEFAULT_BUFFER_SIZE
```

## How?

Applied suggestion from @tjwp in https://github.com/zendesk/ruby-kafka/pull/810#issuecomment-589415255.

### DIFF
--- a/lib/kafka/datadog.rb
+++ b/lib/kafka/datadog.rb
@@ -40,7 +40,7 @@ module Kafka
       end
 
       def host
-        @host ||= default_host
+        @host
       end
 
       def host=(host)
@@ -49,7 +49,7 @@ module Kafka
       end
 
       def port
-        @port ||= default_port
+        @port
       end
 
       def port=(port)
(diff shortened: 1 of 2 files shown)


tjwp:
I don't think the right fix for this is to lock the version. The dogstatsd-ruby gem will supply the default values for host and port if nil is passed for those arguments. This looks to be true for v4.6.0 & v4.7.0, but not v3.3.0.

So I'd suggest requiring >= 4.0 & < 5.0 and allowing Datadog::Statsd to provide the defaults instead of referencing them in this gem.

jonmagic:
Updated my racecar 2 beta PR as well https://github.com/zendesk/racecar/pull/147, thanks @tjwp!

dasch:
Thanks!

renjiexu:
Thanks for the fix, wondering if you can release a new version including the fix?

dasch:
I've released v1.0.0 containing the fix.

PR #813, opened 2020-02-28
Add Envy functionality

Tim-Willard:
Hello!
​
We (@magmastonealex, @omstrumpf, @gbateman, @Tim-Willard) have been working on a tool called ENVy.
​
ENVy is an environment manager to make working on disparate projects easier. You don't need to worry about what a system package is called on your Linux distribution, or figure out how to build for Mac from scratch.

​Setup with ENVy is as simple as typing `envy up` for any project you want to work on. Running `envy nuke` removes all traces of that environment.
With ENVy, you don't need to hunt down a lint command or how you run the tests. `envy -h` shows you all the commands you can run - or just jump into a shell with `envy shell`.
​
You can read more about ENVy [here](http://envy-project.github.io/), but the short version is that we think it'll make it easier for contributors to help out. 
​
This PR adds an `ENVyfile`, the file that describes a development environment and contains the commands to run. We used ruby-kafka while developing in order to test ENVy, and we hoped you'd consider committing this file to help out casual contributors.

While ruby-kafka doesn't have a particularly complex setup guide, one thing that might be difficult to new users is setting up a kafka cluster to test on. The envyfile provided here also starts up the docker-compose file containing the kafka cluster, and `envy console` will run the ruby console script inside the container with a helpful message on how to connect to the 3 test brokers.
 
In order to get this working, I modified the docker-compose file that was in the repository. I couldn't figure out how to connect to the brokers inside at all with the given configuration. This hopefully shouldn't break anything (I think the configuration change I made should be backwards compatible), but I couldn't find what the docker-compose file is used for so I can't be certain. If there is a use case I should test please let me know.
​

### DIFF
--- a/.envyfile
+++ b/.envyfile
@@ -1,0 +1,27 @@
+environment:
+    base:
+        image: ruby:stretch
+    setup-steps:
+    - name: bundle
+      label: 'Installing Ruby Dependencies'
+      type: script
+      run:
+        - 'bundle install'
+      triggers:
+        files:
+          - Gemfile
+          - Gemfile.lock
+actions:
+    - name: bundle
+      script: 'bundle'
+      help: 'Execute ruby bundler inside environment'
+    - name: console
+      script: 'echo "Connect to kafka brokers at kafka1:9092,kafka2:9092,kafka3:9092" && bin/console'
+      help: 'Start an interactive prompt that will allow you to experiment'
+    - name: test
+      script: 'rake spec'
+      help: 'Run tests'
+services:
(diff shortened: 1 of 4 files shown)


Tim-Willard:
Hmm, it looks like circleci is failing due to [this](https://github.com/DataDog/dogstatsd-ruby/issues/134) issue, which should be unrelated to any of the changes made here.

dasch:
Thanks for the PR, I’ll try to find the time to test this out.

PR #818, opened 2020-03-28
[Feat] Allow consumers to refresh the topic lists

lmduc:
## Description

- Enable consumers to automatically subscribe to the new topics that match the regex
- This feature is not enabled by default to keep the backward compatibility

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -44,7 +44,7 @@ module Kafka
   #
   class Consumer
 
-    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:)
+    def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manager:, session_timeout:, heartbeat:, refresh_topic_interval: 0)
       @cluster = cluster
       @logger = TaggedLogger.new(logger)
       @instrumenter = instrumenter
@@ -53,6 +53,7 @@ module Kafka
       @session_timeout = session_timeout
       @fetcher = fetcher
       @heartbeat = heartbeat
+      @refresh_topic_interval = refresh_topic_interval
 
       @pauses = Hash.new {|h, k|
         h[k] = Hash.new {|h2, k2|
@@ -74,9 +75,11 @@ module Kafka
       #   some already consumed messages
       @current_offsets = Hash.new { |h, k| h[k] = {} }
 
-      # Hash storing topics that are already being subscribed
-      # When subcribing to a new topic, if it's already being subscribed before, skip it
-      @subscribed_topics = Set.new
+      # Map storing subscribed topics with their configuration
+      @subscribed_topics = Concurrent::Map.new
+
+      # Set storing topics that matched topics in @subscribed_topics
+      @matched_topics = Set.new
 
       # Whether join_group must be executed again because new topics are added
       @join_group_for_new_topics = false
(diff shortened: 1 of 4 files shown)


lib/kafka/consumer.rb
```diff
@@ -73,6 +73,10 @@ def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manage
       #   when user commits message other than last in a batch, this would make ruby-kafka refetch
       #   some already consumed messages
       @current_offsets = Hash.new { |h, k| h[k] = {} }
+
+      # Hash storing topics that are already being subscribed
+      # When subcribing to a new topic, if it's already being subscribed before, skip it
+      @subscribed_topics = {}
```
dasch: Can you use `Set` here instead?

lmduc: @dasch I updated my PR as your recommendation.

lib/kafka/consumer.rb
```diff
@@ -93,16 +100,28 @@ def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manage
...
+
+      subscribe_to_regex(topic_or_regex, default_offset, start_from_beginning, max_bytes_per_partition)
+      if refresh_topic_interval > 0
+        @thread ||= Thread.new do
+          while true
+            subscribe_to_regex(topic_or_regex, default_offset, start_from_beginning, max_bytes_per_partition) if running?
+            sleep refresh_topic_interval
+          end
+        end
+        @thread.abort_on_exception = true
```
dasch: This part is problematic. `@thread` is only set on the first invocation of `subscribe`, which can be called multiple times. There should either be a single thread that consults a set of regexes or one thread per regex. Also, the thread or threads need to be killed on consumer shutdown.

lib/kafka/consumer.rb
```diff
@@ -93,16 +100,28 @@ def initialize(cluster:, logger:, instrumenter:, group:, fetcher:, offset_manage
...
-        end
-      else
+      unless topic_or_regex.is_a?(Regexp)
         subscribe_to_topic(topic_or_regex, default_offset, start_from_beginning, max_bytes_per_partition)
+        return
+      end
+
+      subscribe_to_regex(topic_or_regex, default_offset, start_from_beginning, max_bytes_per_partition)
+      if refresh_topic_interval > 0
+        @thread ||= Thread.new do
```
nguyenquangminh0711, line 118: Spawning a thread here is error-prone, and hard to control. And like @dasch said, if the user calls this method multiple times, it may spawns multiple threads doing the same thing. In this context, the consumer should only be aware of the new refereshed set of topics before fetching a new fetch. You can choose another approach:
- At the beginning of method `fetch_batches`, check for topic cache expiration. If expired, fetch new list of topics from the cluster. If the topic stay the same, continue.
- If the list of topic changes:
  - Subscribe the new topic to fetcher
  - Subcribe the new topic to group manager
  - Subscribe the new topic to offset manager
  - Mark cluster information as stale, just to be sure
  - Force join group

lmduc:
Thanks @dasch and @nguyenquangminh0711 for pointing out the problem. I took the approach based on Minh's suggestion and remove the thread.

lib/kafka/consumer.rb
```diff
@@ -572,10 +592,35 @@ def clear_current_offsets(excluding: {})
       end
     end
 
+    def scan_for_subscribing
+      @subscribed_topics.keys.each do |topic_or_regex|
+        default_offset = @subscribed_topics[topic_or_regex][:default_offset]
+        start_from_beginning = @subscribed_topics[topic_or_regex][:start_from_beginning]
+        max_bytes_per_partition = @subscribed_topics[topic_or_regex][:max_bytes_per_partition]
```
dasch, line 599: Would it not make sense to use `each` here instead, e.g.

```ruby
@subscribed_topics.each do |topic_or_regex, topic_config|
  default_offset = topic_config.fetch(:default_offset) # also I prefer `fetch` because it crashes explicitly
  ...
end
```

lmduc: Done

dasch:
Thanks!

PR #822, opened 2020-04-13
Add client function for fetching topic replica count

lihightower:
For tooling / automation scripts, it would be super helpful to expose a method in the client object for querying a topic's replica count, which is possible from the `cluster.partitions_for` function. 

https://github.com/zendesk/ruby-kafka/blob/1b2ed7cd0c79d4ae7e915c4be616df6f59a93eba/lib/kafka/cluster.rb#L180

Another option could be to write a client function that returns the full partition array instead of just the `count` then I can grab the replica count from my code.

https://github.com/zendesk/ruby-kafka/blob/1b2ed7cd0c79d4ae7e915c4be616df6f59a93eba/lib/kafka/client.rb#L694

Let me know what you think or if there's a better place for this.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -694,6 +694,15 @@ module Kafka
       @cluster.partitions_for(topic).count
     end
 
+    # Counts the number of replicas for a topic's partition
+    #
+    # @param topic [String]
+    # @return [Integer] the number of replica nodes for the topic's partition
+
+    def replica_count_for(topic)
+      @cluster.partitions_for(topic).first.replicas.count
+    end
+
     # Retrieve the offset of the last message in a partition. If there are no
     # messages in the partition -1 is returned.
     #
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -698,7 +698,6 @@ module Kafka
     #
     # @param topic [String]
     # @return [Integer] the number of replica nodes for the topic's partition
-
     def replica_count_for(topic)
       @cluster.partitions_for(topic).first.replicas.count
     end
--- a/lib/kafka/protocol/metadata_response.rb
+++ b/lib/kafka/protocol/metadata_response.rb
@@ -34,7 +34,7 @@ module Kafka
     #
     class MetadataResponse
       class PartitionMetadata
-        attr_reader :partition_id, :leader
+        attr_reader :partition_id, :leader, :replicas
 
         attr_reader :partition_error_code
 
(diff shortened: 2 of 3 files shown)


lib/kafka/client.rb
```diff
@@ -694,6 +694,15 @@ def partitions_for(topic)
       @cluster.partitions_for(topic).count
     end
 
+    # Counts the number of replicas for a topic's partition
+    #
+    # @param topic [String]
+    # @return [Integer] the number of replica nodes for the topic's partition
+
```
dasch, line 701: Can you remove the blank line? It'll screw up the docs I think.

lihightower: Done

dasch:
Thanks!

theturtle32:
Nice convenience function. Keep in mind, though, that while I imagine it would be very uncommon, it's technically possible that not all partitions will have the same replica count.

lihightower:
@theturtle32 I've found that to be true during partition reassignment operations specifically, so that is one caveat to this function. Even though that is uncommon in my experience, it might be worth noting that in the documentation. 

It'd be great if there was a more consistent way to get a topic's replication factor from Kafka, but I wasn't able to find one.

lihightower:
In earlier versions of Kafka, the `kafka-topics` command line tool evaluates the topic's replication factor by counting the replicas on the first partition object. However, it looks like in Kafka v2.5, the tool checks to see if a partition reassignment is in progress when evaluating the replication factor (e.g. [`TopicCommand.scala`](https://github.com/apache/kafka/blob/89317c6a9eac24a1c9468260881db06818c1b019/core/src/main/scala/kafka/admin/TopicCommand.scala#L558) ).

I see now that the Kafka AdminClient has a [`ListPartitionReassignments`](https://kafka.apache.org/protocol#The_Messages_ListPartitionReassignments) API  that could be implemented in this library in a similar way to make this function more reliable  🤔 
```
ListPartitionReassignments Response (Version: 0) => throttle_time_ms error_code error_message [topics] TAG_BUFFER 
  throttle_time_ms => INT32
  error_code => INT16
  error_message => COMPACT_NULLABLE_STRING
  topics => name [partitions] TAG_BUFFER 
    name => COMPACT_STRING
    partitions => partition_index [replicas] [adding_replicas] [removing_replicas] TAG_BUFFER 
      partition_index => INT32
      replicas => INT32
      adding_replicas => INT32
      removing_replicas => INT32
```

Issue #826, opened 2020-05-05
ssl_verify_hostname var is ignored unless also specifying a ca_cert var


lairen:
* Version of Ruby: 2.6.0
* Version of Kafka: 1.1.1
* Version of ruby-kafka: 1.0.0

###### Steps to reproduce

```ruby
kafka = Kafka.new(
seed_brokers: seed_brokers,
ssl_client_cert: File.read(cert_path),
ssl_client_cert_key: File.read(cert_key_path),
ssl_verify_hostname: false
)

kafka.topics 
```

###### Expected outcome

I expected the `verify_hostname` attr of the `ssl_context` to be set to false and passed to the OpenSSL gem with that value so that hostname is not verified and I can successfully connect to the broker. 

###### Actual outcome

The `verify_hostname` attribute is ignored unless I explicitly pass in a `ca_cert*` parameter. Since it is defaulted to `true` in the OpenSSL gem, I'm getting these SSL errors:

``` 
ERROR: OpenSSL::SSL::SSLError - SSL_connect returned=1 errno=0 state=error: 
certificate verify failed (unspecified certificate verification error)
```
I'm able to get around this by either downgraded to Ruby v2.3.3 (where that version of OpenSSL does not default `verify_hostname`) or by setting `ssl_ca_certs_from_system: true` in v2.6.0, which I don't need in my case since my `client_cert` is a combined cert. I can open up a PR to move the `verify_hostname` assignment out of this conditional but wanted to check if that was intentional and expected.

https://github.com/zendesk/ruby-kafka/blob/0d64a9b05326c02c5f148c710f5da25990a84d0b/lib/kafka/ssl_context.rb#L45-L60

dasch:
It’s not intentional - please do open a PR!

PR #828, opened 2020-05-11
Make verify_hostname settable for ssl contexts

lairen:
Resolves https://github.com/zendesk/ruby-kafka/issues/826

@dasch


lairen:
@dasch Updated!

PR #832, opened 2020-05-29
Use the standard SPDX ID for license in gemspec

kenhys:
It is better to use SPDX ID for license field.

  https://guides.rubygems.org/specification-reference/#license=

The list of licenses:

  https://spdx.org/licenses/


dasch:
Thanks!

PR #835, opened 2020-06-05
Add concurrent-ruby dependency.

theturtle32:
Somewhere between v1.0 and v1.1.0-beta1, a change was committed that uses `Concurrent::Map` but without adding `concurrent-ruby` to the dependencies in the gemspec.

The tests pass because `activesupport` is a development dependency, and they've had `concurrent-ruby` as one of their dependencies since version 5.0.0.

But without this in the gemspec, any project that doesn't load in activesupport or concurrent-ruby itself won't work with ruby-kafka.

### DIFF
--- a/lib/kafka/consumer.rb
+++ b/lib/kafka/consumer.rb
@@ -4,6 +4,7 @@ require "kafka/consumer_group"
 require "kafka/offset_manager"
 require "kafka/fetcher"
 require "kafka/pause"
+require "concurrent"
 
 module Kafka
 
--- a/lib/kafka/version.rb
+++ b/lib/kafka/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Kafka
-  VERSION = "1.1.0.beta1"
+  VERSION = "1.1.0.beta2"
 end
--- a/ruby-kafka.gemspec
+++ b/ruby-kafka.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'digest-crc'
+  spec.add_dependency 'concurrent-ruby', '~> 1.0', '>= 1.0.2'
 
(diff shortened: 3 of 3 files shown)


theturtle32:
@dasch - It would have been good to get this included in the v1.1.0 release... maybe as a follow-up as v1.1.1?

theturtle32:
I'm not sure what the rationale was for bringing in Concurrent::Map. Maybe it could be replaced with a Hash with mutex-guarded access? Is it necessary at all?

dasch:
I'm not exactly sure why it is needed – I missed it in my review. If you could do some digging it would be great!

vvuibert:
@lmduc is [concurrent](https://github.com/zendesk/ruby-kafka/pull/818/files#diff-e5915ca854c22ecd19d031b280448fdcR79) needed?

TJC:
See related issues #840 and #842

lib/kafka/consumer.rb
```diff
@@ -4,6 +4,7 @@
 require "kafka/offset_manager"
 require "kafka/fetcher"
 require "kafka/pause"
+require "concurrent"
```
fernandes, line 7: maybe it can be `require "concurrent/map"` to avoid loading the whole library, only the Map that is the needed one, wdyt?

dasch: I’d like to get rid of the gem dependency altogether. ruby-kafka should be mostly self-contained.

dasch:
Looking at the code, I’m not sure why it needs to be thread safe – it doesn’t look like e.g. the fetcher thread accesses the data structure directly?

lib/kafka/consumer.rb
fernandes: if it isn't needed, removal is a way better solution 😉

d1egoaz:
This was merged https://github.com/zendesk/ruby-kafka/pull/845
Not sure if this PR is still needed, it seems it can be closed

PR #837, opened 2020-06-11
add producer consumer interceptors

vvuibert:
Add producer and consumer interceptors that can intercept messages at different points on producer and consumer in order to trace the path of individual messages across the cluster.

- inspired in [KIP-42](https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors#KIP42:AddProducerandConsumerInterceptors-Motivation)
- intercepts produced messages before they are sent to the brokers
- intercepts consumed messages before are returned to the client
- does not intercept the broker's acknowledgement
- does not intercept when the consumer commits

### DIFF
--- a/spec/consumer_spec.rb
+++ b/spec/consumer_spec.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require "timecop"
+require "fake_consumer_interceptor"
+require "kafka/interceptors"
 
 describe Kafka::Consumer do
   let(:cluster) { double(:cluster) }
@@ -523,5 +525,51 @@ describe Kafka::Consumer do
     subject(:method_original_name) { consumer.method(:send_heartbeat).original_name }
 
     it { expect(method_original_name).to eq(:trigger_heartbeat!) }
+  end
+
+  describe '#interceptor' do
+    it "creates and stops a consumer with interceptor" do
+      interceptor = FakeConsumerInterceptor.new
+      consumer = Kafka::Consumer.new(
+        cluster: cluster,
+        logger: logger,
+        instrumenter: instrumenter,
+        group: group,
+        offset_manager: offset_manager,
+        fetcher: fetcher,
+        session_timeout: session_timeout,
+        heartbeat: heartbeat,
+        interceptors: [interceptor]
+      )
+
+      consumer.stop
+    end
+
+    it "chains call" do
+      interceptor1 = FakeConsumerInterceptor.new(append_s: 'hello2')
+      interceptor2 = FakeConsumerInterceptor.new(append_s: 'hello3')
+      interceptors = Kafka::Interceptors.new(interceptors: [interceptor1, interceptor2], logger: logger)
+      intercepted_batch = interceptors.call(fetched_batches[0])
+
+      expect(intercepted_batch.messages[0].value).to eq "hellohello2hello3"
+    end
+
+    it "does not break the call chain" do
+      interceptor1 = FakeConsumerInterceptor.new(append_s: 'hello2', on_call_error: true)
+      interceptor2 = FakeConsumerInterceptor.new(append_s: 'hello3')
+      interceptors = Kafka::Interceptors.new(interceptors: [interceptor1, interceptor2], logger: logger)
+      intercepted_batch = interceptors.call(fetched_batches[0])
+
+      expect(intercepted_batch.messages[0].value).to eq "hellohello3"
+    end
+
+    it "returns original batch when all interceptors fail" do
+      interceptor1 = FakeConsumerInterceptor.new(append_s: 'hello2', on_call_error: true)
+      interceptor2 = FakeConsumerInterceptor.new(append_s: 'hello3', on_call_error: true)
+      interceptors = Kafka::Interceptors.new(interceptors: [interceptor1, interceptor2], logger: logger)
+      intercepted_batch = interceptors.call(fetched_batches[0])
+
+      expect(intercepted_batch.messages[0].value).to eq "hello"
+    end
   end
 end
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -248,6 +248,9 @@ module Kafka
     #   be in a message set before it should be compressed. Note that message sets
     #   are per-partition rather than per-topic or per-producer.
     #
+    # @param interceptors [Array<Object>] a list of producer interceptors the implement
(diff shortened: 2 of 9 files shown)


d1egoaz:
ping @dasch to see if he can schedule some time to review this, thanks!

dasch:
We already use `ActiveSupport::Notifications` to provide granular pub/sub hooks – what is the use case for introducing this new mechanism?

vvuibert:
@dasch with interceptors we can add the [span and trace ID](https://github.com/dynatrace-oss-contrib/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md) to the message headers so that traces emitted by producers can be linked by the consumer.
[Interceptor API will allow mutate the records to support the ability to add metadata to a message for auditing/end-to-end monitoring.](https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors#KIP42:AddProducerandConsumerInterceptors-Motivation)

dasch:
@vvuibert ah OK, that’s a good use case, and something we need at Zendesk as well. I’ll review this then.

lib/kafka/producer_interceptor.rb
```diff
@@ -0,0 +1,22 @@
...
+    # @param value [Kafka::PendingMessage] the message.
+
+    # @return value [Kafka::PendingMessage] the intercepted message.
+    def on_send(pending_message)
+      raise NotImplementedError, "Implement this method in a child class"
+    end
+
+    # This is called when the interceptor is closed.
+    # @return [nil]
+    def close
```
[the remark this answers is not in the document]
vvuibert: removed

lib/kafka/producer_interceptor.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Kafka
+  # Allows the client to intercept the messages before they get sent to the Kafka brokers.
+  class ProducerInterceptor
```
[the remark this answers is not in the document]
vvuibert: removed

lib/kafka/consumer_interceptors.rb
```diff
@@ -0,0 +1,31 @@
...
+    # The batch returned from the first ConsumerInterceptor.on_consume is passed to the second interceptor on_consume,
+    # and so on in an interceptor chain. This method does not throw exceptions.
+    #
+    # @param batch [Kafka::FetchedBatch] fetched batch.
+    #
+    # @return [Kafka::FetchedBatch] intercepted batch.
+    def on_consume(batch)
+      @interceptors.each do |interceptor|
+        begin
+          batch = interceptor.on_consume(batch)
```
dasch, line 22:
```suggestion
          batch = interceptor.call(batch)
```

lib/kafka/consumer_interceptors.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    # This is called from Kafka::Consumer once the batches are fetched.
+    # The batch returned from the first ConsumerInterceptor.on_consume is passed to the second interceptor on_consume,
+    # and so on in an interceptor chain. This method does not throw exceptions.
+    #
+    # @param batch [Kafka::FetchedBatch] fetched batch.
+    #
+    # @return [Kafka::FetchedBatch] intercepted batch.
+    def on_consume(batch)
```
dasch, line 19:
```suggestion
    def call(batch)
```

lib/kafka/consumer_interceptors.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Kafka
+  # Holds a list of consumer interceptors that implement on_consume
+  # and wraps calls to a chain of custom consumer interceptors.
+  class ConsumerInterceptors
```
dasch, line 6:
```suggestion
  class Interceptors
```

lib/kafka/producer_interceptors.rb
```diff
@@ -0,0 +1,31 @@
...
+    #
+    # @param pending_message [Kafka::PendingMessage] the produced message.
+    #
+    # @return [Kafka::PendingMessage] the intercepted message returned by the last interceptor.
+    def on_send(pending_message)
+      @interceptors.each do |interceptor|
+        begin
+          pending_message = interceptor.on_send(pending_message)
+        rescue Exception => e
+          @logger.warn "Error executing interceptor on_send for topic: #{pending_message.topic} partition: #{pending_message.partition} #{e.class}: #{e.message}\n#{e.backtrace.join("\n")}"
```
dasch, line 24: The only difference between `ConsumerInterceptors` and this class is the error message here, right? I think we could use a single `Interceptors` class and either use a generic error message _or_ use some kind of parameter to `initialize` that we use to construct the error message.

vvuibert: made the error message generic

spec/consumer_spec.rb
```diff
@@ -524,4 +526,50 @@
...
+    it "chains on_consume" do
+      interceptor1 = FakeConsumerInterceptor.new(append_s: 'hello2')
+      interceptor2 = FakeConsumerInterceptor.new(append_s: 'hello3')
+      interceptors = Kafka::Interceptors.new(interceptors: [interceptor1, interceptor2], logger: logger)
+      intercepted_batch = interceptors.call(fetched_batches[0])
+
+      expect(intercepted_batch.messages[0].value).to eq "hellohello2hello3"
+    end
+
+    it "does not break the on_consume chain" do
```
dasch, line 557: Small change: replace `on_consume` with `call` here and see if it’s used elsewhere – same with `on_send`.

vvuibert: oops. fixed :)

dasch:
Thanks!

d1egoaz:
Thanks @dasch for the detailed review!

I wonder what are the plans for an upcoming release that includes this change.
Is ruby-kafka using some kind of release cycle?

I see there are some issue with the `concurrent` dependency https://github.com/zendesk/ruby-kafka/pull/835 that might need to be addressed first before a new release.

We'll keep and 👀 on new releases!

Thanks

dasch:
Yeah, hoping to get that resolved before issuing a new release. There’s no firm cycle.

dasch:
Addressing that issue would certainly expedite a new release :D

PR #838, opened 2020-06-12
Cleanup setting of consecutive return values in stubs

Darhazer:
Minor cleanup of a spec. You don't need a counter variable to setup multiple return values.


dasch:
Thanks!

PR #839, opened 2020-06-12
README calls out problems starting from Kafka 2.2.

robinp-tw:
To prevent giving the false impression that it is safe to use Ruby client with those clusters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,25 +110,27 @@ Or install it yourself as:
   </tr>
   <tr>
     <th>Kafka 2.2</th>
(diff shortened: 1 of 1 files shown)


dasch:
I didn't know it was unsafe to use ruby-kafka with those versions? Is there a reproducible example of that? The linked fluentd issues seems very superficial.

robinp-tw:
Sorry, just anecdotes. Someone might want to run a stress-test, but I don't have the bandwidth for it now.

Issue #842, opened 2020-06-18
undefined method `monotonic_time' for Concurrent:Module (NoMethodError)


TJC:
* Version of Ruby: 2.7.1
* Version of Kafka: (not applicable)
* Version of ruby-kafka: 1.1.0

###### Steps to reproduce

```ruby
require "kafka"
require "kafka/datadog"

kafka = Kafka.new(brokers, client_id: "example")
consumer = kafka.consumer(group_id: "example_group")
consumer.subscribe("my_topic")
consumer.each_message do |msg|
  puts "hi!"
end
```

Attempt to run above code, and receive error:
```
undefined method `monotonic_time' for Concurrent:Module
```

##  Fix

Include this extra require prior to the kafka/datadog import:

```
require "concurrent/utility/monotonic_time"
require "kafka/datadog"
```

TJC:
Hi,
I know this issue was closed 20 days ago, but the issue is still present in the current stable release  - 1.1.0.

Are there any plans to release a new version that contains the fix?

dasch:
v1.2.0 has just been released.

Issue #840, opened 2020-06-16
NameError: uninitialized constant Kafka::Consumer::Concurrent on v1.1.0


fernandes:
If this is a bug report, please fill out the following:

* Version of Ruby: 2.6.6p146
* Version of Kafka: 2.5.0
* Version of ruby-kafka: 1.1.0

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.

It it's not, still showing using [Concurrent::Map](https://github.com/zendesk/ruby-kafka/blob/703bf8065390a6c2c01268d08a245b53b045f738/lib/kafka/consumer.rb#L79)

###### Steps to reproduce

1. Use ruby kafka outside rails (In my case it's a racecar consumer)

###### Expected outcome

Was expected to run normally

###### Actual outcome

=> Crashed: NameError: uninitialized constant Kafka::Consumer::Concurrent
vendor/bundle/ruby/2.6.0/gems/ruby-kafka-1.1.0/lib/kafka/consumer.rb:79:in `initialize'

###### Possible solution

In my case I solved adding `concurrent-ruby` as dependency and then `require 'concurrent/map'` on my lib file.

I suspect this is caused because ActiveSupport install `ruby-concurrency` as its dependency and it's being installed as dev dependency on the [gemspec](https://github.com/zendesk/ruby-kafka/blob/master/ruby-kafka.gemspec#L39) so it wouldn't break in development/tests.

My suggestion is to add `ruby-concurrent` as a runtime dependency

TJC:
I just hit this bug independently!

Likewise, the fix was to add a `require "concurrent/map"

PR #845, opened 2020-06-24
remove concurrent map and use a Hash

d1egoaz:
From the comment: `It seems it doesn't need to be threadsafe.`
Via https://github.com/zendesk/ruby-kafka/pull/835#issuecomment-646030465

I took a quick look and it seems it's safe to use a simple Hash

Fixes: https://github.com/zendesk/ruby-kafka/issues/842
Fixes: https://github.com/zendesk/ruby-kafka/issues/840


dasch:
Thanks!

PR #848, opened 2020-07-01
Add support for configuring the client partitioner

TomerPeleg2:


.gitignore
```diff
@@ -11,3 +11,7 @@
 *.log
 *.swp
 .byebug_history
+
+# IDEA
+*.idea
+*.iml
```
dasch, line 17: We don't add editor specific rules to the project gitignore. You can add global rules to ignore such files in _all_ repos.

README.md
```diff
@@ -349,6 +349,19 @@ partition = PartitioningScheme.assign(partitions, event)
...
+
+```ruby
+class CustomPartitioner
+  def partition_for_key(partition_count, message)
+    ...
+  end
+end
+  
+partitioner = CustomPartitioner.new
+Kafka.new(partitioner: partitioner, ...)
```
dasch, line 362: Looking at this, I think we should use `call` instead of `partition_for_key`. That would allow passing Procs instead of having to add a new class – a more Rubyesque approach in my view.

```ruby
partitioner = -> (partition_count, message) -> { ... }
Kafka.new(partitioner: partitioner, ...)
```

dasch:
Can you add a note to the changelog as well?

tomerpe:
Done

dasch:
Thanks! Sorry for the late merge, just returned from vacation.

PR #849, opened 2020-07-09
Recover from Kafka::MessageSizeTooLarge exception in async producer.

RyanAmos:
You can get into a bad unrecoverable state when a Kafka::MessageSizeTooLarge exception is raised.

When we receive this exception, if we're using the sync producer, then we simply re-raise this exception. If we're using the async producer, then we attempt to remove the largest message from the buffer and simply retry. Once retried, if the buffer is still too large, we'll continue removing the largest message until we can successfully flush the buffer.

This will re-raise the exception since it is done synchronously:
```ruby
DeliveryBoy.deliver('a' * 1_500_000, topic: 'test')
```

This will remove the largest message and continue to retry:
```ruby
DeliveryBoy.deliver_async('a' * 1_500_000, topic: 'test')
```

Would love to get some feedback on this approach.  Specifically @dasch since you had some opinions on the approach here:
https://github.com/zendesk/ruby-kafka/pull/358

### DIFF
--- a/lib/kafka/message_buffer.rb
+++ b/lib/kafka/message_buffer.rb
@@ -48,6 +48,40 @@ module Kafka
       end
     end
 
+    def clear_largest_message
+      target = nil
+      @buffer.each do |topic, messages_for_topic|
+        messages_for_topic.each do |partition, messages_for_partition|
+          messages_for_partition.each do |message|
+            if target.nil? || target[:message].bytesize.to_i < message.bytesize.to_i
+              target = {
+                topic: topic,
+                partition: partition,
+                message: message,
+              }
+            end
+          end
+        end
+      end
+
+      clear_message(target) unless target.nil?
+    end
+
+    def clear_message(topic:, partition:, message:)
+      return if topic.nil? || partition.nil? || message.nil?
+      return unless @buffer.key?(topic) && @buffer[topic].key?(partition)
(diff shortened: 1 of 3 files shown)


lib/kafka/message_buffer.rb
```diff
@@ -48,6 +48,40 @@ def each
...
+                topic: topic,
+                partition: partition,
+                message: message,
+              }
+            end
+          end
+        end
+      end
+
+      clear_message(target) unless target.nil?
```
dasch, line 67:
```suggestion
      clear_message(**target) unless target.nil?
```

This will cause a warning in Ruby 2.7 otherwise, I believe.

lib/kafka/producer.rb
```diff
@@ -245,7 +245,7 @@ def produce(value, key: nil, headers: {}, topic:, partition: nil, partition_key:
     #
     # @raise [DeliveryFailed] if not all messages could be successfully sent.
     # @return [nil]
-    def deliver_messages
+    def deliver_messages(async: false)
```
dasch, line 248: We need to document `async` here.

dasch:
Also, ❤️

RyanAmos:
@dasch I dropped the ball here.  Are you still interested in getting this in?  If so, I'll rebase and add the functional test as asked in the next couple weeks.

PR #852, opened 2020-07-20
Resolve ruby 2.7 warnings

jurre:
In ruby 3.0 positional arguments and keyword arguments will be
separated, and to prepare us for this, ruby 2.7 will start raising
warnings about potential issues.

This change resolves most of the warnings in the test-suite and some
that we have noticed in production.

I bumped .ruby-version for testing and kept it in this PR to make verifying
the changes locally a bit easier, but CI runs everything on 2.5.1 so might
be better to revert that, _or_ update CI to run on 2.7. Happy to update the
PR to accomodate either.


jurre:
It seems like that kafka-2.5 failure is a known flake? Lmk if I need to look into it instead!

jurre:
Hi @dasch, anything I can do to help get these changes in? 💛

dasch:
Thanks!

PR #855, opened 2020-07-28
Various AsyncProducer improvements

cstyles:
See the commit messages for details.

### DIFF
--- a/lib/kafka/async_producer.rb
+++ b/lib/kafka/async_producer.rb
@@ -131,6 +131,8 @@ module Kafka
     # @see Kafka::Producer#deliver_messages
     # @return [nil]
     def deliver_messages
+      ensure_threads_running!
+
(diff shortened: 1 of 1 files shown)


spec/async_producer_spec.rb
```diff
@@ -33,6 +33,24 @@ def instrument(name, payload = {})
...
+
+      [async_producer, async_producer2].each do |p|
+        class << p
+          attr_reader :thread_mutex
+        end
+      end
+
+      expect(async_producer.thread_mutex).to_not eq(async_producer2.thread_mutex)
+    end
+  end
```
dasch, line 52: I don't think we need a test for this – it's not testing functionality, but rather implementation details.

cstyles: Fixed up in 5f22fbd2bc46af22b10e15ead966e5651cc47d2d

PR #856, opened 2020-08-06
change parameter to ssl_verify_hostname in readme

contrun:
Instead of `verify_hostname`, the acceptable parameter for `Kafka` class constructor is `ssl_verify_hostname`. Current readme is kind of confusing, I personally tried the parameter `verify_hostname`, but this failed.


dasch:
Thanks!

Issue #857, opened 2020-08-06
Deprecation warnings in Ruby 2.7 for keyword parameters usage 


tsrivishnu:
Hi,

I have upgraded to ruby-kaka v1.2.0 to use with Ruby 2.7.1 but I see a deprecation warning. Below are the details.

* Version of Ruby: 2.7.1
* Version of Kafka: 2.4.0
* Version of ruby-kafka: 1.2.0

Please verify that the problem you're seeing hasn't been fixed by the current `master` of ruby-kafka.

###### Steps to reproduce

```ruby
kafka = Kafka.new(...)
producer = kafka.async_producer
producer.produce("test message", key: "key1", topic: "dummy")

=> /app/.bundle/ruby/2.7.0/gems/ruby-kafka-1.2.0/lib/kafka/async_producer.rb:211: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/app/.bundle/ruby/2.7.0/gems/ruby-kafka-1.2.0/lib/kafka/async_producer.rb:249: warning: The called method `produce' is defined here
```

###### Expected outcome
No warning about keyword parameters

###### Actual outcome
Warnings about keyword parameters usage that are deprecated in Ruby version 2.7 are shown

dasch:
Can you try on latest master and if it's still a problem, open a PR?

PR #859, opened 2020-08-23
Positional and keyword argument handling in Ruby 2.7

tsrivishnu:
Ruby 2.7 deprecates automatic conversiom between positional and keyword
arguments.

See: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

Resolves: #857


dasch:
Thanks!

PR #862, opened 2020-09-02
Improved Exceptions in TransactionManager

jwilger:
The use of generic RuntimeError exceptions coming from the `TransactionManager` requires brittle error-handling in dependant code, so I've changed most of the `raise` statements in the `TransactionManager` to use the `Kafka::InvalidTxnStateError` exception.

Additionally, there may be cases where dependant code starts a transaction using the producer's `#transaction` block syntax and then an exception is raised within that block *before* the transaction manager actually opens up a transaction. In such cases, prior to this change, the real exception is obfuscated, because the transaction manager would raise an exception
about not being in a valid transactional state.

With this change, attempting to abort a transaction that has not actually been opened will no longer result in an exception. Instead it will log a warning message and leave the transaction manager in its ready state.

### DIFF
--- a/lib/kafka/transaction_manager.rb
+++ b/lib/kafka/transaction_manager.rb
@@ -95,7 +95,7 @@ module Kafka
       force_transactional!
 
       if @transaction_state.uninitialized?
-        raise 'Transaction is uninitialized'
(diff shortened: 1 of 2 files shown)


dasch:
Thanks!

PR #868, opened 2020-10-28
Push compression metrics to StatsD/Datadog/Prometheus

ellimist:
The [Compressor](https://github.com/zendesk/ruby-kafka/blob/d3796daf8433611043f6680412b27355317896e1/lib/kafka/compressor.rb#L78-L81) class is pushing ActiveSupport notifications with metrics regarding the messages getting compressed.

However these don't get pushed to StatsD/Datadog, since there's no subscriber for them.

This PR adds the missing subscribers.


livathinos (approved):
cc @dasch

lib/kafka/datadog.rb
```diff
@@ -108,6 +108,18 @@ def emit(type, *args, tags: {})
...
     end
 
+    class CompressionSubscriber < StatsdSubscriber
+      def compress(event)
+        message_count = event.payload.fetch(:message_count)
+        compressed_bytesize = event.payload.fetch(:compressed_bytesize)
+
+        gauge("compressor.compressed_bytesize", compressed_bytesize)
+        count("compressor.message_count", message_count)
+      end
```
dasch, line 118: I think we should add another histogram here – the compression _factor_. That's much easier to use in comparisons and evaluations of efficacy.

```ruby
histogram("compressor.compression_factor", uncompressed_bytesize / compressed_bytesize)
```

dasch (changes requested):

lib/kafka/prometheus.rb
```diff
@@ -66,6 +67,22 @@ def request(event)
...
+        super
+        @compressed_message_count = Prometheus.registry.counter(:compressed_message_count, docstring: 'Compressed Messages count')
+        @compressed_bytesize = Prometheus.registry.gauge(:compressed_bytesize, docstring: 'Compressed Messages bytesize')
+      end
+
+      def compress(event)
+        message_count = event.payload.fetch(:message_count)
+        compressed_bytesize = event.payload.fetch(:compressed_bytesize)
+
+        @compressed_bytesize.set(compressed_bytesize)
```
dasch, line 81: Does Prometheus support histograms? I'd assume so.

PR #869, opened 2020-11-11
Add kafka 2.6.0 to Circle CI

methodmissing:
Compatibility check for Kafka 2.6.0


vvuibert:
we should also update the readme

dasch:
Thanks!

PR #877, opened 2020-12-13
Support seed brokers' hostname with multiple addresses

abicky:
By default, the Java implementation of a Kafka client can use each returned IP address from a hostname.
cf. https://kafka.apache.org/documentation/#client.dns.lookup

librdkafka also does:
> If host resolves to multiple addresses librdkafka will
> round-robin the addresses for each connection attempt.

cf. https://github.com/edenhill/librdkafka/blob/v1.5.3/INTRODUCTION.md#brokers

Such a feature helps us to manage seed brokers' IP addresses easily, so this PR implements it.

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "kafka/broker_pool"
+require "resolv"
 require "set"
 
 module Kafka
@@ -18,7 +19,8 @@ module Kafka
     # @param seed_brokers [Array<URI>]
     # @param broker_pool [Kafka::BrokerPool]
     # @param logger [Logger]
-    def initialize(seed_brokers:, broker_pool:, logger:)
+    # @param resolve_seed_brokers [Boolean] See {Kafka::Client#initialize}
+    def initialize(seed_brokers:, broker_pool:, logger:, resolve_seed_brokers: false)
       if seed_brokers.empty?
         raise ArgumentError, "At least one seed broker must be configured"
       end
@@ -26,6 +28,7 @@ module Kafka
       @logger = TaggedLogger.new(logger)
       @seed_brokers = seed_brokers
       @broker_pool = broker_pool
+      @resolve_seed_brokers = resolve_seed_brokers
       @cluster_info = nil
       @stale = true
 
@@ -418,32 +421,35 @@ module Kafka
     # @return [Protocol::MetadataResponse] the cluster metadata.
     def fetch_cluster_info
       errors = []
-
       @seed_brokers.shuffle.each do |node|
-        @logger.info "Fetching cluster metadata from #{node}"
+        (@resolve_seed_brokers ? Resolv.getaddresses(node.hostname).shuffle : [node.hostname]).each do |hostname_or_ip|
+          node_info = node.to_s
+          node_info << " (#{hostname_or_ip})" if node.hostname != hostname_or_ip
+          @logger.info "Fetching cluster metadata from #{node_info}"
 
-        begin
-          broker = @broker_pool.connect(node.hostname, node.port)
-          cluster_info = broker.fetch_metadata(topics: @target_topics)
+          begin
+            broker = @broker_pool.connect(hostname_or_ip, node.port)
+            cluster_info = broker.fetch_metadata(topics: @target_topics)
 
-          if cluster_info.brokers.empty?
-            @logger.error "No brokers in cluster"
-          else
-            @logger.info "Discovered cluster metadata; nodes: #{cluster_info.brokers.join(', ')}"
+            if cluster_info.brokers.empty?
+              @logger.error "No brokers in cluster"
+            else
+              @logger.info "Discovered cluster metadata; nodes: #{cluster_info.brokers.join(', ')}"
 
-            @stale = false
+              @stale = false
 
-            return cluster_info
+              return cluster_info
+            end
+          rescue Error => e
+            @logger.error "Failed to fetch metadata from #{node_info}: #{e}"
+            errors << [node_info, e]
+          ensure
+            broker.disconnect unless broker.nil?
           end
-        rescue Error => e
-          @logger.error "Failed to fetch metadata from #{node}: #{e}"
-          errors << [node, e]
-        ensure
-          broker.disconnect unless broker.nil?
         end
       end
 
-      error_description = errors.map {|node, exception| "- #{node}: #{exception}" }.join("\n")
+      error_description = errors.map {|node_info, exception| "- #{node_info}: #{exception}" }.join("\n")
 
       raise ConnectionError, "Could not connect to any of the seed brokers:\n#{error_description}"
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changes and additions to the library will be listed here.
 
 - Fix `Kafka::TransactionManager#send_offsets_to_txn` (#866).
 - Add support for `murmur2` based partitioning.
+- Add `resolve_seed_brokers` option to support seed brokers' hostname with multiple addresses (#877).
 
 ## 1.3.0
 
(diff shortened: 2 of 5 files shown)


dasch:
This seems like a risky change, and I haven't seen other requests for this functionality. Would you consider using rdkafka-ruby instead? I'm wary of introducing risky changes to ruby-kafka unless there's broad demand.

abicky:
What kind of risk do you think there is? I think this change makes ruby-kafka easier to use without any risk.
Not only other Kafka client implementations but the ruby driver of Cassandra also has a similar feature to specify the seed nodes.
cf. https://github.com/datastax/ruby-driver/blob/v3.2.5/lib/cassandra.rb#L727-L741

lib/kafka/cluster.rb
```diff
@@ -420,30 +421,34 @@ def fetch_cluster_info
...
-
-          if cluster_info.brokers.empty?
-            @logger.error "No brokers in cluster"
-          else
-            @logger.info "Discovered cluster metadata; nodes: #{cluster_info.brokers.join(', ')}"
-
-            @stale = false
-
-            return cluster_info
+        Resolv.getaddresses(node.hostname).select { |ip| ip =~ Resolv::IPv4::Regex }.shuffle.each do |ip|
```
dasch, line 424: What about IPv6 addresses?

dasch:
If this new behavior can be made _opt-in_, i.e. disabled by default, then I'd be open to it.

lib/kafka/cluster.rb
abicky: I think currently ruby-kafka doesn't support IPv6 addresses because `Kafka::BrokerUri.parse` can't parse IPv6 addresses.

abicky:
OK, I'll think about another option or something like that.

lib/kafka/cluster.rb
abicky: I noticed that a hostname, a DNS record, could have a IPv6 address, and I didn’t think of such a case...

.circleci/config.yml
```diff
@@ -7,6 +7,7 @@ jobs:
           LOG_LEVEL: DEBUG
     steps:
       - checkout
+      - run: sudo apt-get update && sudo apt-get install -y cmake # For installing snappy
```
sodabrew, line 10: Is this line relevant to the multiple address support?

lib/kafka/cluster.rb
sodabrew: Is there any reason to have this IPv4 regex at all? Why not simply:
```
Resolv.getaddresses(node.hostname).shuffle.each do |ip|
```

If an IPv6 address causes an exception further down the stack, better to solve that further down the stack, right? I don't actually follow what the issue would be with `Kafka::BrokerUri.parse` -- that's further up the stack and would be handling the text hostname before passing it down to this method.

If someone wanted to pass an IPv6 address in natively, the format should be like `kafka://[::1]:9092`

sodabrew:
I strongly agree with this PR! In my experience it is a common practice to use a single hostname that expands to multiple IP address records, and in this case (i.e. using Kafka in the first place because HA is important) it would be helpful to explicitly iterate over the list of records.

I think the PR itself has two issues: 1) unrelated change in circleci.yml and 2) IPv4 limitation for some reason, and would ask if @abicky would remove those lines if that would make this eligible to land the PR?

`<soapbox>`
The current behavior, if you pass in a hostname that resolves to multiple addresses, falls back on a lot of system-implicit-dependencies. Somewhere down the call chain, the hostname **must** be resolved to an IP address in order to make a TCP connection. The DNS resolver _may_ randomize the return order of IPs such that different servers in the same network talk to different seed brokers, but what's twiddly is if one instance does the DNS lookup, uses the first address, and that address is dead... now what? If the client code (i.e. my code that uses Kafka) loops on, say, first N failures, and tries `Kafka.new("server:9192")` again, there's no way to know if you'll get the same dead address from cache, or get a fresh lookup with the same order of records, or get a fresh lookup with a randomized order of records and a chance at connecting to a live server. And if you did know that, you'd either have carefully configured your DNS server to handle this problem, or you'd have a lucky default setting and set a timebomb for your next OS upgrade if the defaults change.
`</soapbox>`

Thank you!

wamaral:
Looking forward to this, it matches our use case: we run a kubernetes cluster with the Kafka brokers exposed through a single DNS entry. Whenever one of the brokers go down (rare but can happen due to our k8s setup), there's a chance our services will try to reach the IP of the unavailable broker due to local DNS caching, and when it can't connect to that particular IP the service will just crash, thinking there are no brokers available

@abicky , are you still working on this? Let me know if I can be of help

lib/kafka/cluster.rb
abicky: I've changed the code as follows:

```
(@resolve_seed_brokers ? Resolv.getaddresses(node.hostname).shuffle : [node.hostname]).each
```

> I think currently ruby-kafka doesn't support IPv6 addresses because Kafka::BrokerUri.parse can't parse IPv6 addresses.

This is my misunderstanding. Although I don't remember well, I might have tried `Kafka.new("kafka://::1:9092")`...

> Is there any reason to have this IPv4 regex at all?

At first, I thought it was meaningless to try IPv6 addresses because both IPv4 addresses and IPv6 addresses might belong to the same broker. However we cannot know whether they belong to the same broker or not, and the DNS record might have only IPv6 addresses.
I also checked the behavior of librdkafka and found that it used both IPv4 addresses and IPv6 addresses even if they belonged to the same broker.
cf. https://github.com/edenhill/librdkafka/blob/v1.6.1/src/rdaddr.c#L162-L232

For the above reasons, I decided not to exclude IPv6 addresses.

abicky:
@dasch I've added the option `resolve_seed_brokers`. Could you review the changes? Although some tests failed, I don't think they are not related to these changes.

@sodabrew @wamaral Thank you for your comments and I'm sorry for my late response. I've pushed the latest code.

.circleci/config.yml
abicky: This line was needed to run the tests. I've removed it because the current master branch includes the change.

lib/kafka/cluster.rb
```diff
@@ -418,32 +421,35 @@ def get_leader_id(topic, partition)
...
-          else
-            @logger.info "Discovered cluster metadata; nodes: #{cluster_info.brokers.join(', ')}"
-
-            @stale = false
-
-            return cluster_info
+        (@resolve_seed_brokers ? Resolv.getaddresses(node.hostname).shuffle : [node.hostname]).each do |hostname_or_ip|
+          node_info = node.to_s
+          node_info << " (#{hostname_or_ip})" if node.hostname != hostname_or_ip
+          @logger.info "Fetching cluster metadata from #{node_info}"
```
sodabrew, line 428:
```suggestion
          @logger.info "Fetching cluster metadata from #{node.hostname}#{node_info}"
```

abicky: `node.hostname` is not necessary because `node.to_s` has the information:

```ruby
node_info = node.to_s # e.g. "kafka://localhost:9092"
# e.g. "kafka://localhost:9092" => ""kafka://localhost:9092 (127.0.0.1)"
node_info << " (#{hostname_or_ip})" if node.hostname != hostname_or_ip
```

Here is an example output:

```
/Users/arabiki/ghq/src/github.com/zendesk/ruby-kafka/lib/kafka/cluster.rb:454:in `fetch_cluster_info': Could not connect to any of the seed brokers: (Kafka::ConnectionError)
- kafka://localhost0:9092 (10.0.0.1): Operation timed out
```

sodabrew: Oh, right on, that looks good already!

dasch (approved):
:+1: looks good!

Can you also add an entry to the changelog?

abicky:
@dasch Thank you for reviewing the PR! I've added the entry to the changelog in the commit https://github.com/zendesk/ruby-kafka/pull/877/commits/ef7c21c8828f51368f914ebac488cda87918cd0f.

dasch (approved):
Thanks!

PR #881, opened 2021-01-10
Fix RubyDoc.info broken links

rafaelsales:
http://www.rubydoc.info/gems/ruby-kafka => https://www.rubydoc.info/github/zendesk/ruby-kafka

The link on GitHub has to be updated as well but only an insider can do it:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ This library requires Ruby 2.1 or higher.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


dasch:
The links don't seem to be broken anymore?

rafaelsales:
Yes! They are working again!

PR #884, opened 2021-01-29
Add `murmur2_random` support for message partitioning.

divo:
Add `murmur2_random` support for message partitioning compatibility with Java producers.

I'm not 100% certain on the new configuration parameter. The caller shouldn't need to know how to instantiate the partitioner, so this is what I came up with. Open to suggestions.

The test failures seem to be unrelated to the changes. Not sure what to do there.

@dasch


lib/kafka/client.rb
```diff
@@ -124,7 +126,14 @@ def initialize(seed_brokers:, client_id: "ruby-kafka", logger: nil, connect_time
...
       @cluster = initialize_cluster
-      @partitioner = partitioner || Partitioner.new
+      @partitioner =
+        if partitioner
+          partitioner
+        elsif partitioner_klass
+          Object.const_get(partitioner_klass).new
+        else
+          Partitioner.new
+        end
```
dasch, line 136: I don't think we should introduce a new concept _next to_ `Partitioner`. Instead, `Partitioner` should be configured with the hash/digest algorithm, with the default being `crc32` but `murmur2` being an option. The "random" part would just always be the behavior when no key is present.

So: can we introduce a new parameter to `Partitioner#initialize` called e.g. `hash_function`, which allows either `crc32` (default) or `murmur2`. Clients that wish to use `murmur2` would need to instantiate `Partitioner` with that value and pass the instance to `producer`.

In Racecar and other high-level frameworks, we can add configuration keys for controlling this declaratively.

ruby-kafka.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'digest-crc'
+  spec.add_dependency "digest-murmurhash"
```
dasch, line 31: What are the dependencies of this gem? The dependency policy for ruby-kafka is pretty strict, so for opt-in behavior the default would be to give an error message when the library cannot be detected and ask the user to install it themselves (would love optional dependencies in gems, but 🤷)

If the gem is really trivial and doesn't include C code then we can add it as a hard dep, but only then.

dasch (changes requested):

ruby-kafka.gemspec
divo: Ah I see how the various compression codecs are used. Would a similar setup to that be ok here?

The gem is fairly trivial but it's basically just C code.

lib/kafka/client.rb
divo: That sounds good. I was taking inspiration from `librdkafka` but your idea will also work well with the opt-in behaviour below.

Just to be clear, you would like the `Partitioner` instance to be passed in, ala [`sasl_oauth_token_provider`](https://github.com/zendesk/ruby-kafka/blob/0d6c4510cb0bd881311d45d3c7cef148f0737805/README.md#oauthbearer), and not the `hash_function` as a symbol ala [`compression_codec`](https://github.com/zendesk/ruby-kafka/blob/0d6c4510cb0bd881311d45d3c7cef148f0737805/README.md#compression)? I'm only asking because the latter seems like a better fit here, but happy to be corrected 🙂

dasch: I'd prefer the former, since we're talking about varying the hash algo used by Partitioner, but not changing its fundamental strategy. Also, I prefer a minimal change in API – the high level libraries can easily make this a config key without any extra APIs.

ruby-kafka.gemspec
dasch: If it's C code then I think it should be opt-in, so that would be a good change :+1:

dasch (commented):
Looks great! Can you also 1) update the README to document this, and perhaps add docs to `Partitioner` describing the setup, and 2) update the CHANGELOG?

lib/kafka/client.rb
divo: Great, thanks for breaking it down for me. I've removed the API change and moved it inside the `Partitioner`

ruby-kafka.gemspec
divo: Done, thanks!

divo:
@dasch Thank you, and yes of course. I'm still thinking about the naming of the `Digest` module but if your happy with it that's fine.

divo:
@dasch I've updated the documentation, let me know what you think.

dasch (approved):
Awesome, thanks!

PR #888, opened 2021-02-26
Bump DescribeConfigs API to version 1

rkrage:
Version 1 of this API exposes `config_source` which describes the origin of a particular config (topic, broker, default value, etc). We have tooling that manages topic-level configs so being able to filter on `config_source` would be very helpful.

As far as I can tell, `DescribeConfigs` is used in two places: `describe_topic` and `describe_configs` (broker configs). The former is noted to be an alpha API while the latter is not.

I'm not sure about the compatibility goals of this library (version 1 of this API was introduced in Kafka 1.1.0), so I just wanted to start a conversation.

Reference: https://kafka.apache.org/protocol#The_Messages_DescribeConfigs

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -637,8 +637,8 @@ module Kafka
     # @param name [String] the name of the topic.
     # @param configs [Array<String>] array of desired config names.
     # @return [Hash<String, String>]
-    def describe_topic(name, configs = [])
-      @cluster.describe_topic(name, configs)
+    def describe_topic(name, configs: nil, config_source: nil)
+      @cluster.describe_topic(name, configs, config_source)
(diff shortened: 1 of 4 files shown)


dasch:
Looks like this change breaks the integration tests.

I'm not quite sure how many people are using Kafka versions older than v1.1 – since there's no support for multiple versions of an API I'm generally conservative about bumping the required version.

rkrage:
I can get the tests passing by updating a few expectations and skipping some specs for v1.0 / v0.11. But based on your comments it seems like maybe we don't want to go forward with this? We can probably just live with our monkey patch for now and work towards getting these APIs implemented in `rdkafka-ruby`.

PR #893, opened 2021-04-07
Updated to ignore all control batches

rkruze:
Control batches can show up even outside of transactions. This change drops the requirement that you must be in a transaction for a record to be labeled a control batch.

All tests pass with this change. 

This change is very similar to the ones made in other Kafka clients: 

- Sarama: https://github.com/Shopify/sarama/pull/1898
- librdkafka: https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_msgset_reader.c#L860


dasch (approved):
Thanks!

PR #896, opened 2021-04-14
Handle SyncGroup responses with a non-zero error and no assignments

BrentWheeldon:
When the error code is non-zero there isn't necessarily (ever?) an assignment. Currently the decoder raises a `TypeError` ("no implicit conversion of nil into String") - this updates the logic to only try to decode the assignment if there are any bytes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changes and additions to the library will be listed here.
 - Fix `Kafka::TransactionManager#send_offsets_to_txn` (#866).
 - Add support for `murmur2` based partitioning.
 - Add `resolve_seed_brokers` option to support seed brokers' hostname with multiple addresses (#877).
+- Handle SyncGroup responses with a non-zero error and no assignments (#896).
 
 ## 1.3.0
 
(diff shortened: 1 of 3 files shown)


dasch:
Shouldn't we raise an error in that scenario?

BrentWheeldon:
@dasch the appropriate error is raised by [`Protocol.handle_error`](https://github.com/zendesk/ruby-kafka/blob/733a47df7bd02b3195255ff071cb7687d47b652e/lib/kafka/protocol.rb#L160), at least in the situation I ran into (because of code in [`ConsumerGroup#synchronize`](https://github.com/zendesk/ruby-kafka/blob/733a47df7bd02b3195255ff071cb7687d47b652e/lib/kafka/consumer_group.rb#L206)). I'm not overly familiar with the internals of this gem but it seems like the error raising isn't generally happening in the response objects, but rather in the code that's making the request and using the response (like `ConsumerGroup#synchronize` in the situation I ran into).

Having said that, definitely happy to make this raise if you think that's appropriate. Would calling `Protocol.handle_response` in the `.decode` method make the most sense in that case?

dasch:
No, you're right – and if the error is already being raised then it's fine :+1:

dasch (approved):

dasch:
Thanks!

PR #897, opened 2021-04-14
Fix "@param tag has unknown parameter name"

abicky:
This PR resolves the following warnings:

```
% yard doc
[warn]: @param tag has unknown parameter name: group_id:
    in file `lib/kafka/cluster.rb' near line 122
[warn]: @param tag has unknown parameter name: transactional_id:
    in file `lib/kafka/cluster.rb' near line 132
[warn]: @param tag has unknown parameter name: string
    in file `lib/kafka/protocol/encoder.rb' near line 131
-- snip --
```

### DIFF
--- a/lib/kafka/cluster.rb
+++ b/lib/kafka/cluster.rb
@@ -117,7 +117,7 @@ module Kafka
 
     # Finds the broker acting as the coordinator of the given group.
     #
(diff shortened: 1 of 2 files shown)


dasch (approved):
Thanks!

PR #903, opened 2021-06-08
ISSUE-525 / 764: ruby-kafka does not support different topic subscriptions in the same consumer group

eduardopoleoflipp:
There are 2 issues that currently prevent supporting this feature:

1. **The current `RoundRobinAssignmentStrategy` assumes all subscriptions are equal**

We need a more generalized algorithm to perform assignments now that we have different subscriptions across the consumer group. This PR introduces a new additional strategy heavily inspired in the java kafka client [RoundRobinAssignor](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/RoundRobinAssignor.java) that properly handle this case.

2. **The leader consumer thread assumes that all consumers subscription are equivalent**

If different consumers were to have different topic subscription we can't rely on the in-memory `@topics` variable to produce correct assignments. Instead, we can leverage the subscription information provided by the cluster, and stored in`@members`, to determine the correct consumer subscriptions for a particular consumer group. Once we have the correct list of topics the downstream code in the assignor can then fetch the corresponding partitions from the cluster in preparation for the assignments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Changes and additions to the library will be listed here.
 - Add support for `murmur2` based partitioning.
 - Add `resolve_seed_brokers` option to support seed brokers' hostname with multiple addresses (#877).
 - Handle SyncGroup responses with a non-zero error and no assignments (#896).
+- Add support for non-identical topic subscriptions within the same consumer group (#525 / #764).
 
 ## 1.3.0
 
--- a/lib/kafka/consumer_group.rb
+++ b/lib/kafka/consumer_group.rb
@@ -189,9 +189,14 @@ module Kafka
       if group_leader?
         @logger.info "Chosen as leader of group `#{@group_id}`"
 
+        topics = Set.new
+        @members.each do |_member, metadata|
+          metadata.topics.each { |t| topics.add(t) }
+        end
+
         group_assignment = @assignor.assign(
           members: @members,
-          topics: @topics,
+          topics: topics,
         )
       end
 
--- a/lib/kafka/round_robin_assignment_strategy.rb
+++ b/lib/kafka/round_robin_assignment_strategy.rb
@@ -1,9 +1,9 @@
-# frozen_string_literal: true
-
 module Kafka
 
-  # A consumer group partition assignment strategy that assigns partitions to
-  # consumers in a round-robin fashion.
+  # A round robin assignment strategy inpired on the
+  # original java client round robin assignor. It's capable
+  # of handling identical as well as different topic subscriptions
+  # accross the same consumer group.
   class RoundRobinAssignmentStrategy
     def protocol_name
       "roundrobin"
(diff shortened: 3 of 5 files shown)


mrnonz:
Can't wait!!

dasch:
Would there be a reason to not _always_ use the new assignment strategy? I'd prefer to have fewer options to confuse people.

lib/kafka/multi_subscription_round_robin_assignment_strategy.rb
```diff
@@ -0,0 +1,56 @@
...
+        .select { |partition| subscribed_topics.include?(partition.topic) }
+        .sort_by { |partition| partition.topic }
+    end
+
+    def next_index(members_ids, idx)
+      idx += 1
+      idx = 0 if idx == members_ids.size
+
+      idx
+    end
```
dasch, line 54: You can use this fancy Ruby trick instead, which I think is more readable:

```
irb(main):001:0> x = (0...5).cycle
irb(main):003:0> x.next
=> 0
irb(main):004:0> x.next
=> 1
irb(main):005:0> x.next
=> 2
irb(main):006:0> x.next
=> 3
irb(main):007:0> x.next
=> 4
irb(main):008:0> x.next
=> 0
```

dasch: So you'd do e.g. this in `call`:

```ruby
idx = (0...members.size).cycle

# ...
id = idx.next
```

dasch:
Looks good! I'd like to see if we could perhaps merge this with the existing strategy rather than adding a new one, if this is purely additional in terms of functionality and reliability.

lib/kafka/multi_subscription_round_robin_assignment_strategy.rb
eduardopoleoflipp: oh nice will do!

eduardopoleoflipp:
I just wanted to be cautious since I'm not all too familiarized with the project but yes, I can merge this with the original strategy. It should be easy enough since this already passes all the original strategy unit tests. Thanks for taking time to review.

eduardopoleoflipp:
Hey @dasch ! I incorporated your suggestions. There are a few failing tests but I have a strong suspicious that these are flaky tests. There are all of the type of:
```
Kafka::ConnectionError:
       Could not connect to any of the seed brokers:
       - kafka://localhost:9092: Connection error IOError: stream closed in another thread
      
Kafka::UnknownError:
       Kafka::UnknownError

Kafka::DeliveryFailed:
       Could not connect to any of the seed brokers:
       - kafka://localhost:9092: Connection error IOError: stream closed in another thread
```
I've pushed already 3 times and I'm always getting similar errors but for different specs. Also I find it hard to believe that my changes would break the connection to the broker.

dasch:
Yeah, unfortunately the test suite is really flake – I would love to have someone improve that :D

dasch (changes requested):
This looks good! Can you add an entry to the changelog?

eduardopoleoflipp:
Great! just added the entry to the changelog. Thank you so much!

dasch:
Awesome! I'd like to cut an alpha release and have you test that out in production, if that's okay?

eduardopoleoflipp:
Great yes of course, we will!

eduardopoleoflipp:
@dasch I'll ping you on the dev slack channel once we've ascertained that everything is looking good on our system with the new changes.

dasch:
Great!

PR #905, opened 2021-07-08
Make ssl_ca_cert_file_path support an array of files

syedriko:
A proposed change to facilitate fixing https://github.com/fluent/fluent-plugin-kafka/issues/287. It would help to avoid parsing certs in the upper level code that should not need to know how to do that.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # frozen_string_literal: true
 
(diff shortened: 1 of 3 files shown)


dasch (approved):
Thanks!

syedriko:
> Thanks!

You're welcome! I didn't mean to just drop a PR on your lap without an explanation but got pulled into something else. This change really helps us avoid parsing certs in the Fluentd Kafka plugin.

PR #907, opened 2021-07-21
AWS IAM SASL auth for MSK

jobeus:
There wasn't support for authenticating with AWS's IAM, so here it is.

Based off instructions @ https://github.com/aws/aws-msk-iam-auth

Something along these lines to test/connect:

```
k = Kafka.new(
  seed_brokers: [array of seed brokers as defined by aws],
  client_id: 'client_id',
  logger: Rails.logger,
  partitioner: Kafka::Partitioner.new,
  sasl_aws_msk_iam_access_key_id: 'iam_access_key',
  sasl_aws_msk_iam_secret_key_id: 'iam_secret_key',
  sasl_aws_msk_iam_aws_region: 'us-west-2',
  ssl_ca_certs_from_system: true
)
k.topics
```

IAM auth-enabled kafka instances should be on post 9098, so seed broker array such as `['host1.aws.com:9098', 'host2.aws.com:9098']` etc.

More instructions on how to find these hostnames for your MSK setup @ https://github.com/aws/aws-msk-iam-auth

### DIFF
--- a/lib/kafka/sasl/awsmskiam.rb
+++ b/lib/kafka/sasl/awsmskiam.rb
@@ -1,0 +1,149 @@
+# frozen_string_literal: true
+
+require 'securerandom'
+require 'base64'
+require 'json'
+
+module Kafka
+  module Sasl
+    class AwsMskIam
+      AWS_MSK_IAM = "AWS_MSK_IAM"
+
+      def initialize(aws_region:, access_key_id:, secret_key_id:, logger:)
+        @semaphore = Mutex.new
+
+        @aws_region = aws_region
+        @access_key_id = access_key_id
+        @secret_key_id = secret_key_id
+        @logger = TaggedLogger.new(logger)
+      end
+
+      def ident
+        AWS_MSK_IAM
+      end
+
+      def configured?
+        @aws_region && @access_key_id && @secret_key_id
+      end
+
+      def authenticate!(host, encoder, decoder)
+        @logger.debug "Authenticating #{@access_key_id} with SASL #{AWS_MSK_IAM}"
+
+        host_without_port = host.split(':', -1).first
+
+        msg = authentication_payload(host: host_without_port)
+        @logger.debug "Sending first client SASL AWS_MSK_IAM message:"
+        @logger.debug msg
+        encoder.write_bytes(msg)
+
+        begin
+          @server_first_message = decoder.bytes
+          @logger.debug "Received first server SASL AWS_MSK_IAM message: #{@server_first_message}"
+
+          raise Kafka::Error, "SASL AWS_MSK_IAM authentication failed: unknown error" unless @server_first_message
+        rescue Errno::ETIMEDOUT, EOFError => e
+          raise Kafka::Error, "SASL AWS_MSK_IAM authentication failed: #{e.message}"
+        end
+
+        @logger.debug "SASL #{AWS_MSK_IAM} authentication successful"
+      end
+
+      private
+
+      def aws_uri_encode(str, encode_slash = true)
+        result = ''
+        chars = str.split('')
+        chars.each do |char|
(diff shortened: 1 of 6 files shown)


lib/kafka/sasl/awsmskiam.rb
```diff
@@ -0,0 +1,149 @@
...
+        @logger.debug "SASL #{AWS_MSK_IAM} authentication successful"
+      end
+
+      private
+
+      def aws_uri_encode(str, encode_slash = true)
+        result = ''
+        chars = str.split('')
+        chars.each do |char|
+          if ('A'..'Z').include?(char) || ('a'..'z').include?(char) || ('0'..'9').include?(char) || char == '_' || char == '-' || char == '~' || char == '.'
```
dasch, line 57: Is this function somehow specific to AWS? We can't use normal URI encoding here? At the very least we'd need a comment describing what this does, and why.

lib/kafka/sasl/awsmskiam.rb
```diff
@@ -0,0 +1,149 @@
...
+      end
+
+      def canonical_query_string
+        now = Time.now
+        aws_uri_encode("Action") + "=" + aws_uri_encode("kafka-cluster:Connect") + "&" +
+        aws_uri_encode("X-Amz-Algorithm") + "=" + aws_uri_encode("AWS4-HMAC-SHA256") + "&" +
+        aws_uri_encode("X-Amz-Credential") + "=" + aws_uri_encode(@access_key_id + "/" + now.strftime("%Y%m%d") + "/" + @aws_region + "/kafka-cluster/aws4_request") + "&" +
+        aws_uri_encode("X-Amz-Date") + "=" + aws_uri_encode(now.strftime("%Y%m%dT%H%M%SZ")) + "&" +
+        aws_uri_encode("X-Amz-Expires") + "=" + aws_uri_encode("900") + "&" +
+        aws_uri_encode("X-Amz-SignedHeaders") + "=" + aws_uri_encode("host")
```
dasch, line 113: Perhaps use <https://ruby-doc.org/stdlib-2.6.3/libdoc/uri/rdoc/URI.html#method-c-encode_www_form> here instead.

dasch (changes requested):
Thanks for opening this PR! I've got some feedback before I can merge this:

1. It would be great to leverage standard library methods a bit more – I'm a bit worried about the amount of low level code introduced here.
2. I would love to see this tested in production before merging. Are you currently using this?
3. There needs to be an update to the README and the changelog.

garrett528:
@jobeus pulled this branch and keep getting `Could not connect to any of the seed brokers: (Kafka::ConnectionError)
SASL AWS_MSK_IAM authentication failed: end of file reached`. doesn't seem to make it past the `encoder.write_bytes(msg)` call. 

have you run into this? i can get this to work with multiple other clients on my machine (`librdkafka`, the standard Java client, etc.)

i don't see anything wrong with your code but was wondering if you had run into this since it's unclear why the metadata fetch is failing. 

once i can get this working, i want to extend this to add support for temp credentials / AssumeRole refreshes.

garrett528:
figured out the issue here. when building the timestamp to be used to calculate the various parts of the signature, the code needs to call `Time.now.utc` instead of `Time.now` to ensure the timestamp is using UTC and not a local timezone. furthermore, the timestamp should be determined before calling `authentication_payload` and that single timestamp value should be passed through the call to each area of the signature process that requires the current datetime timestamp instead of calling `Time.now` each time.

kirkokada:
Hi @jobeus @dasch , I am interested in using this feature in production, is there anything I can help with to get this merged? 

Looks like to fix the unit tests all you need to do is add a few lines to `spec/sasl_authenticator_spec.rb` 

```ruby
# spec/sasl_authenticator_spec.rb:33

  let(:auth_options) {
    {
      # ...
      sasl_aws_msk_iam_access_key_id: nil,
      sasl_aws_msk_iam_secret_key_id: nil,
      sasl_aws_msk_iam_aws_region: nil
    }
  }
```

jobeus:
Continued better in #937 let's get this merged :D

PR #910, opened 2021-07-30
Fix multiple `[Producer name]` tags on failures

ojab:
On `rescue Kafka::Error` we're rerunning the whole method block
including `@logger.push_tags(@producer.to_s)`, but clearing tags
only in `ensure` which is not run on `retry`, so each retry adds
another one `[Producer name]` to tags and, as a consequence, all
the log messages after that.


ojab:
CI failures look unrelated.

dasch (approved):
Thanks!

PR #940, opened 2022-05-02
Use CircleCI matrix jobs to test with the latest version

mishina2228:
This PR updates `.circleci/config.yml` to use CircleCI matrix jobs.
It can make it easy to test with multiple Ruby and Kafka versions, and also can simplify duplicate statements.
Here are the details:
- Upgrade to next-gen Docker convenience images
  The following message is shown in Circle CI job page.
  > You’re using a [deprecated Docker convenience image.](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) Upgrade to a next-gen Docker convenience image.
  
  The link above states the following.

  > Moving from a legacy to next-gen image requires a change to the namespace. All legacy images have a Docker namespace of circleci, while next-gen images have a Docker namespace of cimg. For example, migrating from the legacy Ruby or Python image to the respective next-gen image can be done as follows:

  > circleci/ruby:2.3.0 → cimg/ruby:2.3.0
  > circleci/python:3.8.4 → cimg/python:3.8.4

  `circleci/ruby:2.5.1-node` is used now, but because of the following reasons, `cimg/ruby:2.5` will be used.
  - cimg doesn't provide Ruby 2.5.1, and the latest patch version is 2.5.9.
  - Specify a minor release tag for the latest patch version to use.
    https://circleci.com/developer/images/image/cimg/ruby#image-tags
  - It seems that Node.js is not needed, so I omit the suffix `-node`.

  BTW, Ruby 2.5 has already reached EOL, but it's left as is in this PR.
- While unit tests are currently only run on Ruby 2.5, matrix is used so that we can add test for other Ruby versions later.
- There are two types of kafka images(provided by wurstmeister/bitnami), with different environment variables to pass.
  So I defined jobs for each(`functional-with-wurstmeister`/`functional-with-bitnami`).
- The workflow name was set for clarity. Note that the name has changed from the previous name.
  Before: 
  <details><summary>Show image</summary>

  image
  </details>

  After: 
  <details><summary>Show image</summary>

  image
  </details>

The CI status can be seen [here](https://app.circleci.com/pipelines/github/mishina2228/ruby-kafka/11/workflows/bb2f4945-9b3f-42ce-9d27-91670afd69d7).
Functional tests fail randomly, but I believe this is not related with changes in this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,11 @@
-version: 2
+version: 2.1
 jobs:
   unit:
+    parameters:
+      ruby-version:
+        type: string
     docker:
-      - image: circleci/ruby:2.5.1-node
+      - image: cimg/ruby:<< parameters.ruby-version >>
         environment:
           LOG_LEVEL: DEBUG
     steps:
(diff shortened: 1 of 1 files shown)


mishina2228:
@dasch Sorry to bother you, but I would appreciate it if you review this PR, thanks!

dasch:
Thanks!

mishina2228:
Thank you for merging it! But it appears that CI failed due to authentication issues.
https://app.circleci.com/pipelines/github/zendesk/ruby-kafka/253
> Could not find a usable config.yml, you need to reauthorize and approve the SAML-protected org.
Please sign out of CircleCI and log back in with your VCS before triggering a new pipeline.

Could you take a look at it?

Issue #941, opened 2022-05-17
Join request timeout doesn't consider rebalance_timeout


abicky:
* Version of Ruby: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin20]
* Version of Kafka: 6.1.1-ccs (Commit:c209f70c6c2e52ae)
* Version of ruby-kafka: 1.4.0 (555281eb407c9fc7aa08e7ebedc950959ad4ac5f)

###### Steps to reproduce

```ruby
require 'logger'
require 'kafka'

GROUP_ID = ENV.fetch('GROUP_ID')
TOPIC = ENV.fetch('TOPIC')
BROKERS = %w[127.0.0.1:9092]

PARTITIONS = 4

def create_consumer(kafka, heartbeat_interval: 10)
  consumer = kafka.consumer(group_id: GROUP_ID, session_timeout: 60, heartbeat_interval: heartbeat_interval)
  consumer.subscribe(TOPIC)
  consumer.each_message do |message|
  end
end

kafka = Kafka.new(BROKERS, logger: Logger.new($stdout).tap { |l| l.level = Logger::WARN })
unless kafka.topics.include?(TOPIC)
  kafka.create_topic(TOPIC, num_partitions: PARTITIONS)
end

logger = Logger.new($stdout)
Thread.new do
  command = "kafka-consumer-groups --bootstrap-server #{BROKERS.first} --describe --group #{GROUP_ID}"
  loop do
    logger.info(command)
    system(command)
    sleep 5
  end
end

logger.info 'Create the first consumer'
Thread.new { create_consumer(kafka, heartbeat_interval: 15) }

sleep 1

logger.info 'Create the second consumer'
Thread.new { create_consumer(kafka) }

sleep 20

logger.info 'Create the third consumer'
create_consumer(kafka)
```

###### Expected outcome

Join requests wait more than `rebalance_timeout`, and consumers don't resend join requests over and over as follows:

```
% TOPIC=test GROUP_ID=test ruby -Ilib ~/ghq/src/github.com/zendesk/ruby-kafka/join.rb
I, [2022-05-18T05:11:39.793155 #73498]  INFO -- : Create the first consumer
I, [2022-05-18T05:11:39.793328 #73498]  INFO -- : kafka-consumer-groups --bootstrap-server 127.0.0.1:9092 --describe --group test

[...]

PR #943, opened 2022-05-17
Consider rebalance_timeout on join request

abicky:
This PR resolves https://github.com/zendesk/ruby-kafka/issues/941.

As the coordinator waits for each member to rejoin when rebalancing the group, the timeout of join request should be greater than rebalance_timeout. 

> rebalance_timeout_ms : The maximum time in milliseconds that the coordinator will wait for each member to rejoin when rebalancing the group.

cf. https://kafka.apache.org/protocol.html#The_Messages_JoinGroup

Otherwise, each member might send join requests over and over, and as a result, the following problems occur:

* Some partitions are never processed until session_timeout passes
* It takes much time for rebalance that occurs in a short time to finish

### DIFF
--- a/lib/kafka/broker.rb
+++ b/lib/kafka/broker.rb
@@ -89,7 +89,7 @@ module Kafka
     def join_group(**options)
       request = Protocol::JoinGroupRequest.new(**options)
 
-      send_request(request)
+      send_request(request, read_timeout: options[:rebalance_timeout])
     end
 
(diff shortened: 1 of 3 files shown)


abicky:
I'll close this PR because this PR will never seem to be merged cf. https://github.com/zendesk/ruby-kafka/commit/a5491ac19ec5d427dede5b63f66c55db240e64f1.

PR #947, opened 2022-06-11
Make example "create partitions" stable

abicky:
This PR prevents the example "create partitions" from failing like https://app.circleci.com/pipelines/github/zendesk/ruby-kafka/221/workflows/7d51470f-c102-4b30-b2e0-68c618eeb1d2/jobs/6799?invite=true#step-104-122.

As there is a time lag for Kafka's metadata to be updated, `Kafka::Client#partitions_for` is needed to be executed until the metadata is updated to check the latest partition count. However, `Kafka::Cluster` never fetches new metadata until it considers the current metadata stale.
So, this PR adds the method `Kafka#refresh_metadata` in order to force Kafka::Cluster to fetch metadata. The option helps in production code as well to ensure we get the latest metadata.

### DIFF
--- a/lib/kafka/client.rb
+++ b/lib/kafka/client.rb
@@ -814,6 +814,10 @@ module Kafka
       @cluster.disconnect
     end
 
+    def refresh_metadata
+      @cluster.refresh_metadata!
+    end
+
     private
 
     def initialize_cluster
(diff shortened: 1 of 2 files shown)


abicky:
I'll close this PR because this PR will never seem to be merged cf. https://github.com/zendesk/ruby-kafka/commit/a5491ac19ec5d427dede5b63f66c55db240e64f1.

Issue #944, opened 2022-06-06
MSK IAM: add support for AssumeRole auto refresh 


hnaoto:
Hi, MSK IAM auth support was added recently https://github.com/zendesk/ruby-kafka/pull/937, but it doesn't support temp credentials/ assume role refreshes.  This feature was initially mentioned by @garrett528 in a thread https://github.com/zendesk/ruby-kafka/pull/907#issuecomment-978146627 

I am interested in adding support for auto refreshing AssumeRole. 

Ideally, the client will retrieve temporary credentials from STS and refresh those credentials in the background. 

```
      sts = Aws::STS::Client.new
      role_credentials = Aws::AssumeRoleCredentials.new(
        client: sts,
        role_arn: "arn:aws:iam::......",
        role_session_name: "example_session"
      )
    
  # can access access_key_id by using  role_credentials.access_key_id
```

The `sasl_authenticator` is initiated when the kafka client got created https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/sasl_authenticator.rb#L42. I am wondering what is the appropriate way of refreshing the credentials used by in `sasl_authenticator` the background. Theoretically, I can recreate the `sasl_authenticator` after the credentials get updated, but that feels a little bit hacky to me..

Any suggestions are much appreciated.

garrett528:
@hnaoto the way that i've seen this done, specifically in `librdkafka` which is the standard C client for Kafka, is that the credentials are stored in an object. that object is refreshed on a certain cadence (about 80% of the way through the credential lifetime set by `duration_sec`). however, this requires that the refresh mechanism is 1. scheduled and 2. running on a background thread. the credential object requires that it be locked and unlocked upon update since multiple threads may be trying to access it concurrently.

here's what i got to in late 2021. https://github.com/garrett528/ruby-kafka/pull/2/files. it's not complete and i don't think i ever got the thread to be scheduled properly (i'm no rubyist so i may be heading in the wrong direction trying to port C to Ruby).

here's the `librdkafka` C implementation that i wrote to do this if it helps. https://github.com/UrbanCompass/librdkafka/blob/master/src/rdkafka_sasl_aws_msk_iam.c

hnaoto:
@garrett528 Thanks you so much for sharing all the details. The solution that you tried sounds promising to me. Let me see whether I can figure out the scheduling part.

hnaoto:
[...]

PR #951, opened 2022-06-15
add support for aws assume role credentials

hnaoto:
#### Description
- The existing implementation doesn't support temporary IAM credentials. A user has to create an IAM user, get `iam_access_key` and `iam_secret_key` and then use those values to create a Ruby Kafka client. 
- This MR will add support for aws iam assume role credentials. User will be able to retrieve `AssumeRoleCredentials` from a STS client and use the `AssumeRoleCredentials` to create a Ruby Kafka client in the following fashion.
```
sts = Aws::STS::Client.new
# to get more information about how to create AssumeRoleCredentials
# please refer to https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleCredentials.html
role_credentials = Aws::AssumeRoleCredentials.new(
  client: sts,
  role_arn: "role_arn",
  role_session_name: "role_session_name"
  
)
kafka_client = Kafka.new(
  ["broker_address"],
  aws_iam_assume_role_credentials: role_credentials,
  sasl_aws_msk_iam_aws_region: "us-east-2",
  ssl_ca_certs_from_system: true,
)
```

#### Implementation Details
- Although the `AssumeRoleCredentials`  are temporary credentials and will expire in 1 hour by default. The code change in this MR won't need to refresh `AssumeRoleCredentials`.
- The `AssumeRoleCredentials` will be refreshed by AWS SDK in the background automatically (For more details, please refer to https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/refreshing_credentials.rb). The  `AssumeRoleCredentials` is passed to the Kafka client as an object and it will get updated in the background as well.
-  Either `assume_role_credentials` object or the combination of `access_key_id` and `secret_key_id` has to be supplied. If neither of those parameters are supplied, an `AwsMskIamCredentialsException` error will be raised internally and an error message will be logged. The Kafka client will crash eventually due to `ConnectionError`. 

#### Discussions 
Discussions regarding implementation details and testing can be found in https://github.com/zendesk/ruby-kafka/issues/944

#### Documentations and Comments
- [x] add comments
- [x] add descriptions to readme
- [x] add sample code to example directory

#### Testing
- [x] test with a small MSK cluster and a skeleton ruby program...
- [x] add unit tests
- [x] test in working environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -1135,6 +1135,28 @@ k = Kafka.new(
 )
 ```
 
+or use a STS client to retrieve temporary AssumeRoleCredentials. AssumeRoleCredentials will be updated in the background automatically.
+```ruby
+sts = Aws::STS::Client.new
+
+# to get more information about how to create AssumeRoleCredentials
+# please refer to https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/AssumeRoleCredentials.html
+role_credentials = Aws::AssumeRoleCredentials.new(
+  client: sts,
+  role_arn: "role_arn",
+  role_session_name: "role_session_name"
+  
+)
+
+kafka_client = Kafka.new(
+  ["broker_address"],
+  aws_iam_assume_role_credentials: role_credentials,
+  sasl_aws_msk_iam_aws_region: "us-east-2",
+  ssl_ca_certs_from_system: true,
+)
+```
+
+
 ##### PLAIN
 In order to authenticate using PLAIN, you must set your username and password when initializing the Kafka client:
 
--- a/lib/kafka/sasl/awsmskiam.rb
+++ b/lib/kafka/sasl/awsmskiam.rb
@@ -14,6 +14,11 @@ module Kafka
       end
     end
 
+    # AWS IAM Credentials will be encapsulated in an object
+    # so when the assume role credential object gets refreshed in the background by AWS SDK,
+    # the instance variable @assume_role_credentials will be "refreshed" as well(Ruby is "pass by object reference");
+    # if one just tried to retrieve access_key_id and secret_key_id from the temporary credential and pass those values as strings,
+    # those strings would not get updated in the background
     class AwsMskIamCredentials
       def initialize(access_key_id: nil, secret_key_id: nil, session_token: nil, assume_role_credentials: nil, logger:)
         @access_key_id = access_key_id
--- a/lib/kafka/sasl/awsmskiam.rb
+++ b/lib/kafka/sasl/awsmskiam.rb
@@ -6,16 +6,36 @@ require 'json'
 
 module Kafka
   module Sasl
-    class AwsMskIam
-      AWS_MSK_IAM = "AWS_MSK_IAM"
 
-      def initialize(aws_region:, access_key_id:, secret_key_id:, session_token: nil, logger:)
-        @semaphore = Mutex.new
-
-        @aws_region = aws_region
+    class AwsMskIamCredentials
+      def initialize(access_key_id: nil, secret_key_id: nil, session_token: nil, assume_role_credentials: nil)
         @access_key_id = access_key_id
         @secret_key_id = secret_key_id
         @session_token = session_token
+        @assume_role_credentials = assume_role_credentials
+      end
+
+      def get_access_key_id
+        @assume_role_credentials.nil? ? @access_key_id : @assume_role_credentials.credentials.access_key_id
+      end
+
+      def get_secret_key_id
+        @assume_role_credentials.nil? ? @secret_key_id : @assume_role_credentials.credentials.secret_access_key
+      end
+
+      def get_session_token
+        @assume_role_credentials.nil? ? @session_token : @assume_role_credentials.credentials.session_token
+      end
+    end
+
+    class AwsMskIam
+      AWS_MSK_IAM = "AWS_MSK_IAM"
+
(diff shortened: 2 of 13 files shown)


hnaoto:
Hi @garrett528 and @jobeus, would you mind taking a look at the MR?
Please let me know if there is anything appears to be unclear to you.

(I will try to loop in the maintainers of this repo later.

lib/kafka/sasl/awsmskiam.rb
```diff
@@ -90,10 +135,10 @@ def canonical_query_string(time_now:)
         params = {
           "Action" => "kafka-cluster:Connect",
           "X-Amz-Algorithm" => "AWS4-HMAC-SHA256",
-          "X-Amz-Credential" => @access_key_id + "/" + time_now.strftime("%Y%m%d") + "/" + @aws_region + "/kafka-cluster/aws4_request",
+          "X-Amz-Credential" => @aws_msk_iam_credentials.get_access_key_id + "/" + time_now.strftime("%Y%m%d") + "/" + @aws_region + "/kafka-cluster/aws4_request",
           "X-Amz-Date" => time_now.strftime("%Y%m%dT%H%M%SZ"),
           "X-Amz-Expires" => "900",
-          "X-Amz-Security-Token" => @session_token,
+          "X-Amz-Security-Token" => @aws_msk_iam_credentials.get_session_token,
```
garrett528, line 141: does this get messed up if the session token isn't passed and defaults to `nil`?

lib/kafka/sasl/awsmskiam.rb
```diff
@@ -68,11 +113,11 @@ def authentication_payload(host:, time_now:)
           'user-agent' => "ruby-kafka",
           'action' => "kafka-cluster:Connect",
           'x-amz-algorithm' => "AWS4-HMAC-SHA256",
-          'x-amz-credential' => @access_key_id + "/" + time_now.strftime("%Y%m%d") + "/" + @aws_region + "/kafka-cluster/aws4_request",
+          'x-amz-credential' => @aws_msk_iam_credentials.get_access_key_id + "/" + time_now.strftime("%Y%m%d") + "/" + @aws_region + "/kafka-cluster/aws4_request",
           'x-amz-date' => time_now.strftime("%Y%m%dT%H%M%SZ"),
           'x-amz-signedheaders' => "host",
           'x-amz-expires' => "900",
-          'x-amz-security-token' => @session_token,
+          'x-amz-security-token' => @aws_msk_iam_credentials.get_session_token,
```
garrett528, line 120: see below. it should be acceptable to not pass a session token if you are using permanent credentials. in that case, we don't want to set this parameter (i think it'll fail if we set it to `nil` but want to confirm that that's the behavior).

hnaoto: Yes, I think you are right.

When I test the existing implementation, I did notice that I have to pass `session_token` when I use static credentials. I didn't realize it was caused by setting 'x-amz-security-token' as `nil`... Let me try to set `x-amz-security-token` as optional

hnaoto: I think that is a good point. `.delete_if { |_, v| v.nil? }` should be able to remove any keypair if the value is nil

lib/kafka/sasl/awsmskiam.rb
hnaoto: should still work.. `.delete_if { |_, v| v.nil? }` will remove `"X-Amz-Security-Token" => nil`

jobeus:
lgtm in general but changed jobs and don't have access to a kafka cluster in AWS right now to test on, etc..

hnaoto:
Thanks @garrett528 😄!  really appreciate your help. Thanks for joining the discussion and sharing your insights

hnaoto:
Thanks for reviewing the MR @jobeus! 
No worry. I will try to test the code change in a working environment

hnaoto:
Hi @dasch and @leonmaia, would you mind taking a look at this MR?

Please let me know if you need any further information.

jordanfbrown:
Hey @hnaoto, this looks great and is something my company is interested in. Wondering what the next steps are to get this approved and merged? I'm happy to help out with testing/finalizing the PR/etc.

hnaoto:
Hi @jordanfbrown, if you are interested in testing the MR, please give it a try. If you encounter any issue, please let me know. really appreciate that.

Please noted that MSK currently has connection constraints when one uses IAM authentication. If a large number of Kafka clients trying to establish connections through IAM authentication at the same time, there will be some serious connection issues. However, it depends on the number of the connections that one attempts to establish simultaneously, and your clients could work perfectly fine if they don't hit the bottleneck.


To get the MR merged we need to have it reviewed by @dasch.

dasch:
I would love to see users try this out before merging.

README.md
```diff
@@ -1135,6 +1135,28 @@ k = Kafka.new(
 )
 ```
 
+or use a STS client to retrieve temporary AssumeRoleCredentials. AssumeRoleCredentials will be updated in the background automatically.
```
dasch, line 1138: Can you make some of these terms links to the relevant AWS docs? First time users would benefit from that.

dasch (approved):
Other than that minor comment I'm :+1:

Would you like to get more testing by other users, or do you think this is stable?

Finally, can you update the CHANGELOG with this change?

PR #967, opened 2022-12-07
Add deprecation notice

mensfeld:
As discussed, it would be good to add deprec notice for folks plus warning in a post install message

### DIFF
--- a/README.md
+++ b/README.md
@@ -1322,11 +1322,13 @@ Latest stable release, with native support for the Kafka 0.10 protocol and event
 
 Currently, there are three actively developed frameworks based on ruby-kafka, that provide higher level API that can be used to work with Kafka messages and two libraries for publishing messages.
 
+**Note**: Many of the frameworks and libraries below use `ruby-kafka` only in older versions. See the deprecation notice at the top of this readme for more details.
+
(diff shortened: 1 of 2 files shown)


geoff2k:
What do you think about also adding a post-install message in the gemspec?

Something similar to:

```ruby
spec.post_install_message = %q{
This library is **no longer** actively developed and has been superseded by librdkafka via rdkafka-ruby bindings.

While this library may still receive security patches and bug fixes, it is no longer recommended for production usage.
}
```

mensfeld:
Great idea. Let me add it...

geoff2k (approved):
:shipit:

mensfeld:
@geoff2k added one more notice in the higher level libs section to indicate correct branches /versions of tools still using ruby-kafka

geoff2k (approved):
> @geoff2k added one more notice in the higher level libs section to indicate correct branches /versions of tools still using ruby-kafka

Excellent!  Hopefully those that need this information will be able to trip over it in one of the many places it now exists. 😆

mensfeld:
@dasch can we craft a release? :pray:

dasch (approved):

PR #969, opened 2023-01-10
add parallel heartbeat option

i512:


mensfeld:
Please read the deprecation notice in the wiki. This library is no longer recommended.

PR #989, opened 2026-07-27
Strip trailing `.` in broker hostnames

HeyNonster:
TSL Sever Name Indicators must not contain a trailing dot in the `HostName`[^1]. Currently, passing an
[fqdn](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) we get an  `SSLError`:

```ruby
OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 peeraddr=10.234.154.163:9093 state=error: sslv3 alert illegal parameter (SSL alert number 47) (OpenSSL::SSL::SSLError)
```

[^1]:(https://datatracker.ietf.org/doc/html/rfc6066#section-3:~:text=without%20a%20trailing%20dot)

### DIFF
--- a/lib/kafka/broker_uri.rb
+++ b/lib/kafka/broker_uri.rb
@@ -33,6 +33,9 @@ module Kafka
         uri.scheme = 'kafka+ssl'
       end
 
+      # See rfc6066: https://datatracker.ietf.org/doc/html/rfc6066#section-3:~:text=without%20a%20trailing%20dot
+      uri.hostname.delete_suffix!(".")
+
       unless URI_SCHEMES.include?(uri.scheme)
         raise Kafka::Error, "invalid protocol `#{uri.scheme}` in `#{str}`"
       end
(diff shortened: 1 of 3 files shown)


lib/kafka/broker_uri.rb
```diff
@@ -33,6 +33,9 @@ def self.parse(str)
         uri.scheme = 'kafka+ssl'
       end
 
+      # See rfc6066: https://datatracker.ietf.org/doc/html/rfc6066#section-3:~:text=without%20a%20trailing%20dot
+      uri.hostname.delete_suffix!(".")
```
bquorning, line 37: Does `URI` really not ship with a way to ignore trailing dots?

HeyNonster: I looked but couldn't find anything. `#normalize` doesn't work.

HeyNonster: `Addressable` has a `#normalized_host` method which works but nothing in URI.

razumau (approved):

